PR #2, opened 2020-04-11
issue_1 update serialize-javascript version

zqian:


ssciolla:
@zqian, not sure how you want about making these changes, but I believe changes to `package-lock.json` should be automatically generated. `serialize-javascript` appears to be a sub-dependency, so it won't be specified in `package.json`.

FWIW, I ran `npm audit` on the `src` directory and got the following results. 

```
found 1125 vulnerabilities (1123 low, 2 moderate) in 1806641 scanned packages
  run `npm audit fix` to fix 532 of them.
  3 vulnerabilities require semver-major dependency updates.
  590 vulnerabilities require manual review. See the full report for details.
```

Running `npm audit fix` resulted in numerous changes to `package-lock.json`, most of which we probably want. That seems to have bumped `serialize-javascript` to the right version, and the vulnerability summary looked something like this afterward:

```
found 593 low severity vulnerabilities in 1806645 scanned packages
  3 vulnerabilities require semver-major dependency updates.
  590 vulnerabilities require manual review. See the full report for details.
```

pushyamig:
Yeah, `package-lock.json` is like a snapshot of what version the current build is running on. we shouldn't be updating that. 
It Looks like `serialize-javascript` is a dependency of `terser-webpack-plugin` and this package is a dependency of `webpack`. I think Updating to the newest version of `webpack` in `package.json` should solve the problem

jlost:
Thanks for the report and the fix! The approach of changing `package-lock.json` seems to be the right one, as the `package.json` already correctly allows patch updates. I'm going to merge the results of `npm audit fix` (including those not related to serialize-javascript) into the dev branch per @ssciolla's suggestion. #27

PR #40, opened 2020-05-05
ASGI (uvicorn)

jlost:
Switches from gunicorn to uvicorn. Lays the groundwork for Django 3 async views and possibly Channels.


jlost:
Looks like I overestimated the current level of support. Async views aren't coming until Django 3.1, which is slated for release in August.

jonespm:
I haven't used ASGI yet but this looks like the switch that we'd use and the recommended server. All of our apps are still on the Django 2.2 LTS. 

My biggest concern I think is that Django 3 is still in development and Django 3.2 LTS won't be available until April 2021. 

If staying on 3.x, in the requirements.txt I'd personally lock down the Django versions (and maybe some other dependencies) a little more so it doesn't update automatically to 3.1 when that's released, but also be ready to upgrade to 3.1 in/around August when it's available. (And 3.2 next spring)

jlost:
Thanks for the feedback! I locked the version of Django (and all other pip packages) to the latest patch version in `requirements.txt`. My plan is to use 3.1 shortly after it's available, as I think we could benefit from the async views where we're frequently proxying other (BlueJeans) API calls.

jlost:
I updated the Dockerfile to use uvicorn as a gunicorn worker type as recommended in the docs:
https://www.uvicorn.org/deployment/

PR #45, opened 2020-05-11
Django Channels

### LINKED ISSUE #80: Use Django channels instead of polling at 3 second intervals
To reduce server load and allow application to scale

jlost:
Closes #80


jlost:
Working toward a basic prototype of websockets functionality. So far, the backend exposes a websockets route which sends initial state upon connection, and will then push an update whenever the Queue model is saved. The Channels Consumer uses our DRF Serializers to render the models.

Front-end is WIP.

src/officehours_api/consumers.py
```diff
@@ -0,0 +1,312 @@
...
+            'type': 'queue.deleted',
+        }
+    )
+
+
+@receiver(post_save, sender=Queue)
+def trigger_queue_update(sender, instance: Queue, created, **kwargs):
+    if instance.deleted:
+        return
+    transaction.on_commit(lambda: send_queue_update(instance.id))
```
jlost, line 119: Changes driven from the admin panel use `transaction.atomic()`, which causes a race condition where the message event handlers (e.g. `queue_update`) may outrun the transaction commit, causing out-of-date data to be sent. `on_commit` makes sure the messages aren't triggered until after the commit.

src/assets/src/components/home.tsx
```diff
@@ -30,22 +28,12 @@ function QueueLookup() {
...
-        if (!props.user) return;
-        return await apiGetUser(props.user.id)
-    }
     const [user, setUser] = useState(undefined as MyUser | undefined);
-    const [doRefreshUser, refreshUserLoading, refreshUserError] = usePromise(getUser, setUser);
-    useEffect(() => {
-        doRefreshUser();
-    }, []);
-    useAutoRefresh(doRefreshUser, 10000);
+    const userWebSocketError = useUserWebSocket(props.user!.id, (u) => setUser(u as MyUser));
```
bruinc, line 32: While I was testing this, if the user is not logged in, the home page body doesn't render because `useUserWebSocket` results in an error (there isn't a user yet). Do you get the same behavior?

jlost: Thanks for catching that! It should be fixed if you pull latest.

PR #47, opened 2020-05-15
Agenda field

bruinc:
Functionality goes smoothly until the host clicks "Join Info" to view the agenda, where there's a warning that reads:
Warning: Each child in a list should have a unique "key" prop.
image


src/assets/src/components/edit.tsx
```diff
@@ -241,6 +241,7 @@ const MeetingInfoDialog = (props: MeetingInfoProps) => {
                 Attendees: {props.meeting.attendees.map(a => <UserDisplay user={a}/>)}
             </p>
         );
+    const agendaInfo = <p>Agenda: {props.meeting?.agenda}</p>
```
jlost, line 244: I think this should be moved into `generalInfo`, since it applies to any Meeting regardless of its backend. Also, in the `generalInfo` block above, the type system will have already verified that `props.meeting` must exist, so there will be no need for a safe navigation operator (`meeting?.`).

src/assets/src/services/api.ts
```diff
@@ -171,3 +171,15 @@ export const searchQueue = async (term: string) => {
...
+    const resp = await fetch(`/api/meetings/${meeting_id}/`, {
+        method: "PATCH",
+        headers: getPatchHeaders(),
+        body: JSON.stringify({
+            agenda: agenda,
+        }),
+    });
+    await handleErrors(resp);
+    return await resp.json();
+}
```
jlost, line 185: This might have already been there, but can you add a newline to end of this file?

src/officehours_api/migrations/0015_auto_20200515_0941.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.6 on 2020-05-15 13:41
```
jlost, line 1: Can you squash these two migrations together? That will remove the redundancies (they're mostly the same) and combine them both into a single migration, which is easier to read, easier to reverse in the future, etc. Here's how:

https://docs.djangoproject.com/en/3.0/topics/migrations/#migration-squashing

src/officehours_api/serializers.py
```diff
@@ -105,11 +105,12 @@ class MeetingSerializer(serializers.ModelSerializer):
         source='attendees',
         write_only=True,
     )
+    #agenda = serializers.CharField()
```
jlost, line 108: Delete

src/assets/src/components/edit.tsx
```diff
@@ -237,9 +237,14 @@ interface MeetingInfoProps {
 const MeetingInfoDialog = (props: MeetingInfoProps) => {
     const generalInfo = props.meeting
         && (
+            <>
             <p>
                 Attendees: {props.meeting.attendees.map(a => <UserDisplay user={a}/>)}
             </p>
+            <p>
+                Agenda: {props.meeting?.agenda}
```
jlost, line 245: I think the safe navigation operator (`?.`) isn't needed here, because `props.meeting` definitely exists. Let me know if I'm mistaken

jlost:
Resolves #47

src/assets/src/components/edit.tsx
bruinc: Oops, sorry! This was part of your first comment. Should be fixed now.

jlost:
Approved, but let's hold off on merging until we have a 1.2 branch to merge it into.

jlost:
I changed the target branch to 1.2. There are some merge conflicts now - can you resolve them? Thanks.

bruinc:
Should be all set now!

jlost:
Looks great, merging!

PR #49, opened 2020-05-18
Message of the day

bruinc:
Current functionality: a blue banner with the flatpage content will be added for every flatpage titled "User Comm Banner".

Editing a flatpage in admin:
admin_edit

Result in UI:
banner_ui


jlost:
Looking good. Can we move the entire "alert" div into the flatpage, so that the alert variant (info, warning, danger etc.) can be changed in the flatpage? We could then replace the existing banners, since they could be fully replicated by flatpages.

bruinc:
base.html now assumes that the whole alert will be in the flatpage content.

Also, the banner depending on whether or not the program is running in debug mode is generated from the flatpage titled "Debug Banner" or "Prod Banner".

jlost:
I'm inclined to remove the `Debug Banner`/`Prod Banner` and only keep the `User Comm Banner` because I can't picture a scenario where we would benefit from having them toggle based on `DEBUG`.  Different Office Hours environments have their own databases, so each env must be individually configured with a banner anyway, and the `DEBUG` value would generally be static per environment. What do you think?

bruinc:
Okay, that makes sense. So the `User Comm Banner` will be the only banner that can have any/all messages needed.

jlost:
I tested this out and it all looks great. I'm going to merge it into dev and we'll release it with 1.1.

PR #57, opened 2020-05-22
Improve Backend Error Handling

jlost:
Made HTTP exceptions coming from backends return 502 and a message hinting that something went wrong in e.g. BlueJeans.

Made 404s from BlueJeans `/user` give a specific message telling users what to do.

Performed some minor refactors along the way.

### DIFF
--- a/src/assets/src/services/api.ts
+++ b/src/assets/src/services/api.ts
@@ -28,16 +28,22 @@ class ForbiddenError extends Error {
 
 const handleErrors = async (resp: Response) => {
     if (resp.ok) return;
+    let text: string;
+    let json: any;
     switch (resp.status) {
         case 400:
-            const json = await resp.json();
+            json = await resp.json();
             const messages = ([] as string[][]).concat(...Object.values<string[]>(json));
             const formatted = messages.join("\n");
             throw new Error(formatted);
         case 403:
-            const text = await resp.text();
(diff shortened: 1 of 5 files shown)


bruinc:
I've been able to get the 400 error "[email] does not exist in BlueJeans" to display when an attendee tries to join a queue that the host has not logged in to BlueJeans, and the email shown in the message is the host's email. If the attendee does not have a BlueJeans account, I can join the queue and get into a call without errors as long as the host does have a BlueJeans account.

If it's just the host that needs to have logged in to BlueJeans, it may be more useful to ping BlueJeans and display this message to the host when they are creating a queue. That way they are notified that they need to log in to BlueJeans before the office hours necessarily start. Just a thought.

Also, if the behavior I'm describing isn't what should be happening, let me know because it could be something I'm doing.

jlost:
Thanks for checking this! 

>If it's just the host that needs to have logged in to BlueJeans, it may be more useful to ping BlueJeans and display this message to the host when they are creating a queue. That way they are notified that they need to log in to BlueJeans before the office hours necessarily start. Just a thought.

I agree it would be better to send them a tip (or auto-create the BJ user for them) earlier in the process. If you want to create an issue to track a more user-friendly way to handle this problem, I think that would be worthwhile.

PR #63, opened 2020-05-30
Responsive meetings table

jlost:
Addresses bad meetings editor UX on mobile.

At all screen widths, Assignee select's width is retained.

image

At small screen size, the "Queue #" column is hidden.

image

At way-too-small screen sizes, the table becomes horizontally scollable.

image


mfldavidson:
I think that is the most elegant possible fix for this. Nice work @jlost

mfldavidson:
Just tested this in dev and it is much more usable thank you!!!

PR #75, opened 2020-06-04
User Preferences page - alternate contact info

bruinc:


src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { ErrorDisplay, LoadingDisplay, EditToggleField, LoginDialog, Breadcrumbs } from "./common";
+import { usePromise } from "../hooks/usePromise";
+import { useAutoRefresh } from "../hooks/useAutoRefresh";
+import { redirectToLogin } from "../utils";
+import { PageProps } from "./page";
+
+interface PreferencesEditorProps {
+    user: User;
+    disabled: boolean;
+    onUpdateInfo: (phone_number: string) => void;
```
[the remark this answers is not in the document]
bruinc: All of the attributes in models.ts are snake case; would you want me to change all of them to camel case?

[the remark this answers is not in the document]
jlost: Nah, the models should stay how they are - they're reflective of the JSON that is returned by the API, and the API uses snake_case because it's written in Python. (a typescript `interface` just hints to typescript how it expects a data structure to look for type checking purposes)

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const [doRefresh, refreshLoading, refreshError] = usePromise(() => api.getMyUser(user_id) as Promise<User>, setUser);
+    useEffect(() => {
+        doRefresh();
+    }, []);
+    const [interactions] = useAutoRefresh(doRefresh);
+
+    //Setup interactions
+    const updateInfo = async (phone_number: string) => {
+        interactions.next(true);
+        await api.updateMyUser(user_id, phone_number)
```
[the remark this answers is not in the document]
bruinc: Would `return await doRefresh()` do that as well?

bruinc:
@jlost I think I've addressed all of your comments so far.

src/officehours_api/migrations/0015_profile_phone_number.py
```diff
@@ -0,0 +1,41 @@
...
+        migrations.AlterField(
+            model_name='attendee',
+            name='user',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AlterField(
+            model_name='meeting',
+            name='attendees',
+            field=models.ManyToManyField(through='officehours_api.Attendee', to=settings.AUTH_USER_MODEL),
+        ),
```
jlost, line 40: Any idea why these `AlterField`s were generated or what they're doing? It looks like `attendee.user` and `meeting.attendees` have been changed to link to `Profile` instead of `User`? But I don't see any major changes in `models.py`.

src/officehours_api/serializers.py
```diff
@@ -28,7 +29,24 @@ def get_my_queue(self, obj):
...
+        instance = super().update(instance, validated_data)
+        instance.profile.phone_number = profile.get('phone_number', instance.profile.phone_number)
+        instance.profile.save()            
+        return instance
+    
+    
+
+class ProfileSerializer(serializers.ModelSerializer):
+    user = NestedUserSerializer(required=True)
+    #phone_number = serializers.CharField(source='pk')
```
jlost, line 45: rm

src/officehours_api/nested_serializers.py
```diff
@@ -1,12 +1,20 @@
...
         model = User
-        fields = ['id', 'username', 'first_name', 'last_name']
+        fields = ['id', 'username', 'first_name', 'last_name', 'phone_number']
+
+""" class NestedProfileSerializer(serializers.ModelSerializer):
+    user = NestedUserSerializer(required=True)
+
+    class Meta:
+        model = Profile
+        fields = ['user', 'phone_number'] """
```
jlost, line 17: rm (you can always find deleted code in your git history later. I see what you were thinking here, and I'm not sure yet which is the best way)

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const [doRefresh, refreshLoading, refreshError] = usePromise(() => api.getMyUser(userId) as Promise<User>, setUser);
+    useEffect(() => {
+        doRefresh();
+    }, []);
+
+    //Setup interactions
+    const updateInfo = async (phoneNumber: string) => {
+         
+        return await api.updateMyUser(userId, phoneNumber)
+
```
jlost, line 66: You should remove the extra spacing here

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { PageProps } from "./page";
+
+interface PreferencesEditorProps {
+    user: User;
+    disabled: boolean;
+    onUpdateInfo: (phoneNumber: string) => void;
+}
+
+function PreferencesEditor(props: PreferencesEditorProps) {
+    var currentPhoneField = ""
```
jlost, line 20: It looks like you're trying to persist state. Did this come from example code somewhere? It seems to be working when I run it, but I think a bug would manifest pretty quickly because of the (somewhat weird) way React handles state. If you look up the `useState` hook and see the way I'm using it in other components, that's the template that I think you should be following in this spot.

bruinc: This was the first way I thought of keeping track of what the user has typed in the phone number field without having to update the user's phone number with every key stroke. I will definitely look at how you've used the `useState` hook and adapt this.

src/officehours_api/migrations/0015_profile_phone_number.py
bruinc: I changed them to link to Profile while I was playing around with things, then changed them back, that's why there's four of those. Only the first migration in this file actually does anything. Could I just delete the four AlterFields since they cancel each other out?

bruinc:
@jlost I resolved your feedback on my code, and now I'm going to work on the usability considerations you mentioned.

bruinc:
@jlost Feel free to take a look at what I have so far for the usability considerations. I also have some questions and comments below.

**Form**
- Putting the phone number field inside the form somehow disabled the dropdown for choosing the country code, and I'm not sure why.
- Hitting the enter key in the phone number field does not save/submit the form. I think this might be because I don't have an explicit `<input/>` element, but I'm not sure how to use the input tag with the `react-phone-input-2` component.
- Regarding this default behavior, how should this scale up when there are multiple fields in the form? Would hitting enter in any field submit the whole form?

**Validation**
- The current validation behavior is checking for 11 characters in the phone number upon clicking the save button. If it's true, the phone number is updated, if not, the form is re-loaded, and the field is reset back to what it was at the last save. This validation scheme is not ideal because many country codes are more than one digit, and some countries' area codes are four digits. I'll poke around online to see if I can find any regular expressions or logic for phone number validation.

bruinc:
Update: the issue with the country code drop down being disabled has been fixed.

I added basic length validation for USA phone numbers. If the user made a change to the field and the selected country is the USA, then it only updates the phone number if the length is exactly 11. If the user made a change and the selected country is any other country, there is no validation and it submits. 

I also added the case if the user deletes everything in the field, it will update their phone number to be the empty string. I felt like users should be able to change their mind and completely delete their phone number.

@jlost Let me know if there are any changes to this procedure that you'd like me to make. Are there any fields that should be added to the Preferences page before this PR is merged? Or should any additions to the Preferences page beyond the phone number be separate issues?

src/officehours_api/serializers.py
```diff
@@ -28,7 +29,23 @@ def get_my_queue(self, obj):
 
         serializer = QueueAttendeeSerializer(meeting.queue, context={'request': self.context['request']})
         return serializer.data
+    
+    def update(self, instance, validated_data):
+        print("***SERIALIZER VALIDATED DATA:", validated_data)
+        profile = validated_data.pop('profile')
```
jlost, line 35: I take it `validated_data` is a dict here? I would typically prefer `validated_data['profile']` (AKA the `__getitem__` accessor) over `.pop(...)` because `pop` mutates the underlying array, which can cause confusion. I do see that `validated_data` is being passed into the `super().update(...)` after the pop, so maybe there's a reason to remove that key that I'm not seeing?

src/officehours_api/serializers.py
```diff
@@ -28,7 +29,23 @@ def get_my_queue(self, obj):
 
         serializer = QueueAttendeeSerializer(meeting.queue, context={'request': self.context['request']})
         return serializer.data
+    
+    def update(self, instance, validated_data):
+        print("***SERIALIZER VALIDATED DATA:", validated_data)
```
jlost, line 34: Looks like you left this in by mistake while debugging

src/assets/src/services/api.ts
```diff
@@ -167,6 +167,20 @@ export const getMyUser = async (user_id: number) => {
...
+export const updateMyUser = async (user_id: number, phone_number:string) => {
+    
+    const resp = await fetch(`/api/profiles/${user_id}/`, {
+        method: "PATCH",
+        headers: getPatchHeaders(),
+        body: JSON.stringify({
+            phone_number: phone_number,
+        }),
+    });
+    console.log(resp)
```
jlost, line 179: Debugging line to remove

src/officehours_api/serializers.py
bruinc: My understanding was that the User serializer doesn't need to package the whole profile model with it, it just needs the phone number. So I take the profile out of the validated data, update the user fields with the remaining validated data, then manually update the phone number field.

bruinc: In commit 103543a I changed the `pop()` to just accessing the profile from the validated data, and it seems to work exactly the same.

bruinc:
I'm going to merge 1.2 into this branch so it can  have the correct migrations. 

@jlost This branch has the new dependency `react-phone-input-2`, so whenever I go to or from this branch, I've been having to run `docker-compose build`. Will we have to do something special when merging into dev to make sure that it installs the new dependency?

jlost:
Everything should work fine after we merge into dev, essentially because deployments happen with a `docker build` which is almost identical to the way your local environment gets built.

bruinc:
@jlost Merged into dev!

PR #79, opened 2020-06-10
Bigger attendee turn number and "Joined Queue" UI changes

bruinc:


bruinc:
@jlost Just merged this into dev so you can take a look. Let me know if I missed something or if there's any other tweaks I should make before we move it to QA.

jlost:
I think this looks great, and it's ready to merge. I did notice one imperfection that you might want to touch up first though. In this spot, there's a lot more padding at the top of the card than at the bottom:

image

You can add `mt-0` to the `h5` like so to make them look more even:

image

PR #93, opened 2020-06-22
Show Remaining Character Count on Queue Description

bruinc:
The functionality is there, but some styling changes may be needed.


bruinc:
I also know I have a hard-coded "1000" in here, I'm going to get rid of that

src/officehours_ui/static/css/main.css
```diff
@@ -467,6 +467,12 @@ h3.alert-heading {
 	border: 1px solid #cecbc9;
 }
 
+.form-p {
```
jlost, line 470: It might be good to rename this something like `renaming-controls` to make it clear that's the group of elements it's intended for.

jlost:
The code looks good so far! De-hardcoding the max length by pulling in Django's max length might actually be pretty challenging, and maybe not worth the effort.

jlost:
If you haven't tried doing so yet, this might be a good time to test pushing your changes to dev. You could checkout dev on the command line, `git merge` this branch, and then push dev, which would add these changes to whatever is currently in dev.

bruinc:
This branch still has the old incorrectly squashed agenda migration, should I merge 1.2 into this branch before merging this branch into dev?

jlost:
Yeah, nice catch - you should merge 1.2 into this first then (or rebase this on 1.2). Otherwise the bad migrations would be applied and they would be hard to undo.

bruinc:
Rebasing didn't seem to add the migrations from 1.2, but merging worked. I just merged the changes into dev!

bruinc:
And the changes are there in dev.officehours.it.umich.edu.

jlost:
Nice, looks good! Feel free to merge if @teshsopher is good with it.

PR #101, opened 2020-07-07
Support multiple errors in ErrorDisplay component

bruinc:
Errors from different components/events are treated independently and listed out on a given page.


jlost:
I see what you're going for here, and it's an interesting approach. 

It took me a little while to wrap my head around because I found the typing `(string | Error | undefined)[][]` to be confusing -- one of the benefits of using static typing is that you can make your data structures very self-descriptive in a way that lends to readability and reduces the room for error. With that typing, I could have any permutation of variably-sized arrays with `string`, `Error` or `undefined` in any index, and using many of those permutations would cause bugs, but really it looks like you want a human-readable label that indicates "source of the error" to be paired with a single error. I'd suggest using an `interface` like this:
```
interface FormError {
    source: string;
    error: Error;
}

interface ErrorDisplayProps {
    errors: FormError[];
}
```
Then in the `ErrorDisplay` params, you could take `FormError[]`. That design would allow you to show multiple errors from a single source, which might make sense in some circumstances. If you wanted to ensure one error per source rather than allowing multiple, you could define a dict where the key is the source and the value is the error:
```
interface FormError {
    [source: string]: Error;
}

interface ErrorDisplayProps {
    errors: FormError[];
}
```
Or here's a shorter equivalent (inlined):
```
interface ErrorDisplayProps {
    errors: {[source: string]: Error};
}
```

bruinc:
@jlost I followed your suggestion, but I had to allow the error in `FormError` to be undefined, as the errors in the `errorTypes` list can be undefined (they all are upon page load). I could add a second interface that forces the error to be defined, and use it in the `ErrorDisplay` component because the error display isn't created until after the undefined errors are filtered out of the `errorTypes` list using the `checkError` function.

jlost:
I think you could do something like this: 
```
    const errorSources = [
        {source: 'Refresh', error: refreshError},
        {source: 'Refresh Users', error: refreshUsersError}, 
        {source: 'Remove Host', error: removeHostError}, 
        {source: 'Add Host', error: addHostError}, 
        {source: 'Remove Meeting', error: removeMeetingError}, 
        {source: 'Add Meeting', error: addMeetingError}, 
        {source: 'Queue Name', error: changeNameError}, 
        {source: 'Queue Description', error: changeDescriptionError}, 
        {source: 'Delete Queue', error: removeQueueError}, 
        {source: 'Queue Status', error: setStatusError}, 
        {source: 'Assignee', error: changeAssigneeError}
    ].filter(e => e.error) as FormError[];
    const loginDialogVisible = errorSources.some(checkForbiddenError);
    const loadingDisplay = <LoadingDisplay loading={isLoading}/>
    const errorDisplay = <ErrorDisplay errors={errorSources}/>
```
Allowing you to pare down the bit in common.tsx to:
```
export interface FormError {
    source: string;
    error: Error;
}

export const checkForbiddenError = (pair: FormError) => {
    return (pair.error.name === "ForbiddenError");
}

interface ErrorDisplayProps {
    errors: FormError[];
}
```

The `as FormError[]` is a type assertion, which is sort of like casting but instead of actually changing the value it's just telling the the transpiler "trust me bro, it's a `FormError[]`." They can be misused, but here I think it would be okay since it should be obvious to any reader that the preceding `filter` will make it compliant to `FormError[]`. (in a perfect world TypeScript could figure this out on its own, but alas)

jlost:
In that bit, I also renamed `errorTypes` to `errorSources` since I thought that was a bit more accurate - `formErrors` would be a good naming too.

bruinc:
@jlost Can you explain why the `StatelessSingleInputForm` handles displaying an error on its own? Currently in this branch I commented out the `ErrorDisplay` in the `StatelessSingleInputForm` (and `ShowRemainingField`) component because it doesn't comply with the new `ErrorDisplay`, and the errors seem to correctly propagate to the page and get displayed by the page.

jlost:
@bruinc I was probably thinking that it might make sense to show certain validation errors right next to the `SingleInputForm` they originated from, and then never got around to actually using that. If it looks like nothing uses it then it can be deleted. We can always add it back in later.

bruinc:
@jlost I took out the error handling inside the input forms. Do you have any more feedback for this issue?

PR #105, opened 2020-07-15
Hide Phone Field in Join Info if Nonexistent

bruinc:


src/assets/src/components/edit.tsx
```diff
@@ -265,10 +265,17 @@ interface MeetingInfoProps {
 }
 
 const MeetingInfoDialog = (props: MeetingInfoProps) => {
+    const phoneDetails = (phoneNumber: string) => {
+        if (phoneNumber === "") {
+            return undefined;
+        } else {
+            return (<><strong>  Phone:</strong><PhoneInput value={phoneNumber} disabled={true}/></>);
+        }
+    }
```
jlost, line 274: Is there any need for this to be a function? It looks like the only value that would ever be used here is `a.phone_number`. So removing the function wrapper, you could have:
```
let phoneDetails;
if (a.phone_number === "") {
    phoneDetails = undefined;
} else {
    phoneDetails = (<> (...)value={a.phone_number}(...)<>);
}

(...)

<UserDisplay user={a}/>
{phoneDetails}
```

Condensing further:
```
const phoneDetails = a.phone_number && (<><strong>  Phone:</strong><PhoneInput value={a.phone_number} disabled={true}/></>);

(...)

<UserDisplay user={a}/>
{phoneDetails}
```

bruinc: @jlost I agree with you. I made that function just as separation to avoid javascript inside html inside javascript, but I could've done that with just a variable instead of a function.

bruinc: @jlost I remember another reason why I did this, it was because the attendee details is mapping all attendees, so there may be more than one phone number. For each phone number, I call that function to decide if it should be rendered or not. So if I wanted to condense it, it would have to be inside the attendee details mapping:
`{a.phone_number && (<><strong> Phone:</strong><PhoneInput value={a.phone_number} disabled={true}/></>)}`
I can do this if you'd like, but it is JS inside HTML inside JS inside HTML.

bruinc:
@jlost This will get overwritten by #112 since we aren't going to show phone numbers in the Join Info dialog at all, correct? Should we still merge this now? Or should it be tagged to a different milestone?

jlost:
You're right, but it seems like we might as well merge it. Even if we remove the code soon after, we could "revert" the removal later on to get the code back.

PR #106, opened 2020-07-17
Meeting Type Options

bruinc:


jlost:
This touches on an area that I've been thinking about lately, and I had some ideas in mind for factoring. Instead of typing them out, I thought it would be easier to add a few commits for you to look at. I tried to shift the backends system toward being more data-driven by pulling them from `settings`. The key and friendly_name of the backends are also sent to the client (accessible in `Globals`), which could be passed into `AllowedMeetingTypesForm` and other frontend components. Let me know what you think.

jlost:
Expanding on that a little, using `Globals.backends` in `AllowedMeetingTypesForm` would help the frontend in two ways:
* Duplication in `AllowedMeetingTypesForm` (and possibly elsewhere) could be reduced
* In situations where certain backends aren't available, such as running in development mode where `bluejeans` is disabled, the UI will work as expected (not let you select disabled backends)

bruinc:
I think this reorganization is great! I can work on cleaning up the frontend next.

jlost:
Cool, I also have some very vague ideas relating to how "[backend_type] allowed" is stored - in short, I would like to avoid putting backend details directly into the database schema if possible. This might mean storing a list of `backend_type` strings as `allowed_backends` or something like that. Postgres (/Django) does support array-type columns(/fields), so it should be doable.

bruinc:
@jlost @ssciolla Please feel free to look at what I have currently. The only remaining bug is in edit.tsx in `AllowedMeetingTypesForm`. The checkboxes are dynamically created based on the global backends. The underlying state of the checkbox seems to be toggling correctly when a checkbox is clicked, but it doesn't change in the UI unless you do something to update the page like change the queue name.

src/assets/src/components/edit.tsx
```diff
@@ -93,8 +96,118 @@ function HostEditor(props: HostEditorProps) {
...
+            checkboxes.push(checkbox(thisType));
+        } else {
+            let thisBox = <>
+                <div className="input-group-append">
+                    {checkbox(thisType)}
+                </div> 
+            </>
+            checkboxes.push(thisBox);
+        }
+    }
```
jlost, line 195: This for block could be simplified quite a bit:
```
const checkboxes = Object.keys(props.backends).map((b, i) =>
    i === 0 ? checkbox(b) : <div className="input-group-append">{checkbox(b)}</div>
);
```
Using map/filter is generally better than a for loop when you're only transforming or filtering a set of values, because it makes your intent crystal clear and limits the scope of mistakes that could be made within the block.

What is the `i === 0` check for, though? Seeing if it's some default value?

jlost:
Also looks like a migration needs to be generated.

src/assets/src/components/common.tsx
```diff
@@ -270,6 +270,53 @@ export const EditToggleField: React.FC<EditToggleFieldProps> = (props) => {
...
+];
+
+export const convertMeetingTypeToDisplay = (backendType: string) => {
+    if (backendType === "bluejeans") {
+        return ("BlueJeans");
+    }
+    else if (backendType === "inperson") {
+        return ("In Person");
+    }
+}
```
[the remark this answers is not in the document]
bruinc: It wasn't being used, forgot to delete

jlost:
And don't sweat it if you don't have time to finish this before your internship ends! I can finish it up.

bruinc:
> Also looks like a migration needs to be generated.

There's one: 0018_backend_type_options.py

jlost:
I ran `makemigrations`, which created another:
```
# Generated by Django 3.0.8 on 2020-08-05 18:10

import django.contrib.postgres.fields
from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('officehours_api', '0019_auto_20200729_1506'),
    ]

    operations = [
        migrations.AlterField(
            model_name='meeting',
            name='backend_type',
            field=models.CharField(choices=[('inperson', 'InPerson'), ('zoom', 'Zoom'), ('bluejeans', 'BlueJeans')], default='zoom', max_length=20),
        ),
        migrations.AlterField(
            model_name='queue',
            name='allowed_backends',
            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(choices=[('inperson', 'InPerson'), ('zoom', 'Zoom'), ('bluejeans', 'BlueJeans')], max_length=20), default=['zoom'], size=2),
        ),
    ]

```

jlost:
I see now that's  the same as what you already have, so I must have done something wrong. I probably need to delete my database volume and start from scratch.

bruinc:
Yours has zoom, mine does not. That's why you're getting another migration. I didn't want to add a zoom option in before there was a zoom backend.

jlost:
Yes, that was an artifact of changes on my own branch.

ssciolla:
Sorry, @bruinc, I've been consumed by some other project work and haven't had time to review. I will help push this forward, probably next week.

bruinc:
@jlost I resolved what I could of your feedback! The checkbox UI not updating is still an issue that I'm stumped on, so I will be interested to hear the cause of it if you figure it out (even I don't come back soon).

jlost:
@ssciolla This is ready for a review if you have some time.
@bruinc Feel free to have a look if you're curious about where I went with this - I resolved the checkbox issue by always changing and referencing the state instead of the DOM (unidirectional flow) and I also simplified the state by calculating whether it's checked by comparing the allowed backends to the available backends.

src/assets/src/components/common.tsx
```diff
@@ -266,6 +265,32 @@ export const EditToggleField: React.FC<EditToggleFieldProps> = (props) => {
...
+    const dropdownOptions = props.options.map(a => 
+        <option key={a.value} value={a.value}>{a.displayValue}</option>
+    );
+    const handleChange = (event: React.FormEvent<HTMLSelectElement>) => {
+        props.onChange(event.currentTarget.value);
+    }
+    return (
+        <select className="btn btn-sm select-dropdown" onChange={handleChange} value={props.selectedBackend}>
+            {dropdownOptions}
+        </select>
```
ssciolla, line 290: https://react-bootstrap.github.io/components/dropdowns/#single-button-dropdowns

src/assets/src/components/edit.tsx
```diff
@@ -190,6 +269,17 @@ function QueueEditor(props: QueueEditorProps) {
...
                 </div>
+                <div className="form-group row">
+                    <label htmlFor="allowed meeting types" className="col-md-2 col-form-label">Allowed Meeting Types:</label>
+                    <div className="col-md-6">
+                        <AllowedBackendsForm 
+                            queue={props.queue}
+                            allowed={new Set(props.queue.allowed_backends)}
+                            onChange={props.onUpdateAllowedBackends}
+                            disabled={props.disabled}
+                            backends={props.backends}/>
```
ssciolla, line 280: Not absolutely sure this is where I'm talking about, but I noticed when BlueJeans isn't configured (and I guess this would be the case whenever you have one backend), the `Checkbox` is still click-able. Might be nice to flip the `disabled` switch when this happens. Probably not a super common use case to have one backend, but I found it to be confusing. Similarly, if you try to un-click the last checked backend, it throws an error, which is somewhat hard to see since it's at the top of the page. Disabling the last checkbox and adding a note that you can't go down to zero might improve the interface.

ssciolla:
Also, @jlost, maybe this is out of scope for this PR, but I think I saw an error thrown when I ran the new code without starting with a fresh database volume. I'll have to revisit that to see if I can identify the problem, but that could be an issue depending on how we plan to release.

src/officehours_api/models.py
```diff
@@ -57,6 +52,15 @@ class Queue(SafeDeleteModel):
...
     )
+    MEETING_BACKEND_TYPES = [
+        (key, value.friendly_name)
+        for key, value in settings.BACKENDS.items()
+    ]
+    DEFAULT_ALLOWED_TYPES = [settings.DEFAULT_BACKEND]
+    allowed_backends = ArrayField(
+        models.CharField(max_length=20, choices=MEETING_BACKEND_TYPES, blank=False),
+        default=list(DEFAULT_ALLOWED_TYPES)
+    )
```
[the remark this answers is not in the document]
jlost: I had fixed that, but I think I lost it during a merge.  The solution was: `default=lambda: list(DEFAULT_ALLOWED_TYPES)` (return it from a lambda to make it callable)

src/assets/src/components/edit.tsx
jlost: I agree, but let's leave this one for a future release.

jlost:
I'll check that the migration will work by checking out master, migrating from a fresh volume, then checking out this branch and migrating again.

ssciolla:
> I'll check that the migration will work by checking out master, migrating from a fresh volume, then checking out this branch and migrating again.

I'd try to add some data (queues, meetings), too, after loading `1.2`

jlost:
No trouble migrating from 1.1.1 to 1.2. Data I input in 1.1.1 was preserved, and basic operations on it worked in 1.2

ssciolla:
> No trouble migrating from 1.1.1 to 1.2. Data I input in 1.1.1 was preserved, and basic operations on it worked in 1.2

Okay, I'll keep an eye on this and try to replicate it earlier. Could have been some other issue.

src/assets/src/components/edit.tsx
ssciolla: You want I should open a new issue?

ssciolla: https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/133

PR #110, opened 2020-07-24
Fix docker-compose for SELinux

jlost:
Adds `:z` to volume mounts to fix the development environment for systems that have SELinux enabled.


jlost:
Confirmed working on OS X, Windows (WSL2), Fedora 32 (moby)

Issue #111, opened 2020-07-24
Description field - Red Accents When Over Character Limit


teshsopher:
When the description field has more than 1000 characters, the field should be outlined in red and the character indicator should be red as well (e.g. if the character limit reads 1100/1000, the number should be red, and the field should be outlined in red).

This red should have a minimum contrast ratio of 4.5 with the white background.

teshsopher:
QA tested and working in dev on 8/12

PR #122, opened 2020-07-31
Update Manage description field form to indicate entry is over the limit (#111)

ssciolla:
This PR modifies the existing `SingleInputFormShowRemaining` component to make the form field outline and the color of the remaining characters text red when the user's entry is over the limit. The red in use is [the `danger` Bootstrap color](https://getbootstrap.com/docs/4.0/getting-started/theming/#theme-colors), which seems to have [just above a 4.5 contrast ratio with white](https://contrast-ratio.com/#rgb%28220%2C%2053%2C%2069%29-on-white). The "Change" button is disabled if the entry is over the limit, and I changed `main.css` to remove the default "X" from the description form when it is invalid. Additionally, I changed some code to use more React Bootstrap components, modified the remaining characters text for clarity, added some spaces after edit icons (following patterns seen elsewhere), and re-organized imports in `common.tsx`. Feedback on anything is welcome. The PR aims to resolve issue #111.

Example:
Screen Shot 2020-08-05 at 9 47 26 AM

### DIFF
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -293,6 +293,7 @@ export const SingleInputFormShowRemaining: React.FC<SingleInputFormShowRemaining
                     id={props.id}
                     as='textarea'
                     rows={5}
+                    bsPrefix='form-control form-control-remaining'
                     value={props.value}
                     placeholder={props.placeholder}
                     onChange={(e) => handleChange(e.currentTarget.value)}
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -1,12 +1,10 @@
-import * as React from "react";
+import * as React from "react"
+import { createRef, useEffect, useState } from "react"
+import { Link } from "react-router-dom"
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faSyncAlt, faClipboard, faClipboardCheck, faPencilAlt, faTrashAlt, faHome } from '@fortawesome/free-solid-svg-icons'
-import { User, QueueAttendee } from "../models";
-import { useState, createRef, useEffect } from "react";
-import { Link } from "react-router-dom";
-import Alert from "react-bootstrap/Alert";
-import Breadcrumb from "react-bootstrap/Breadcrumb";
-import Modal from "react-bootstrap/Modal";
+import { Alert, Breadcrumb, Button, Form, Modal } from "react-bootstrap"
+import { QueueAttendee, User } from "../models"
 
 type BootstrapButtonTypes = "info" | "warning" | "success" | "primary" | "alternate" | "danger";
 
@@ -261,8 +259,7 @@ export const EditToggleField: React.FC<EditToggleFieldProps> = (props) => {
             <div className="input-group">
                 <span>{props.text}</span>
                 <button onClick={enableEditMode} type="button" className="btn btn-sm">
-                    <FontAwesomeIcon icon={faPencilAlt} />
-                    Edit
+                    <FontAwesomeIcon icon={faPencilAlt} /> Edit
                 </button>
             </div>
         );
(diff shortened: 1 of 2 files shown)


src/assets/src/components/common.tsx
```diff
@@ -274,38 +271,47 @@ interface SingleInputFormShowRemainingProps extends StatelessSingleInputFormProp
...
-    const handleChange = (newValue:string) => {
+    const handleChange = (newValue: string) => {
         props.onChangeValue(newValue);
+        if (newValue.length <= props.maxLength) { setIsInvalid(false) } else { setIsInvalid(true) }
         setRemaining(props.maxLength - newValue.length)
     }
+
+    const charsRemaining = (remaining > 0) ? remaining : 0 as number
+    const charsOver = (remaining < 0) ? ` (${remaining * -1} over limit)` : '' as string
+    const feedbackText = `Remaining characters: ${charsRemaining}/${props.maxLength}` + charsOver as string
```
ssciolla, line 293: I'm happy to change this based on feedback from @mfldavidson or @teshsopher. I just found the use of only numbers and negatives to be a little jarring. This code might also be simplified a bit if I changed the `remaining` `useState` hook, but I opted to not mess with things even more.

src/assets/src/components/common.tsx
```diff
@@ -274,38 +271,47 @@ interface SingleInputFormShowRemainingProps extends StatelessSingleInputFormProp
...
-        props.onChangeValue("");
+    const [remaining, setRemaining] = useState(props.maxLength - props.value.length as number)
+    const [isInvalid, setIsInvalid] = useState(false as boolean)
+    const buttonClass = `btn btn-${props.buttonType} remaining-controls` as string
+
+    const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+        e.preventDefault()
+        if (!isInvalid) {
+            props.onSubmit(props.value)
+            props.onChangeValue('')
```
ssciolla, line 282: @jlost, would it be better to just disable the "Change" button if it's invalid?

ssciolla: Yeah, I'm just gonna do that, seems like a better user experience to me.

ssciolla:
> One thing I wanted to point out (and this may be Chrome-specific) is that the red "X" is cut off a little bit, regardless of window size/dimensions.

Thanks for the review and observation, @bruinc ! I think the overlap with the red "X" is a Windows issue, since it looks okay with the scrollbar on my work Mac. I could remove the "X" all together with an inline style. Would that be better?

Screen Shot 2020-07-31 at 4 04 04 PM

src/assets/src/components/common.tsx
```diff
@@ -274,38 +271,45 @@ interface SingleInputFormShowRemainingProps extends StatelessSingleInputFormProp
...
-        if (!props.autofocus) return;
-        inputRef.current!.focus();
-    }, []);
-    const submit = (e: React.FormEvent<HTMLFormElement>) => {
-        e.preventDefault();
-        props.onSubmit(props.value);
-        props.onChangeValue("");
+    const [remaining, setRemaining] = useState(props.maxLength - props.value.length as number)
+    const [isInvalid, setIsInvalid] = useState(false as boolean)
+    const buttonClass = `btn btn-${props.buttonType} remaining-controls` as string
```
jlost, line 276: No need to type-assert `as number`/`as boolean`/`as string` here as they can be inferred from the context.

src/assets/src/components/common.tsx
jlost: No need for the type assertions here, but otherwise I think it's good.

src/assets/src/components/common.tsx
```diff
@@ -274,38 +271,45 @@ interface SingleInputFormShowRemainingProps extends StatelessSingleInputFormProp
...
-    useEffect(() => {
-        if (!props.autofocus) return;
-        inputRef.current!.focus();
-    }, []);
-    const submit = (e: React.FormEvent<HTMLFormElement>) => {
-        e.preventDefault();
-        props.onSubmit(props.value);
-        props.onChangeValue("");
+    const [remaining, setRemaining] = useState(props.maxLength - props.value.length as number)
+    const [isInvalid, setIsInvalid] = useState(false as boolean)
```
jlost, line 275: I think `isValid` can be derived statelessly from `remaining`? And although I didn't notice it before, it looks like maybe there's no need for state `remaining` either, as that can be derived from `props.maxLength` and `props.value.length`. So I think these two can be replaced with:
```
const remaining = props.maxLength - props.value.length;
```
and
```
const isInvalid = remaining < 0
```
All of the state updates could then be removed.
Let me know if I'm overlooking something.

src/assets/src/components/common.tsx
```diff
@@ -274,38 +271,45 @@ interface SingleInputFormShowRemainingProps extends StatelessSingleInputFormProp
...
+            <Form.Group>
+                <Form.Control
+                    id={props.id}
+                    as='textarea'
+                    rows={5}
+                    value={props.value}
+                    placeholder={props.placeholder}
+                    onChange={(e) => handleChange(e.currentTarget.value)}
+                    disabled={props.disabled}
+                    isInvalid={!!isInvalid}
```
jlost, line 304: It's already a boolean, so no need to coerce (`!!`)

src/assets/src/components/common.tsx
ssciolla: Hm, good point, I'll take a look and simplify if possible. Thanks!

src/assets/src/components/common.tsx
ssciolla: Ah, thanks, I think this may have been a vestige from when I was I initializing as undefined. Certainly doesn't look needed any more, haha

src/assets/src/components/common.tsx
ssciolla: Sure, removed.

src/assets/src/components/common.tsx
ssciolla: Done.

src/assets/src/components/common.tsx
ssciolla: Done. Makes more sense to me now that I look back and see the higher-level `ShowRemainingField`.

src/assets/src/components/common.tsx
ssciolla: Removed the type assertions.

mfldavidson:
I think removing the X is a good idea. It looks like an overlap with the scroll bar Windows puts in. All in all this implementation looks great to me! Nice work!

ssciolla:
> I think removing the X is a good idea. It looks like an overlap with the scroll bar Windows puts in. All in all this implementation looks great to me! Nice work!

K, I'll remove the X. Thanks for the feedback!

ssciolla:
> I think removing the X is a good idea. It looks like an overlap with the scroll bar Windows puts in. All in all this implementation looks great to me! Nice work!

@mfldavidson, I did find a way to move the "X". Should we do that instead? While we're not using too much of this kind of form validation elsewhere yet, if we do, it might be nice to be consistent (either keep them all or remove them all). The overlap should only be a problem with the `textarea` fields that are large and need a scrollbar. I'm ready to push it far enough left out of the Windows scrollbar, or remove it, just let me know what you'd prefer, given that either is possible.

jlost:
I see similar overlap (Brave/Fedora/GTK3). I doubt most will hit the size limit, and if they do, I don't think any confusion resulting from a slightly overlapped "X" justifies fiddling with overriding Bootstrap CSS. I think it's fine as-is.

src/assets/src/components/common.tsx
ssciolla: Seems like I got the okay from the team on the text changes.

ssciolla:
@jlost, do you want to glance at the "X" removal changes, or should I just rebase/merge?

jlost:
Already glanced, looks good to me! Feel free to rebase & merge.

PR #131, opened 2020-08-13
Fix Queue.NotFoundError in QueueConsumer upon Queue Delete when Attendee Exists

### LINKED ISSUE #118: Show error if queue deleted while host or attendee is on page
Came up when testing #80 

If the queue is deleted while a host has the window open, that window refreshes to show a fully blank page with only the header nav and nothing else.

If the queue is deleted while an attendee has the window open and is in the line, that window refreshes to show the queue page as if the attendee is NOT in line, with an error at the top `Queue Connection: 1011`. Attempting to click "Join Queue" again results in error `Join Queue: Invalid pk "9" - object does not exist.`

Better UX for both IMO would be to show the `Queue Connection: The resource you're looking for could not be found. Maybe it was deleted?` error. @teshsopher can you weigh in?

jlost:
This happened because Queue.delete triggers Meeting.delete, which manually triggers the m2m signals via manual save. The signals trigger QueueConsumer to issue an update, which tries to query the queue which has already been deleted. There might be a more elegant fix than this (a way to remove the manual m2m signal trigger without breaking any functionality?), but this seems to get the job done without side-effects.

Closes #118

### DIFF
--- a/src/officehours_api/consumers.py
+++ b/src/officehours_api/consumers.py
@@ -71,7 +71,13 @@ class QueueConsumer(JsonWebsocketConsumer):
         )
 
     def queue_update(self, event):
-        queue = Queue.objects.get(pk=self.queue_id)
+        try:
+            queue = Queue.objects.get(pk=self.queue_id)
+        except Queue.DoesNotExist:
+            self.send_json({
+                'type': 'deleted',
+            })
+            return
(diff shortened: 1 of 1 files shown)


teshsopher:
Since #118 works in dev, I'm moving this to Done

PR #134, opened 2020-08-14
Remove Phone Number

### LINKED ISSUE #112: Temporarily remove user phone_number from API & edit page
Because anyone can be a host and any host can look up any user's phone_number, it's effectively public. Let's make it private by not returning it from the API and removing the "alternate contact info" display on the edit page for now.  If we need to return the phone_number for the edit profile page, the phone_number field can be made to only render if the user making the request is the same as the user being requested.

The field can still be used for notifications, and we may tighten the security around who is allowed to be a host in the future, which would allow us to add the "alternate contact info" display back in. Another way we could add it back in is by removing the ability to force-add a user to a queue from the edit page. Because the user will have added themselves, they will have effectively volunteered to display their contact information.

jlost:
Removes phone number from API & UI (also removes preferences page). Preserves it in the model.
Closes #112


jlost:
This will be reverted (at least partially) in the future when we need the preferences page again for notifications.

teshsopher:
Since #112 works, I'm moving this to Done

PR #140, opened 2020-08-18
Quick fix for UsersConsumer slowness

jlost:
Use `NestedUserSerializer` and disable/show spinner on `/edit` while `users` aren't yet loaded

### DIFF
--- a/src/officehours_api/consumers.py
+++ b/src/officehours_api/consumers.py
@@ -13,7 +13,8 @@ from safedelete.signals import post_softdelete
 from officehours_api.models import Queue, Meeting, Profile
 from officehours_api.permissions import is_host
 from officehours_api.serializers import (
-    QueueHostSerializer, QueueAttendeeSerializer, UserSerializer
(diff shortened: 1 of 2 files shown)


src/officehours_api/consumers.py
```diff
@@ -154,7 +155,7 @@ def user(self):
 
     def _get_users(self):
         return list(
-            UserSerializer(u, context={'user': self.user}).data
+            NestedUserSerializer(u).data
             for u in User.objects.all()
```
ssciolla, line 159: So, just to make sure I understand, this is faster because it's bypassing the queue-related data fetching that is in `UserSerializer` (and not in `NestedUserSerializer`)?

jlost: Yeah, looking up the queue data in `UserSerializer` is slow can could probably be optimized. But nothing that uses this consumer needs that data anyway. I may rename `NestedUserSerializer` to `ShallowUserSerializer` or similar now that it's no longer 'just' for nesting within other serializers, but not for this patch.

Issue #108, opened 2020-07-23
Agenda Field - Edit State Changes


teshsopher:
If there is an error saving the agenda (e.g. the attendee has entered an agenda with more than 100 characters), the error should display as normal, but the Agenda field should be in the edit state.

Additionally, if there is an agenda already when the page loads, the field should **not** be in the edit state.

teshsopher:
@bruinc If possible, it would also be good to indicate the character limit like you did for the Description field on the Host's side.

mfldavidson:
Was this resolved by your change @ssciolla ?

ssciolla:
I don't think so, but I'll check.

ssciolla:
@mfldavidson, are we still trying to get this in the release? I started work on this, so I could have it done today. Let me know what you think; I'm fine either way.

mfldavidson:
Tested and working in QA

PR #143, opened 2020-08-19
Refactor EditToggleField and stateless form components, consolidate client-side validation (#108)

ssciolla:
This PR refactors a number of components related to input forms in order to assist with client-side length validation and submit-button disabling, when desired.  The result is multiple new components, `StatelessInputGroupForm` and `StatelessTextAreaForm` that create fields of different styles with length validation.  I also re-worked and re-purposed `EditToggleField`, `SingleInputForm`, and `ShowRemainingField` to try to avoid duplication. Validation code was also simplified by leveraging the [`yup`](https://github.com/jquense/yup) library and consolidated into a new module, `validation.ts`. The changes expand upon work started in PR #122. The PR aims to resolve issue #108.

A couple notes:
* I did not yet pursue @jlost 's suggestion that we look into consuming server-side errors and co-locating them with the fields. I think some of these changes might make that easier to do in the future, but this PR is already getting a bit chunky. Doing this will require more research and thinking.
* There is additional custom validation code for phone numbers in `preferences.tsx`, but that component works a little differently, and we may need to replace the phone number field in the future. I'm kicking that can down the road.


src/assets/src/components/common.tsx
```diff
@@ -350,21 +389,28 @@ export const ShowRemainingField: React.FC<ShowRemainingFieldProps> = (props) =>
...
+        buttonType: props.buttonType,
+        disabled: props.disabled,
+    };
+
+    let statelessComponent = null;
+    if (props.fieldType === EditToggleFieldType.textArea) {
+        statelessComponent = <StatelessTextAreaForm {...statelessProps}>{props.children}</StatelessTextAreaForm>;
+    } else if (props.fieldType === EditToggleFieldType.inputGroup) {
+        statelessComponent = <StatelessInputGroupForm {...statelessProps}>{props.children}</StatelessInputGroupForm>;
+    }
```
ssciolla, line 410: I imagine there is a better way to do this, but I haven't found it yet. Still, this seems preferable over duplicating the editing `useState` hooks and conditional "Edit" button in `EditToggleField` and `ShowRemainingField`.

src/assets/src/components/common.tsx
```diff
@@ -326,19 +349,35 @@ export const SingleInputFormShowRemaining: React.FC<SingleInputFormShowRemaining
...
+    text: string;
+    placeholder: string;
+    disabled: boolean;
+    buttonType: BootstrapButtonTypes;
+    id: string;
+    onSubmit: (value: string) => void;
+    initialState: boolean;
+    validateLength: boolean;
+    maxLength?: number;
+    allowBlank?: boolean;
```
ssciolla, line 377: I suppose I could tweak `allowBlank` to be `minLength`, maybe set us up for uniqname validation.

src/assets/src/components/common.tsx
jlost: Hmm, what do you think of this?:
```typescript
interface EditToggleFieldProps {
    fieldComponent: React.FC<StatelessValidatedInputFormProps>;
    text: string;
    placeholder: string;
    disabled: boolean;
    buttonType: BootstrapButtonTypes;
    id: string;
    onSubmit: (value: string) => void;
    initialState: boolean;
    validateLength: boolean;
    maxLength?: number;
    allowBlank?: boolean;
}

export const EditToggleField: React.FC<EditToggleFieldProps> = (props) => {
    const [editing, setEditing] = useState(props.initialState);
    const [editorValue, setEditorValue] = useState(props.text);
    const submit = (value: string) => {
        props.onSubmit(value);
        setEditing(false);
    }
    const enableEditMode = () => {
        setEditing(true);
        setEditorValue(props.text);
    }

    const statelessProps: StatelessValidatedInputFormProps = {
        id: props.id,
        onSubmit: submit,
        value: editorValue,
        onChangeValue: setEditorValue,
        placeholder: props.placeholder,
        validateLength: props.validateLength,
        maxLength: props.maxLength,
        allowBlank: props.allowBlank ?? true,
        buttonType: props.buttonType,
        disabled: props.disabled,
    };
    const statelessComponent = <props.fieldComponent {...statelessProps} />
    return (editing && !props.disabled)
        ? (statelessComponent)
        : (
            <div className="input-group">
                <span>{props.text}</span>
                <button onClick={enableEditMode} type="button" className="btn btn-sm">
                    <FontAwesomeIcon icon={faPencilAlt} /> Edit
                </button>
            </div>
        );
}
```
This removes the enum in favor of passing in the component itself, and TS will validate that the type meets our requirements for props (`StatelessValidatedInputFormProps`)

src/assets/src/components/common.tsx
```diff
@@ -290,25 +245,93 @@ export const BackendSelector: React.FC<BackendSelectorProps> = (props) => {
...
+    const charsRemaining = (remaining > 0) ? remaining : 0;
+    if (!allowBlank && currentLength === 0) {
+       return 'This field is not allowed to be blank.';
+    } else {
+        const charsOver = (remaining < 0) ? ` (${remaining * -1} over limit)` : '';
+        return `Remaining characters: ${charsRemaining}/${maxLength}${charsOver}`;
+    }
+}
+
+export const StatelessInputGroupForm: React.FC<StatelessValidatedInputFormProps> = (props) => {
```
jlost, line 271: It looks like the point here is to provide a wrapper for a single arbitrary input element and any associated feedback? It might get complicated if we want to have a select or a type-ahead instead of a regular input (thinking ahead to replacing uniqname inputs with typeahead widgets).

src/assets/src/components/common.tsx
```diff
@@ -290,25 +245,93 @@ export const BackendSelector: React.FC<BackendSelectorProps> = (props) => {
...
+    if (props.validateLength && props.maxLength !== undefined) {
+        isInvalid = props.value.length > props.maxLength || (!props.allowBlank && props.value.length === 0);
+        const feedbackText = generateFeedbackText(props.maxLength, props.value.length, props.allowBlank);
+        const textClass = isInvalid ? 'text-danger': '';
+        feedbackEl = (
+            <Form.Text bsPrefix={`form-text remaining-feedback ${textClass}`}>
+                {feedbackText}
+            </Form.Text>
+        );
+    }
```
jlost, line 285: Feels like there's a little room for improvement here... Maybe validation logic could be factored out of here or made more modular somehow (Passing in a validate function? Passing in some kind of validationOptions object?).

jlost:
Looks promising so far!

src/assets/src/components/common.tsx
ssciolla: Good idea! I like it, I'll make the change.

jlost:
It might be worth thinking about server-side errors as well as client-side -- so if there is an unexpected error, or a validation we coudln't or didn't want to duplicate on the client-side, it still shows up next to the associated control.

src/assets/src/components/common.tsx
ssciolla: Done!

src/assets/src/components/common.tsx
ssciolla: I think the idea is that the single input form is a common UI feature of the application. I agree that those future field types wouldn't be easily handled by this, but I think there are ways we continue to modularize over time. I think this is at least a progressive improvement?

src/assets/src/components/common.tsx
ssciolla: You are correct, I need to refactor the validation work here. I'm looking into using [`yup`](https://github.com/jquense/yup#string) for validation (which was used in a React Bootstrap example), which I think might be a nice way to make this more customizable. I'll update once I've had more time to experiment.

ssciolla:
> It might be worth thinking about server-side errors as well as client-side -- so if there is an unexpected error, or a validation we couldn't or didn't want to duplicate on the client-side, it still shows up next to the associated control.

Hm, nice idea. Once I've refactored the validation, I'll see what it would take to do something like this.

src/assets/src/components/common.tsx
ssciolla: I think the validation refactor will be a better way to make this more robust.

src/assets/src/components/common.tsx
ssciolla: Okay, I think I like how this has progressed. I moved validation logic into a new `validation.ts` file, and I've adopted `yup` to create schemas with error messages. I want to do some more testing on whether multiple feedback messages are possible, but it seems potentially useful to keep that option open even if `schema.validate` only ever returns one message.

src/assets/src/components/edit.tsx
```diff
@@ -105,26 +111,53 @@ function AddAttendeeForm(props: AddAttendeeFormProps) {
...
+                    isInvalid={styleAsInvalid}
+                />
+                <InputGroup.Append>
+                    <MeetingBackendSelector
+                        allowedBackends={props.allowedBackends}
+                        backends={props.backends}
+                        onChange={setSelectedBackend}
+                        selectedBackend={selectedBackend}
+                    />
+                </InputGroup.Append>
```
ssciolla, line 152: I refactored here to use the validation approach and React Bootstrap components. There's definitely duplication from `StatelessInputGroupForm`, but it seemed difficult or perhaps over-complicating to integrate an optional "append" block with arbitrary content and extra props as part of that component.

ssciolla:
Hi @jlost, okay, I put some more work into this, and I think it feels like a reasonable approach. When you get a chance, could you take a look and let me know if this feels to you like something worth adopting? In addition to the form refactoring, all fields except the search field (which I don't think needs it?) and phone field in `preferences.tsx` have client-side validation.

Note that 'edit'-toggleable fields behave differently from 'single input' (always visible) fields when zero-length strings or blanks are not allowed. The add button is always disabled, but I felt the user shouldn't see invalid styling or messages with single input fields -- this would be jarring and seem like a punishment when they haven't done anything. This is mainly relevant for the Uniqname fields.

src/assets/src/utils.ts
```diff
@@ -19,22 +20,13 @@ export const redirectToSearch = (term: string) => {
...
 }
 
-const invalidUniqnameMessage = (uniqname: string) =>
-    uniqname + " is not a valid user. Please make sure the uniqname is correct, and that they have logged onto Remote Office Hours Queue at least once."
-
-export const validateAndFetchUser = async (uniqname: string) => {
-    if (uniqname.length < 3) throw new Error("Uniqnames must be at least 3 characters long.");
-    if (uniqname.length > 8) throw new Error("Uniqnames must be at most 8 characters long.");
-    if (!uniqname.match(/^[a-z]+$/i)) throw new Error("Uniqnames cannot contain non-alphanumeric characters.");
+export const confirmUserExists = async (uniqname: string) => {
```
jlost, line 23: What do you think about moving this to `validate.ts`? You are *validating* the existence of a user.

jlost (approved):
I like the approach and the code looks good. My only reservation is that server-side errors show up differently than client-side, which could defy expectations and consistency. To that, I'd say to put it in dev, have tsopher review it, and see if it stands out to her as jarring. Regardless, I think it's worthwhile to move forward with this, as relocating relocating the server errors (if that's appropriate) is a separately solvable problem.

docker-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
       - DEBUG=True
       - DATABASE_URL=postgresql://admin:admin@database/postgres
       - FEEDBACK_EMAIL=office-hours-devs@umich.edu
+    env_file:
+      - .env
```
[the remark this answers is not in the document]
ssciolla: Wups, I did not mean to commit that.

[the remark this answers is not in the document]
ssciolla: Gone now.

src/assets/src/utils.ts
ssciolla: Yeah, it could live somewhere else. Sure, I'll move it there for now. I guess it just felt like a fundamentally different kind of validation (using an API endpoint and depending on dynamic data, rather than a deterministic function check).

ssciolla: I moved this to `validation`.

jlost (approved):

ssciolla:
Thanks for the review, @jlost! I'm going to squash and merge now.

PR #145, opened 2020-08-20
Websocket Reconnect

### LINKED ISSUE #142: Reconnect websockets following disconnections caused by different browsers
Many devices disconnect websockets automatically to save power when the tab is put in the background.  When the tab returns to the foreground, check sockets and reconnect them if necessary. 

https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

Open to other suggestions!

jlost:
Reconnect if the connection has been lost upon tab restore.
Resolves #142

### DIFF
--- a/src/assets/src/hooks/useWebSocket.ts
+++ b/src/assets/src/hooks/useWebSocket.ts
@@ -52,6 +52,7 @@ export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onD
             console.log("handleVisibilityChange");
             if (!document.hidden && ws.readyState === 3) {
                 console.log("Reconnecting...");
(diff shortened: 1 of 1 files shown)


teshsopher:
@jlost I'm still seeing errors in Safari (desktop), but only when clicking on links that load a new site in the same tab (i.e. the links in the Message of the Day). I get the typical "Queue Connection error" and "User Connection error" when I'm any page:
image

However, I'm specifically getting a "Queue Connection: Type" error when I use the back button on the page with the list of all my queues
image
This error disappears quickly, so it may not be an issue, but it's a strange behavior to me (mostly because I'm not sure what it means)

PR #151, opened 2020-09-02
Validate Uniqname via API Call

### LINKED ISSUE #141: Replace UsersConsumer with User Search API
Create a User Search API that returns users by partial uniqname. Use it in place of the UsersConsumer to check validity of uniqnames and get associated IDs. Later, it will power typeahead user lookup widgets.

The main impetus of this is the slowness and poor scalability of the existing UsersConsumer, which returns all users upon page load, blocking certain interactions.

jlost:
I solved this problem in a more minimal way than I originally described in the linked issue: instead of creating a "search users" API endpoint, I simply allowed the existing `/users/<id>` endpoint to also accept `/users/<uniqname>`, and used that to validate the existence of a uniqname and lookup the user ID.

Closes #141

### DIFF
--- a/src/assets/src/components/edit.tsx
+++ b/src/assets/src/components/edit.tsx
@@ -10,7 +10,7 @@ import * as api from "../services/api";
 import { User, QueueHost, Meeting, BluejeansMetadata, isQueueHost, QueueAttendee } from "../models";
 import { UserDisplay, RemoveButton, ErrorDisplay, FormError, checkForbiddenError, LoadingDisplay, SingleInputForm, DateDisplay, CopyField, EditToggleField, LoginDialog, Breadcrumbs, DateTimeDisplay, BlueJeansDialInMessage, ShowRemainingField, BackendSelector as MeetingBackendSelector, DropdownValue } from "./common";
 import { usePromise } from "../hooks/usePromise";
-import { redirectToLogin, sanitizeUniqname, validateAndFetchUserByUniqname, redirectToSearch } from "../utils";
+import { redirectToLogin, sanitizeUniqname, validateAndFetchUser, redirectToSearch } from "../utils";
 import { PageProps } from "./page";
 import { useQueueWebSocket } from "../services/sockets";
 
@@ -444,7 +444,7 @@ export function QueueEditorPage(props: PageProps<EditPageParams>) {
     }
     const addHost = async (uniqname: string) => {
         uniqname = sanitizeUniqname(uniqname);
-        const user = await validateAndFetchUserByUniqname(uniqname);
+        const user = await validateAndFetchUser(uniqname);
         recordQueueManagementEvent("Added Host");
         await api.addHost(queue!.id, user.id);
     }
@@ -459,7 +459,7 @@ export function QueueEditorPage(props: PageProps<EditPageParams>) {
     }
     const addMeeting = async (uniqname: string, backend: string) => {
         uniqname = sanitizeUniqname(uniqname);
-        const user = await validateAndFetchUserByUniqname(uniqname);
+        const user = await validateAndFetchUser(uniqname);
         recordQueueManagementEvent("Added Meeting");
         await api.addMeeting(queue!.id, user.id, backend);
     }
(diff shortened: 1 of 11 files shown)


mfldavidson:
I like this solution. Much simpler!!!

src/assets/src/services/api.ts
```diff
@@ -169,13 +180,19 @@ export const setStatus = async (queue_id: number, open: boolean) => {
...
-    return await resp.json() as MyUser;
+    return await resp.json() as User | MyUser;
 }
 
-export const updateMyUser = async (user_id: number, phone_number: string) => {
+export const getUserById = async (user_id: number) =>
+    getUser(user_id);
+
+export const getUserByUniqname = async (uniqname: string) =>
+    getUser(uniqname);
```
ssciolla, line 193: Nit-pick: are these necessary?

src/assets/src/utils.ts
```diff
@@ -18,10 +19,20 @@ export const redirectToSearch = (term: string) => {
     location.href = `/search/${term}/?redirected=true`;
 }
 
-export const validateUniqname = (uniqname: string) => {
+const invalidUniqnameMessage = (uniqname: string) =>
+    uniqname + " is not a valid user. Please make sure the uniqname is correct, and that they have logged onto Remote Office Hours Queue at least once."
+
+export const validateAndFetchUserByUniqname = async (uniqname: string) => {
```
ssciolla, line 25: I appreciate the move of `invalidUniqnameMessage` out of `common.tsx`. However, it feels a little like we're smushing stuff together that shouldn't necessarily be combined. The call of `getUserByUniqname` and handling of errors feels like it could be its own function in `api.ts` (the error handling could even get moved into `getUserByUniqname`). Also, I'm probably biased because of my previous work on #143, but the validation piece could be left separate from the fetching. I dunno, `validateAndFetchUserByUniqname` just feels like a mouthful, which says to me it may make sense conceptually to keep them apart.

All of that said, this should work, and it doesn't necessarily need to be changed now.

src/officehours_api/consumers.py
```diff
@@ -299,7 +239,8 @@ def trigger_user_deleted(sender, instance: User, **kwargs):
 @receiver(post_save, sender=Profile)
 @receiver(post_delete, sender=Profile)
 def trigger_user_update_for_profile(sender, instance: Profile, **kwargs):
-    transaction.on_commit(lambda: send_user_update(instance.user.id))
+    user_id = instance.user.id
+    transaction.on_commit(lambda: send_user_update(user_id))
```
ssciolla, line 243: Comment might be useful here to explain why the assignment is needed. Also, could you create an issue for the user deletion errors and link it to this PR? I could too if you want, but I think you understand better what the issue is/was.

ssciolla (approved):
Couple comments and a documentation request, but otherwise, looks good and seems to work for me!

src/assets/src/services/api.ts
jlost: I compressed these into `getUser` by renaming its first param to `id_or_username`. I habitually avoid `this_or_that` param names as they can scale poorly with number of options or number of parameters, but in this case I can't imagine more than these two.

src/assets/src/components/edit.tsx
```diff
@@ -4,16 +4,15 @@ import { Link } from "react-router-dom";
 import * as ReactGA from "react-ga";
 import { Modal, Button, Table, Alert, Form } from "react-bootstrap";
 import Dialog from "react-bootstrap-dialog";
-import PhoneInput from "react-phone-input-2";
```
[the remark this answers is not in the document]
jlost: If it's not doing anything, I'd rather not see it. It will be easy to remember to re-add this when the phone input is re-included, because the code won't compile without it.

src/assets/src/utils.ts
jlost: My goal here was to avoid putting validation bits in `api` which has previously been divorced from input validation, but I agree that this piece is a bit odd standing out in `utils` as the only such "validate and call API with some error handling" type of function. This can be revisited in your input validation PR.

Issue #71, opened 2020-06-01
More useful information in queue results list


jonespm:
In both the manage queue page and in the search page when you search for a queue many may come back with the same or similar names since no naming restrictions are in place. The only details in the list are the name.

I'd think for the search results some additional details like possibly created date, host and possibly the number of people in the queued could be presented
![image](https://user-images.githubusercontent.com/27447/83418797-28320000-a3f2-11ea-8dc5-19625f84a5d0.png)

Additionally when managing the queue you can create multiple with the same name. There should be a column at least for created date, and possibly the id of the queue.

mfldavidson:
We can definitely do created date and ID. Those are not as helpful as host but because you can have multiple hosts that brings in a new level of complexity...

mfldavidson:
I'm adding this to 1.2 but for now let's not approach host

mfldavidson:
What if we also added if it's open or not

mfldavidson:
Responsive design is nice, good padding even when small and queue names go onto multiple lines seamlessly

Issue #124, opened 2020-08-06
Use websockets on /manage & remove useAutoRefresh


jlost:
One page, `/manage`, still uses `useAutoRefresh`. Switching it over to using websocket/Channels will allow us to remove `useAutoRefresh` entirely.

Doing this would necessitate a means of subscribing to the queues for which a user has been added as a host. When the user is removed or added as host to a queue, or a queue for which they're a host has been deleted, the page should update. This could be approached in a number of ways...

* Add a `hosted_queues` property to `UserSerializer` and leverage the existing `UserConsumer`. This avoids creating another consumer, and the update would need to trigger off of m2m changes and queue deletions. 
* Add a new consumer that can return queues for a particular user (or only the logged in user?). In this case, the smaller scope of results would have the benefit of simplifying the wiring of signals to consumer updates.

It would be worth considering #71 (or doing it first) because if we want to include full `hosted_queues` data in the `UserSerializer`, that might have an impact on performance (especially for `UsersConsumer`) and the depth of signals that need to be wired to consumer updates (will need to update `UserConsumer` upon hosted queue changes, not just add/remove). This could be worked around by parameterizing `UserSerializer` (pass in whether to render queue details), but it would add some complexity.

jlost:
One other idea: We could just necessitate manual refreshes on this page by removing useAutoRefresh and otherwise leaving it as-is. I doubt users tend to linger on it or watch it for changes.

jlost:
This will also allow `rxjs` to be removed as a dependency.

mfldavidson:
Tested while having this page up:
- Added myself as a host to a queue (with another user)
- Removed myself as a host for a queue
- Added host again
- Opened queue
- Closed queue
- Deleted queue

All updates were super fast, no apparent issues. Nice work!!! 💯

PR #153, opened 2020-09-03
Update Manage Queues page using WebSockets, expand/refactor QueueList (#71, #124, #154)

ssciolla:
This PR makes changes across the stack to modify how the front-end updates the Manage Queues page (API polling -> User WebSocket connection); to re-format queue lists (on the Search and Manage Queues pages) and expand the data shown in them; and to implement minor layout and spacing fixes. More details are provided in a task list below. The PR aims to resolve issues #71, #124, and #154.

### Task List

- [x] Remove `useAutoRefresh`, its usage in `manage.tsx`, and the related explicit `rxjs` dependency from `package.json` (see issue #124).
- [x] Create `ShallowQueueSerializer` for serializing only basic queue information.
- [x] Add `hosted_queues` property to `UserSerializer` leveraging `ShallowQueueSerializer`.
- [x] Replace removal of refresh from `manage.tsx` with `useUserWebSocket` for receiving updates.
- [x] Update signal receivers to trigger updates to users when a hosted queue is created or deleted.
- [x] Change `QueueListSearch` view to leverage `ShallowQueueSerializer`.
- [x] Extract similar queue listing code from `search.tsx` and `manage.tsx`, and create `QueueList` in `common.tsx`, with updates to styling and data displayed (see #71).
- [x] Re-name `input-group` CSS class to `page-content-flow` and adjust margin direction (just top) and magnitude (`0.5rem`); this re-aligns content while keeping appropriate spacing (see #154).

### Screenshots
Manage Queues
Screen Shot 2020-09-11 at 9 55 15 AM

Search Results
Screen Shot 2020-09-11 at 9 55 59 AM


src/package.json
```diff
@@ -14,8 +14,7 @@
     "react-ga": "~2.7.0",
     "react-phone-input-2": "^2.13.5",
     "react-router-dom": "^5.0.0",
-    "reconnecting-websocket": "^4.4.0",
-    "rxjs": "^6.5.4"
+    "reconnecting-websocket": "^4.4.0"
```
jlost, line 17: I think the removal of rxjs will generate some lockfile changes when you run `npm i`, which should be committed.

src/officehours_ui/static/css/main.css
```diff
@@ -641,3 +641,7 @@ div[role=main] {
 .meeting-type-message {
 	padding-left: 5px
 }
+
+.queue-list-badge {
+	font-size: 90%;
+}
```
jlost, line 647: Noticed a few missing LFs at EOF.

src/assets/src/services/api.ts
```diff
@@ -190,7 +190,7 @@ export const updateMyUser = async (user_id: number, phone_number: string) => {
 export const searchQueue = async (term: string) => {
     const resp = await fetch(`/api/queues_search/?search=${term}`, { method: "GET" });
     await handleErrors(resp);
-    return await resp.json() as QueueAttendee[];
+    return await resp.json() as ReadonlyArray<QueueBase>;
```
jlost, line 193: I like the readonly usage - handy for keeping anyone from falling into React's traps.

src/officehours_api/consumers.py
```diff
@@ -289,6 +299,9 @@ def send_user_deleted(user_id: int, channel_layer=None):
 @receiver(post_save, sender=User)
 def trigger_user_update(sender, instance: User, **kwargs):
     transaction.on_commit(lambda: send_user_update(instance.id))
+    # Tried to also get hostnames to update on change in the edit queue page, but this didn't do it alone
+    # for queue in list(instance.queue_set.all()):
+    #     transaction.on_commit(lambda: send_queue_update(queue.id))
```
jlost, line 304: To make that work, it looks like we'd have to trigger updates to `QueueConsumer`s here. Alternatively, we could subscribe individually to every user websocket for which a page has information (costly). I don't think the benefit of doing it would be worth the cost of either fix, as users are essentially never updated, and even if they were, the updates wouldn't need to be seen immediately.

src/package.json
ssciolla: I tried that. It actually seemed to add content, not remove it (maybe those are from another change?). I think `rxjs` is a sub-dependency of a couple other libraries.

src/officehours_api/consumers.py
ssciolla: I agree, probably not worth it. I can just forget about that. Seems like we may want to handle a completely deleted user? Going to keep looking at these signal receivers.

src/officehours_ui/static/css/main.css
ssciolla: So you just wanted a return at the end of the file? Did that.

ssciolla: Oh, there are a couple others.

src/officehours_api/consumers.py
```diff
@@ -140,9 +142,17 @@ def trigger_queue_update_for_meeting(sender, instance: Meeting, **kwargs):
...
+    if action not in ["post_remove", "post_clear", "post_add"]:
+        return
+    if isinstance(instance, Queue):
         transaction.on_commit(lambda: send_queue_update(instance.id))
+        for host_id in pk_set:
+            transaction.on_commit(lambda: send_user_update(host_id))
+    else:  # is User
+        transaction.on_commit(lambda: send_user_update(instance.id))
+        for queue_id in pk_set:
+            transaction.on_commit(lambda: send_queue_update(queue_id))
```
ssciolla, line 155: I wasn't able to trigger this, but I was trying to follow the example from `trigger_user_update_for_meetings`. Do you think this is necessary?

src/officehours_api/consumers.py
ssciolla: So after looking at these receivers more, I think it's okay for now? I played around with getting hosts to disappear from edit queue pages when they are deleted, but that's actually kind of out-of-scope for this PR, and maybe not that important for some of the same reasons you outlined above? I think updates via WebSocket to "Manage Queues" when a `Host`-to-`Queue` relation is added or deleted seems sufficient for this issue, and that all should be working.

src/officehours_api/consumers.py
jlost: I was able to trigger it by making the m2m change from and saving the user (instead of the queue), from the console. It's not strictly necessary since that's a code path that isn't used anywhere in the app right now, but since it could come in handy later, I'd say let's leave it.

src/assets/src/components/common.tsx
```diff
@@ -473,3 +473,45 @@ export const Breadcrumbs = (props: BreadcrumbsProps) => {
...
+                    <Col lg={8} sm={8} xs={7}>{q.name}</Col>
+                    <Col lg={2} sm={2} xs={3}>
+                        <Badge className={badgeClass} variant={q.status === 'open' ? 'success' : 'danger'} pill={true}>
+                            {q.status}
+                        </Badge>
+                    </Col>
+                </Row>
+            </ListGroup.Item>
+        </Link>
+    ));
```
jlost, line 502: Semantically this feels more like a table than a list, but I'm fine with it as long as it causes no screen reader issues.

src/assets/src/components/common.tsx
```diff
@@ -473,3 +473,45 @@ export const Breadcrumbs = (props: BreadcrumbsProps) => {
...
+    queues: ReadonlyArray<QueueBase>;
+    manageLink?: boolean | undefined;
+}
+
+export function QueueList (props: QueueListProps) {
+    const linkBase = props.manageLink ? '/manage/' : '/queue/'
+    const badgeClass = 'queue-list-badge'
+
+    const queueItems = props.queues.map(q => (
+        <Link to={`${linkBase}${q.id}`}>
```
jlost, line 487: Adding a `key` to the top-level element will get rid of the warning in the console (`Each child in a list should have a unique "key" prop.`).

src/assets/src/components/common.tsx
ssciolla: Yes, I would agree. I didn't want to rock the boat too much, so I kinda just went with this. I could shift to a table if you want, probably wouldn't be too much work. My main accessibility concern with the current approach would be that the column headers aren't labeled for screen readers as such, as they would be in a table.

src/assets/src/components/common.tsx
ssciolla: Hm, must have missed that. Will take a look.

ssciolla: Well, there is one on `ListGroup.Item`, but I guess I failed to check for warnings after I moved the `Link` component out. I just moved the `key` up to `Link`; see https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/153/commits/56641a49b12ae8f1823900f9b6d6c8656a8a2dc9

src/assets/src/components/common.tsx
ssciolla: Thinking more about this. We do use tables elsewhere in the application, but this feels less like presenting tabular data and more about offering a choice (of which queue to navigate to). Shifting to a table might disrupt that feeling.

In terms of helping with screen-reading given the current implementation, seems like we can use `aria-labelled-by` for each `Col` with a value and reference the appropriate header, but this seems to result in all the headers being read at once and then the values, and then (at least on Mac's VoiceOver). That could be fine?

Alternatively, we could specify `aria-label` values for each of the `Col`s, which means we could have the screenreader say the field header before the value: For example:

```
                    <Col aria-label={`ID: ${q.id}`} lg={2} sm={2} xs={2}>
                        <Badge  className={badgeClass} variant='primary' pill={true}>{q.id}</Badge>
                    </Col>
```

jlost: @teshsopher Thoughts on reading headers first vs. repeating per value? Or any general thoughts on the semantics of table vs. list wrt accessibility?

teshsopher: In this case, the "Status" data only has meaning if it's associated with the individual queue. So, the screen reader would need to tell the user something along the lines of: "[Queue Name] Status: Open"

Both a list or a table can work for this, I think. However, if it's a table, it should work like a table with two headers (column headers are "Queue Name" and "Status," and the row headers would be the individual queue name). In this case, I'm not 100% sure which would be best. I'm inclined to say table since this is a pattern we've used, but the amount of data in the table would only be one column so a list may work just as well. In either case, we would need there to be an ARIA notification to tell users what information has changed when the table updates (e.g. if two queues close while the user is on this page)

ssciolla: Thanks, @teshsopher, I can draft a table version and reach out to you for feedback. I guess I was thinking there would be three columns, for each of the data elements.

With respect to ARIA notifications, is this something we are doing elsewhere in the application? I'll need to do some research on this, and I guess I wonder whether this can be made a separate issue?

src/officehours_api/consumers.py
```diff
@@ -124,11 +124,15 @@ def trigger_queue_update(sender, instance: Queue, created, **kwargs):
     if instance.deleted:
         return
     transaction.on_commit(lambda: send_queue_update(instance.id))
+    for host in instance.hosts.all():
+        transaction.on_commit(lambda: send_user_update(host.id))
```
ssciolla, line 128: @jlost, I just added this. Based on @teshsopher 's feedback, I realized that we probably should subscribe to queue name and status changes.

jlost (approved):
Looks good to me. I'm not expecting anything earth shattering in the table changes, so feel free to merge when you think it's ready.

src/assets/src/components/common.tsx
ssciolla: Okay, I've refactored to use a `table` approach. I will get this pushed out to `dev` for review. For the ARIA notifications on page updates, this seems important but something that should be resolved site wide and will require more research. I've created an issue for this: https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/163.

Issue #95, opened 2020-06-23
Add Queue workflow


mfldavidson:
Create a workflow for adding a new queue instead of the text entry on the Manage page. Refer to @teshsopher 's mockup for Add Queue Page https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592

1. [Add Queue text entry field is replaced by an Add Queue button on the Manage page](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/2vjp9J)

2. [Add Queue workflow page 1 has host enter Queue Name, Description (not required), and allowed meeting types](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/Mj74kz)

3. [Add Queue workflow page 2 has host enter other hosts if applicable](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/5GZKb8)

ssciolla:
I'd wait to fully QA this until issue #94 is also closed, as there will be some shared code, meaning I'll be making changes that touch these components.

teshsopher:
@ssciolla I did some testing with the Add Queue workflow. Everything seems to work as intended (errors throw as appropriate, queue is set up as expected, etc.). In terms of the accessibility of the Tab links that we discussed before, it may be possible to treat them as a sort of "sub-navigation." The current dev version looks like it handles the issue from before (where users couldn't click on the next section at all) well by allowing users to click on it and progress if they've filled out the fields as needed (and throwing the appropriate error if they haven't).

There are some other accessibility issues:
- The (Queue) Name field and the Description field in the General section and the "Add Host" field in the Manage Hosts section don't have labels (This is also related to issue #159 )
- There is contrast issues with the "Add Host" button (related to #155 ). Additionally, there is contrast issues with the "Cancel" button. I had intended this to be created like a link, so that it would be the "de-emphasized" action (so there aren't two buttons competing for attention, and to prevent errors, in case users mistake the "cancel" button to be the next action they should take), and the primary action ("Save") would be emphasized for users to know what to do next.

ssciolla:
[...]

Issue #94, opened 2020-06-23
Manage Queue settings page


mfldavidson:
Refer to @teshsopher 's Manage Queue Settings page mockups https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592--specific mockups linked below for each section

Move the following into a separate page for managing queue settings, [accessed by a Settings icon in the upper right of the queue management page](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/34dVnz)

[General](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/rvLzQE)
- Queue Name
- Queue Description (see #69 )
- Permitted Meeting Types (see #77 )
- Date Created
- Created By

[Manage Hosts](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/ndYrV4)
- Add, view, and remove hosts

[Delete Queue](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/QJgmkG)
- Delete queue

mfldavidson:
I tried really hard to break this and I could not! Nice work @ssciolla . I'd like to hear @teshsopher 's input from a UX/accessibility perspective before moving to Done, but I have a feeling it is good

teshsopher:
Tested and working in QA!

The only issue here is that the 'Discard Changes' button/link is just _slightly_ too light for the background (since the background is technically FFFEFE). For safety, I'd recommend changing the red to DA2F40 or something darker.

ssciolla:
@teshsopher, can you create a new issue regarding the red text contrast issue, please?

teshsopher:
@ssciolla I've created the new issue and moved this to done!

PR #169, opened 2020-10-01
Implement Add Queue workflow (#95)

ssciolla:
This PR implements the Add Queue Workflow designed by @teshsopher, which can be seen [here](https://www.sketch.com/s/88bccdc1-3ae0-4311-99a2-3dec82df2592/a/Mj74kz). Besides implementing the form composed of multiple forms, I modified the stateless form components in `common` so buttons and submit handlers are optional; I modified the `ErrorDisplay` so it results in multiple `Alert` components, which improves spacing and clarity (in my opinion); and I re-organized around some existing code, creating a `meetingType` module and moving `recordQueueManagementEvent` to `utils`. This PR builds upon work started by @bruinc. I will likely need feedback on a couple choices/remaining issues; I will make comments on the code changes where appropriate. All feedback is welcome; I expect some things will need to be changed. The PR aims to resolve issue #95.

**Note**: Some of the components created here can probably be re-used (to some extent) in the next phase of work on #94. I may refactor and move some code in `addQueue` at that point.


src/assets/src/components/addQueue.tsx
```diff
@@ -0,0 +1,388 @@
...
+// https://react-bootstrap.github.io/components/tabs/#tabs-custom-layout
+function AddQueueEditor(props: AddQueueEditorProps) {
+    return (
+        <Tab.Container id='add-queue-editor' defaultActiveKey='general' activeKey={props.activeKey} onSelect={props.onTabSelect}>
+            <Row>
+                <Col sm={3}>
+                    <Nav variant='pills' className='flex-column mt-5'>
+                        <Nav.Item><Nav.Link eventKey='general'>General</Nav.Link></Nav.Item>
+                        <Nav.Item><Nav.Link eventKey='hosts'>Manage Hosts</Nav.Link></Nav.Item>
+                    </Nav>
```
ssciolla, line 209: @teshsopher, I was looking at how to make both of the Tab links keyboard accessible, but for some reason, the standard implementation of these components disallows that. You have to change the default `role` property from `tab` to something else, which may also be a problem in itself. I was looking at this section and trying to figure out what to do; it doesn't sound like they expect all links to be in the Tab order? https://www.w3.org/TR/wai-aria-practices/#tabpanel

I also implemented the "Next" and "Back" buttons at the bottom, so there should be an accessible way to navigate this. Let me know if you have any thoughts about the Tab links.

src/assets/src/components/addQueue.tsx
```diff
@@ -0,0 +1,388 @@
...
+    const [hosts, setHosts] = useState([props.user] as User[]);
+
+    // Set up API interactions
+    const checkHost = async (uniqname: string): Promise<User> => {
+        return await confirmUserExists(uniqname);
+    }
+    const [doCheckHost, checkHostLoading, checkHostError] = usePromise(
+        checkHost,
+        (value: User) => setHosts([...hosts].concat(value))
+    );
```
ssciolla, line 273: @jlost, you'll notice that I've implemented this form so that we check for the existence of hosts when they are entered (using `doCheckHost`), but we subsequently actually create the hosts as part of the queue creation. I decided that it would simplify the implementation and make more sense to not create the queue, hosts as we go along, but this opens up the (tiny) possibility of a user being deleted after it was confirmed but before the queue was created. This results in an `Add Queue` error that's pretty obscure, referencing the pk of the nonexistent user. Do you think we need to/have any ideas on how might handle this case more elegantly?

src/assets/src/components/addQueue.tsx
```diff
@@ -0,0 +1,388 @@
...
+    };
+
+    // Improve this?
+    const handleManageHostsFinishClick = () => {
+        if (name !== '' && allowedMeetingTypes.size !== 0) {
+            doAddQueue(name, allowedMeetingTypes, description, hosts);
+        } else {
+            throw Error('Attempted to pass invalid data to API for queue creation')
+        }
+    };
```
ssciolla, line 342: This check could probably be cleaned up. I don't think it's possible because of validation (but could be wrong of course) to reach this point with invalid data (i.e. an empty string for a name and no selected meeting types).

jlost: Yeah, I could go either way on this. It will catch errors if we have a regression with validation in the future. OTOH, if code isn't used, I will generally remove it just to minimize cognitive overhead.

jlost (approved):
Looks good!

src/assets/src/components/addQueue.tsx
```diff
@@ -0,0 +1,398 @@
...
+                <CancelAddButton disabled={props.disabled} />
+            </div>
+        </div>
+    );
+}
+
+interface AddQueueEditorProps {
+    // Shared
+    disabled: boolean;
+    activeKey: string;
```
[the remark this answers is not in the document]
ssciolla: Thanks for bringing this up, I also wished that I could use a CV like this, but I couldn't get this to work with the `onSelect` prop on `Tab.Container`, which expects a string as a parameter and then would need to be passed to `setActiveKey`. Are you aware of a way I can make that work? I'll try again.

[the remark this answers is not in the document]
ssciolla: I guess I can do this and not touch the `onSelect` handler (`onTabSelect`), and things still work. So this should be fixed now.

src/assets/src/components/addQueue.tsx
ssciolla: I guess I'll leave this alone for now. I may return to it in the future.

src/assets/src/components/addQueue.tsx
ssciolla: Cool, I generally agree that the exception is quite rare. If possible, it seems like it might be nice to improve the error message. Is there a way to control that in the backend? Looking into it. At least indicating that the entity that doesn't exist is a host would be good.

ssciolla: @jlost, I'm going to create a new issue about this so we don't forget about it, but it can remain low priority or eventually end up as `wontfix`.

ssciolla: See #172.

PR #171, opened 2020-10-06
Notifications

### LINKED ISSUE #23: Notification on queue join in Edit page
"I wish that there was a notification sound (that I could turn off if I wanted to) when another person joined the queue."

Noted as high priority by 2 advisor beta testers

### LINKED ISSUE #25: Notification for student that it is their turn
We need a more "disruptive," if you will, way for the student to know that it is their turn. I know we had issues with browser notifications, but if there's any way we could do that or some other type of notification at the point of "there's only 5 people ahead" and then "it's your turn" but only if they aren't in the meeting yet, that would be very helpful.

Noted as high priority by 2 advisor beta testers

### LINKED ISSUE #174: makemigrations result varies based on settings.BACKENDS
`makemigrations` will generate changes if your local settings.BACKENDS doesn't have all of the backends enabled. It should always generate migrations that have all of the backends. The generated `default` values should point to a function if possible.

jlost:
Adds SMS notifications via Twilio:
* For attendees, when their meeting goes from >1st in line to being 1st in line. (Closes #25)
* For hosts, when their queue size goes from 0 to 1. (Closes #23)

Discovered and fixed a migration bug as migrations needed to be generated (Fixes #174), which led into noticing & doing some of the back-end setup for #175.


jlost:
Still want to check and possibly enhance error handling during meeting save / delete, but I think this is ready for review.

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+            || props.user.notify_me_attendee !== notifyMeAttendee
+            || props.user.notify_me_host !== notifyMeHost;
+        if (!changed) {
+            setValidationStatus(null);
+            return;
+        }
+        if (phoneField.length <= 1) {
+            // Update phone number to be empty if they try to delete everything in the phone field
+            // Seems to be a known issue where the last character can't be removed as part of onChange:
+            // https://github.com/bl00mber/react-phone-input-2/issues/231
```
jlost, line 80: According to this issue, it should be fixed and the fix should be in the latest release, but I was still getting the bug even after updating the package, doing a `docker-compose build`, etc. I might just need to clear volumes (`docker-compose down -v`), so I'll give that a shot before I give up.

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+type ValidationStatus = null | Error[]; // null = no changes, [] = valid
+
+function PreferencesEditor(props: PreferencesEditorProps) {
+    const [phoneField, setPhoneField] = useState(props.user.phone_number);
+    const [countryDialCode, setCountryDialCode] = useState("");
+    const [notifyMeAttendee, setNotifyMeAttendee] = useState(props.user.notify_me_attendee)
+    const [notifyMeHost, setNotifyMeHost] = useState(props.user.notify_me_host)
+    const [validationStatus, setValidationStatus] = useState(undefined as undefined | ValidationStatus);
```
jlost, line 35: I folded the previous "error message" state into this to tighten things up.

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { usePromise } from "../hooks/usePromise";
+import { redirectToLogin } from "../utils";
+import { PageProps } from "./page";
+
+const validatePhoneNumber = (phone: string, countryDialCode: string): Error[] => {
+    return [
+        (countryDialCode === '1' && phone.length !== 11)
+            && new Error("The phone number entered was invalid; USA phone numbers must have 11 digits."),
+    ].filter(e => e) as Error[];
+}
```
jlost, line 19: My thinking here was that most validations can return multiple errors, and it's useful to be able to see them all at once rather than one at a time, although we're only performing a single validation so far.

src/officehours_api/serializers.py
```diff
@@ -36,27 +36,12 @@ class Meta:
...
-        for i in range(0, len(meetings)):
-            if self.context['user'] in meetings[i].attendees.all():
-                in_line = True
-                break
-
-        if in_line:
-            return i
-        else:
-            return None
-
```
jlost, line 59: Moved this into the `Meeting` model (and refactored the contents)

src/assets/src/components/preferences.tsx
jlost: That did the trick, but the old behavior actually made sense, since a user could pick a `countryDialCode` (which autofills "1" etc.) and then not enter their phone number, and I think that state should be treated as "empty phoneField" for the purposes of this block. So I made some changes to submit without validating the phone number if they ONLY entered a `countryDialCode`, and to also leave the opt-ins disabled in that case until they enter more digits. I updated the library anyway and removed the comment.

src/officehours_api/notifications.py
```diff
@@ -0,0 +1,74 @@
...
+        logger.info('notify_next_in_line: %s', p)
+        domain = Site.objects.get_current().domain
+        queue_path = reverse('queue', kwargs={'queue_id': next_in_line.queue.id})
+        queue_url = f"https://{domain}{queue_path}"
+        twilio.messages.create(
+            to=p,
+            from_=settings.TWILIO_PHONE_FROM,
+            body=(
+                f"You're next in line! Please visit {queue_url} "
+                f"for instructions to join."
```
ssciolla, line 34:
```suggestion
                "for instructions to join."
```

src/officehours_api/notifications.py
```diff
@@ -0,0 +1,74 @@
...
+        domain = Site.objects.get_current().domain
+        edit_path = reverse('edit', kwargs={'queue_id': first.queue.id})
+        edit_url = f"https://{domain}{edit_path}"
+        twilio.messages.create(
+            to=p,
+            from_=settings.TWILIO_PHONE_FROM,
+            body=(
+                f"Someone has joined your queue, {first.queue.name}! "
+                f"Please visit {edit_url} "
+                f"to start a meeting."
```
ssciolla, line 57:
```suggestion
                "to start a meeting."
```

src/officehours_api/serializers.py
ssciolla: Seems like the right move.

src/officehours_api/serializers.py
```diff
@@ -109,11 +99,13 @@ def get_hosted_queues(self, obj):
             return []
         serializer = ShallowQueueSerializer(queues_qs, many=True)
         return serializer.data
-
+
```
ssciolla, line 102: Nit-pick: unnecessary whitespace?

src/officehours_api/tests.py
```diff
@@ -0,0 +1,146 @@
+from unittest import mock
+
+from django.test import TestCase, override_settings
```
ssciolla, line 3: Nice to see unit tests!

src/officehours_api/tests.py
```diff
@@ -0,0 +1,146 @@
...
+from officehours_api.models import User, Queue, Meeting
+
+
+@override_settings(TWILIO_ACCOUNT_SID='aaa', TWILIO_AUTH_TOKEN='bbb', TWILIO_PHONE_FROM='+15555552323')
+class NotificationTestCase(TestCase):
+    def setUp(self):
+        self.foo = User.objects.create(username='foo', email='foo@example.com')
+        self.foo.profile.phone_number = '+15555550000'
+        self.foo.profile.notify_me_attendee = True
+        self.foo.profile.notify_me_host = True
```
ssciolla, line 14: It feels like there should be a way to make this more concise, like maybe this?

```
self.foo.profile = Profile(phone_number='...', notify_me_attendee=True, notify_me_host=True)
self.foo.save()
```
or
```
some_profile = Profile(user=self.foo, phone_number='...', notify_me_attendee=True, notify_me_host=True)
```
Just an idea, I haven't tried it this with a 1:1 field before.

ssciolla (commented):
Hi @jlost, this is cool beans and I think it's all working for me. There are a lot of changes though. I'm not quite finished working through them, but I wanted to share my feedback so far. I left a bunch of questions, comments, and ideas; feel free to disregard them as you see fit.

I plan to fully finish the review tomorrow morning.

src/assets/src/components/preferences.tsx
ssciolla: Thanks! I think you're saying the code I wrote before made sense, not the prior library implementation? Anyways, I think you have is working well and it's good to update the lib and remove the comment.

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+        : props.errorOccurred
+            ? <Alert variant='danger'>An error occurred while trying to update your preferences; please try again later.</Alert>
+        : <Alert variant='success'>Your preferences were successfully updated.</Alert>
+
+    return (
+        <div>
+            <h1>View/Update Preferences</h1>
+            {alertBlock}
+            <Form onSubmit={validateAndSubmit}>
+                <p>Enter a phone number in order to opt in to SMS notifications.</p>
```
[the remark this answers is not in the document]
ssciolla: I guess you did wire this up so that if the phone number is blank, the checkboxes are enabled. Hm...

src/officehours_api/notifications.py
```diff
@@ -0,0 +1,74 @@
...
+        return
+    twilio = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
+    phone_numbers = list(
+        u.profile.phone_number for u in
+        next_in_line.attendees_with_phone_numbers.filter(profile__notify_me_attendee__exact=True)
+    )
+    for p in phone_numbers:
+        logger.info('notify_next_in_line: %s', p)
+        domain = Site.objects.get_current().domain
+        queue_path = reverse('queue', kwargs={'queue_id': next_in_line.queue.id})
```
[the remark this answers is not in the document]
ssciolla: I guess you'd maybe need an optional parameter to indicate whether you want the attendee (default?) or host editor view.

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+        />
+    );
+    const notifyMeAttendeeInput = (
+        <Form.Check 
+            type="checkbox"
+            id="notify-me-attendee"
+            disabled={!phoneField} 
+            checked={notifyMeAttendee} 
+            onChange={() => setNotifyMeAttendee(!notifyMeAttendee)}
+            label="As an attendee, notify me via SMS when it becomes my turn." />
```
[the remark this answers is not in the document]
jlost: Yeah, I think you're right, and both of your examples would be more correct.
@teshsopher @mfldavidson Any thoughts on rephrasing:
* "As an attendee, notify me via SMS when it becomes my turn."
* "As a host, notify me via SMS when someone joins my empty queue."

src/assets/src/components/preferences.tsx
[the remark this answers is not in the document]
jlost: @teshsopher What do you think of the UX around the phone number input, its label text, and the inputs below it that become enabled only when you input a phone number?

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+    
+    if (!props.user) throw new Error("user is undefined!");
+    const userId = props.user.id
+
+    // Setup basic state
+    const [user, setUser] = useState(undefined as MyUser | undefined);
+    const [doRefresh, refreshLoading, refreshError] = usePromise(() => api.getUser(userId) as Promise<MyUser>, setUser);
+    useEffect(() => {
+        doRefresh();
+    }, []);
```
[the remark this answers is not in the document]
jlost: `useEffect` expects nothing to be returned (`() => void | undefined`), so I actually need the curly braces here so as not to implicitly return the `Promise<void>` that is returned by `doRefresh()`. Of course I could still shift around whitespace, but it wouldn't be a true one-liner.

src/officehours_api/tests.py
jlost: I think the backends would also be a high-value target for unit tests.

src/assets/src/components/preferences.tsx
jlost: Yeah, the library implementation did have a bug, but checking `phoneField <= 1` still made sense (to me at least) for users with a 1-digit `dialCode`.

src/assets/src/components/preferences.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (!props.user) {
+        redirectToLogin(props.loginUrl);
+    }
+    
+    if (!props.user) throw new Error("user is undefined!");
+    const userId = props.user.id
+
+    // Setup basic state
+    const [user, setUser] = useState(undefined as MyUser | undefined);
+    const [doRefresh, refreshLoading, refreshError] = usePromise(() => api.getUser(userId) as Promise<MyUser>, setUser);
```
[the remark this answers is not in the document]
jlost: Good catch - I added `refreshError` to `FormErrors`, but (on other pages at least) `isChanging` is for when you've submitted a change and are waiting for it to be acknowledged. And on other pages, we don't show the `LoadingDisplay` when we're awaiting initial page load. The original reason for that is because the auto-refresher would also trigger the `LoadingDisplay` and some users found it distracting, but that isn't a problem anymore, so it probably makes sense to show the spinner across the board upon page load now. For this PR though, I'm going to leave it consistent with other pages.

src/assets/src/components/preferences.tsx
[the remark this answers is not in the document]
ssciolla: Good point, I was really just thinking about white space.

src/assets/src/components/preferences.tsx
jlost: Good idea - moved it.

src/assets/src/components/preferences.tsx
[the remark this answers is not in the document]
ssciolla: Okay, that makes sense.

src/officehours_api/notifications.py
ssciolla: Yeah, this isn't a big deal, I just like to not include characters (like trailing commas, for instance) that aren't necessary. Feel free to leave this as is, reject the suggestions.

src/officehours_api/models.py
```diff
@@ -80,18 +101,18 @@ class Meeting(SafeDeleteModel):
...
-        (key, value.friendly_name)
-        for key, value in settings.BACKENDS.items()
-    ]
     backend_type = models.CharField(
         max_length=20,
-        choices=MEETING_BACKEND_TYPES,
+        choices=get_backend_types(),
         null=False,
-        default=settings.DEFAULT_BACKEND,
+        default=get_default_backend,
```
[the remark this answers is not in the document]
ssciolla: Sorry, no need to explain, I did a little research on this in the Django docs.

src/assets/src/components/preferences.tsx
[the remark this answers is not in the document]
teshsopher: I think that the label text and input field is great! For the checkboxes below it, there are a couple issues. A disabled state can be hard for users to understand in general, particularly if they haven't run into this pattern before. Additionally, the color contrast isn't high enough for accessibility purposes. One other thing I noticed is if I use the field and use any non-numerical input, while it isn't added to the field (which is what we want), it enables the checkboxes below it.

I would recommend having the checkboxes be active and pushing an error if they try to use them without filling in the field. I would also add some more space between the checkboxes, as well as between the checkboxes and the input field so these elements aren't so close together.

src/officehours_api/tests.py
jlost: Good idea, but I couldn't get that or a close variation of it (`= Profile(user_id=self.foo.id`, `= Profile.objects.create(`, ...) to work, I think owing to the fact the profile is autogenerated for the user post-save. Instead, I condensed it with a very 1:1 refactoring.

ssciolla: Yeah, that works, thanks for looking into it!

src/officehours_api/tests.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @mock.patch('officehours_api.notifications.TwilioClient')
+    def test_first_meeting_removal_doesnt_notify_second_in_line(self, MockTwilio: mock.MagicMock):
+        m1 = self.create_meeting([self.foo,])
+        self.create_meeting([self.bar,])
+        self.create_meeting([self.baz,])
+        MockTwilio.reset_mock()
+        m1.delete()
+        receivers = self.get_receivers(MockTwilio)
+        self.assertFalse('+15555550002' in receivers)
```
ssciolla, line 132: Would it make sense to add a test simulating how Twilio responds (and then testing how our application responds) if the phone number doesn't exist, or other exceptions? When my phone number was unverified, I was getting some weird behavior and Internal Server Errors that prevented me from loading a queue edit page.

README.md
```diff
@@ -11,7 +11,30 @@ docker-compose run --entrypoint="" web python manage.py createsuperuser
...
+### Notifications
+
+SMS notifications for hosts and attendees are provided via [Twilio](https://www.twilio.com/).
+
+To use or develop with SMS notifications, set `TWILIO_ACCOUNT_SID`, `TWILIO_AUTH_TOKEN`, and `TWILIO_PHONE_FROM` as environment variables. Twilio provides free trial accounts with limited credit. You can use special [test credentials](https://www.twilio.com/docs/iam/test-credentials) to not be charged. You can also test notifications via unit tests, where Twilio is mocked: 
+```
+docker-compose run web python manage.py test officehours_api.tests.NotificationTestCase
+```
+
+## Migrations
```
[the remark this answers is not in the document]
jlost: I think that makes sense, so I changed the organization as you suggested. Let me know if that isn't what you had in mind.

ssciolla (approved):
Nice job on the unit tests! I made a couple more comments. Looks like you're continuing to tweak and think about exception handling. I'm confident you'll handle anything left that's important and incorporate the feedback from @teshsopher and others, so I'm approving. If you want me to take a look at anything new you add, let me know.

src/officehours_api/tests.py
jlost: Yup, good thinking - I made some changes to how exceptions are handled (log the exception at error level with full stacktrace & continue) and was already looking into testing it.

README.md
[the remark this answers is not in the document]
ssciolla: Yep, looks good.

src/officehours_api/models.py
```diff
@@ -140,3 +171,6 @@ def post_save_user_signal_handler(sender, instance: User, created, **kwargs):
         instance.profile
     except User.profile.RelatedObjectDoesNotExist:
         instance.profile = Profile.objects.create(user=instance)
+
+
+import officehours_api.notifications
```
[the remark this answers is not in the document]
jlost: This makes sure the signals in `notifications` are registered whenever the models are loaded. It's effectively similar to pasting the content of `notifications` to the end of this file, which would be an alternative, but I liked this more organizationally. Another alternative would be to keep only the signals registrations here, which would reference imports from `notifications`, but since they felt tightly coupled I kept them in the same place. Let me know if you find a better alternative!

src/officehours_api/tests.py
jlost: Added a couple tests to make sure it writes the exception to an ERROR logger and continues in those cases.

src/assets/src/components/preferences.tsx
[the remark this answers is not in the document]
jlost: I implemented your suggestions, feel free to have a look.

PR #173, opened 2020-10-09
Implement Manage Queue settings page, extract queueEditors from addQueue into separate module (#94)

ssciolla:


src/assets/src/components/queueManager.tsx
```diff
@@ -0,0 +1,446 @@
+import * as React from "react";
+import { useState, createRef, ChangeEvent } from "react";
+import { Link } from "react-router-dom";
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCog } from "@fortawesome/free-solid-svg-icons";
+import { Alert, Button, Col, Form, InputGroup, Modal, Row, Table } from "react-bootstrap";
+import Dialog from "react-bootstrap-dialog";
+import 'react-phone-input-2/lib/style.css';
```
ssciolla, line 8: I changed the name of `edit.tsx` to `queueManager.tsx`, since that made more sense to me. Apparently that and all the changes I made was enough for GitHub to think it was another file (you'll see the deleted `edit.tsx` above).

src/assets/src/components/queueSettings.tsx
```diff
@@ -0,0 +1,315 @@
...
+        if (!q) {
+            setQueue(q);
+        } else if (isQueueHost(q)) {
+            setQueue(q);
+            setAuthError(undefined);
+        } else {
+            setQueue(undefined);
+            setAuthError(new Error("You are not a host of this queue. If you believe you are seeing this message in error, contact the queue host(s)."));
+        }
+    }
```
ssciolla, line 141: This is just repeated from `queueManager.tsx` (formerly `edit.tsx`). We discussed earlier that there might be a better way to do this. Let me know if you have any ideas!

jlost: Off the top of my head, this wrapping could be rolled into a hook (e.g. `useQueueAsHost`) that returns `queue`, `setQueueChecked`, and `authError`. There are probably even better (but more creative and/or change-heavy) solutions. If you want to leave it, I think it's fine for now.

src/assets/src/validation.ts
```diff
@@ -76,3 +77,59 @@ export function validateString (value: string, schema: StringSchema, showRemaini
...
+export function validateAndSetStringResult (
+    someValue: string,
+    schema: StringSchema,
+    resultSetter: StringValidationResultSetter,
+    showRemaining?: boolean
+): ValidationResult {
+    const validationResult = validateString(someValue, schema, !!showRemaining);
+    resultSetter(validationResult);
+    return validationResult;
+};
```
jlost, line 127: This is good, but you could also make this a hook, e.g. `useValidatedStringState`, that returns `[string, setAndValidateString, clear]` where `setAndValidateString` is something like this and `clear` is `setString(undefined)` (because I see you invoked that above).

src/assets/src/components/queueSettings.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+    const handleTabSelect = (eventKey: string) => {
+        if (Object.values(AvailableTabs).some(value => value === eventKey)) {
+            const typedKeyString = eventKey as keyof typeof AvailableTabs;
+            setActiveKey(AvailableTabs[typedKeyString]);
+        } else {
+            throw Error(`eventKey "${eventKey}" is not defined in enum AvailableTabs`);
+        }
+    };
```
jlost, line 257: I think you mentioned this before, but maybe we can just pare this down to:
```
const handleTabSelect = (eventKey: string) => {
    setActiveKey(eventKey as AvailableTabs);
};
```
if there's no serious risk of using a string that isn't present in the enum.

src/assets/src/components/addQueue.tsx
```diff
@@ -260,14 +124,14 @@ export function AddQueuePage(props: PageProps) {
     }
 
     // Set up page state
-    const [activeKey, setActiveKey] = useState('general' as availableTabs);
+    const [activeKey, setActiveKey] = useState(AvailableTabs.General as AvailableTabs);
     const [showCorrectGeneralMessage, setShowCorrectGeneralMessage] = useState(false);
```
jlost, line 128: I might be mistaken, but it seems like this could be a variable derived from some other state? Maybe something like `const showCorrectGeneralMessage = nameValidationResult && descriptValidationResult && allowedValidationResult`?

jlost (commented):
Looks great so far, and sorry it took me so long to get around to this. I'll give another once-over very soon as there was a lot to digest in a single sitting.

src/assets/src/components/addQueue.tsx
ssciolla: So, yes, we could derive this from the validation results. However, that would mean the "Please correct" alert would show up on a change as well as on a submit. The way it's written now, the user has to hit the submit button and then this alert pops up. I think that might be clearer from a UI perspective, since they get a clear response when they first try to hit submit, rather than nothing happening. It's a subtle difference, I guess I would be fine with it either way.

src/assets/src/components/queueSettings.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+    if (queue && !queueInitialized) {
+        setQueueInitialized(true);
+        setName(queue.name);
+        setDescription(queue.description);
+        setAllowedMeetingTypes(new Set(queue.allowed_backends));
+    }
+
+    const resetValidationResults = () => {
+        if (setShowCorrectGeneralMessage) setShowCorrectGeneralMessage(false);
```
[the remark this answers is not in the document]
ssciolla: So, I could remove the `if` for simplicity, but it made sense to me to only update the state if it needed to be. It's possible the user doesn't enter anything invalid and then hits submit, and this would still be `false`. I'm happy to change it if you'd prefer it the other way.

src/assets/src/components/queueSettings.tsx
```diff
@@ -0,0 +1,315 @@
...
+    const [descriptValidationResult, setDescriptValidationResult] = useState(undefined as undefined | ValidationResult);
+    const [allowedMeetingTypes, setAllowedMeetingTypes] = useState(new Set() as Set<string>);
+    const [allowedValidationResult, setAllowedValidationResult] = useState(undefined as undefined | MeetingTypesValidationResult);
+
+    if (queue && !queueInitialized) {
+        setQueueInitialized(true);
+        setName(queue.name);
+        setDescription(queue.description);
+        setAllowedMeetingTypes(new Set(queue.allowed_backends));
+    }
```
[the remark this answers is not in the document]
ssciolla: Yep, that makes sense. I made the change, incorporating it into the preexisting `setQueueChecked` wrapper, which you can see here: https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/173/commits/fc7277b8b9a04d9a24ec904003c4a666bcdf0158

src/assets/src/components/queueSettings.tsx
```diff
@@ -0,0 +1,315 @@
...
+                onSaveGeneralClick={handleSaveGeneralClick}
+                onDiscardGeneralClick={handleDiscardGeneralClick}
+                currentUser={props.user}
+                hosts={queue?.hosts}
+                onAddHost={doAddHost}
+                onRemoveHost={confirmRemoveHost}
+                checkHostError={addHostError ? { source: 'Add Host', error: addHostError } : undefined}
+                onDeleteClick={confirmRemoveQueue}
+            />
+        ) : undefined;
```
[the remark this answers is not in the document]
ssciolla: Done (https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/173/commits/12fd8d01441e25257c93effc20959a1630ba4a72). This isn't super well documented in MDN. I guess if they're both truthy, it returns the latter?

src/assets/src/components/queueSettings.tsx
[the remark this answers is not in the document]
jlost: I think that React only "does" anything (re-renders) if the value you set is different than the previous value, so I think calling the setter is a no-op if it's already `false`. But I could be wrong on this, and this is absolutely a quibble so feel free to resolve & ignore!

src/assets/src/components/addQueue.tsx
jlost: I'm good with that, just always looking for ways to pare down state.

src/assets/src/components/queueSettings.tsx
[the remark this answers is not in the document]
ssciolla: Huh, if that's true, I'm happy to simplify.

src/assets/src/validation.ts
```diff
@@ -76,3 +77,59 @@ export function validateString (value: string, schema: StringSchema, showRemaini
...
+    const noTypesSelected = value.size === 0;
+    if (noTypesSelected) messages.push('You must select at least one allowed meeting type.');
+
+    let existingMeetingConflict = false;
+    if (queue) {
+        const uniqueMeetingTypes = new Set(queue!.meeting_set.map(m => m.backend_type));
+        let conflictingTypes = Array();
+        uniqueMeetingTypes.forEach(uniqueMeetingType => {
+            if (!value.has(uniqueMeetingType)) conflictingTypes.push(uniqueMeetingType);
+        });
```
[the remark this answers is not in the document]
ssciolla: Done. See https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/173/commits/e15dc0e193e26db93aa44912d4f4c2039ffe7303 (note that I subsequently removed the log message).

src/assets/src/components/queueSettings.tsx
ssciolla: I think this is a similar but different case, but yeah, is this this what you were thinking? https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/173/commits/ef09509e5687a3c0548cf45513c20129c81745cd

I switched the `activeKey` variable to use an `enum` member, since I was getting some weird errors when trying to use a `type` declaration. Thus, I also need to use `typeof keyof` (which I came across in my research).

src/assets/src/components/queueSettings.tsx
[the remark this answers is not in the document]
ssciolla: Done. https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/173/commits/7c3baf7b84495f780bb8730d544bf9b1d917e3ff

src/assets/src/services/api.ts
```diff
@@ -96,6 +96,23 @@ export const createQueue = async (
     return await resp.json() as QueueHost;
 }
 
+export const updateQueue = async (
+    queue_id: number, name?: string, description?: string, allowed_backends?: Set<string>
+) => {
+    const queuePatched = Object();
```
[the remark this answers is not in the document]
ssciolla: I had to change this back to `Object`, since TypeScript seemed to be complaining about the object literal.

jlost (approved):
At this point, it looks great and mergeable! Feel free to keep refining it if you like and I can keep providing feedback.

src/assets/src/validation.ts
ssciolla: I'd like to think about this a little more. It's an interesting idea, but I'm not sure what more we'd be getting with the hook version...

src/assets/src/components/queueSettings.tsx
```diff
@@ -0,0 +1,315 @@
...
+            : descriptValidationResult;
+        const curAllowedValidationResult = !allowedValidationResult
+            ? validateAndSetMeetingTypesResult(allowedMeetingTypes, setAllowedValidationResult, queue)
+            : allowedValidationResult;
+
+        if (!curNameValidationResult!.isInvalid && !curDescriptValidationResult!.isInvalid && !curAllowedValidationResult!.isInvalid) {
+            const nameForUpdate = name.trim() !== queue?.name ? name : undefined;
+            const descriptForUpdate = description.trim() !== queue?.description ? description : undefined;
+            const allowedForUpdate = !compareStringArrays(Array.from(allowedMeetingTypes), queue!.allowed_backends)
+                ? allowedMeetingTypes : undefined;
```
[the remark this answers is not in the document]
ssciolla: I re-wrote `compareStringArrays` to use `Set`s, and I do think it's better. It didn't really feel like I could (or necessarily should) get rid of the utility function all together though. I based this on the MDN `symmetricDifference` function, as indicated in the comment. It would be nice if JavaScript had the nice operators for unions and intersections like Python does.

src/assets/src/components/queueSettings.tsx
ssciolla: I do think this should be cleaned up, but it'd be nice to get this merged as well. The custom hook idea was made a little more complicated as well when I moved the `name`, `description`, and `allowedMeetingTypes` initialization into this block in `queueSettings`. I'm going to create a new issue.

ssciolla: See issue #181.

src/assets/src/components/queueSettings.tsx
ssciolla: I started looking at this more carefully, and it's not actually working like I think it is. `AvailableTabs[eventKey as keyof typeof AvailableTabs]` seems to return `undefined`. Hm...

ssciolla: Seems like reverse mapping of string values back to enum member keys doesn't work. People are recommending a https://stackoverflow.com/questions/54039138/how-to-handle-reverse-mapping-of-string-enums-with-different-string-literals-in

The thing is, the tabs still mostly work, except on the first click. I tried removing all the tab management code entirely, and it actually works fine without it. So I'm going to go in that direction.

ssciolla: Yeah, this is just simpler all around, to let the default `Tab.Container` react-boostrap component handle the Tab changes in `queueSettings`.

I think it's still appropriate to override `onSelect` and `activeKey` in the `Tab.Container` in `addQueue.tsx`, that's a workflow and we need to prevent folks from proceeding if the first page is invalid.

ssciolla:
Thanks, @jlost, for the approval and feedback. I'll plan on merging in the morning.

ssciolla:
@jlost, I did one last sweep and found a couple small bugs in `addQueue.tsx` (https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/173/commits/7b974677673700259cfb2652e56307df6f0ed390), as well as some missing semicolons and places I could simplify throughout. Going to merge now.

PR #180, opened 2020-10-21
Fold profile API into user API

### LINKED ISSUE #179: Preferences page updates profile using user ID
Currently the profile page issues an update to the profile with the user ID like so: `PATCH /api/profiles/{user_id}/`. The `profile_id` should be used instead. This will often fail to rear its head in development environments because the `profile_id` and `user_id` are likely to sync up.

jlost:
Fixes #179 
Having the users and profile APIs separate added some complications to how they were read and updated, so I decided to roll them both into one. Narrowing the API surface seemed like a generally good idea. Fields that a user shouldn't be able to modify, like email, have been made read-only.

### DIFF
--- a/src/officehours_api/views.py
+++ b/src/officehours_api/views.py
@@ -17,7 +17,7 @@ from officehours_api.serializers import (
     QueueHostSerializer, MeetingSerializer, AttendeeSerializer,
 )
 from officehours_api.permissions import (
-    IsHostOrReadOnly, IsHostOrAttendee, is_host, IsCurrentProfile
+    IsHostOrReadOnly, IsHostOrAttendee, is_host,
 )
 
 
--- a/src/officehours_api/views.py
+++ b/src/officehours_api/views.py
@@ -57,7 +57,7 @@ class UserDetail(DecoupledContextMixin, LoggingMixin, generics.RetrieveUpdateAPI
     queryset = User.objects.all()
     permission_classes = (IsAuthenticated,)
 
-    def get_serializer(self, instance, data=None, many=None, partial=None):
+    def get_serializer(self, instance=None, data=None, many=None, partial=None):
         ctx = self.get_serializer_context()
         kwargs = {}
         if data:
@@ -80,7 +80,7 @@ class UserDetail(DecoupledContextMixin, LoggingMixin, generics.RetrieveUpdateAPI
     def partial_update(self, request, *args, **kwargs):
         user = self.get_object()
         self.check_change_permission(request, user)
-        return super().update(request, *args, **kwargs)
+        return super().partial_update(request, *args, **kwargs)
 
 
 class UserUniqnameDetail(UserDetail):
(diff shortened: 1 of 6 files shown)


src/officehours_api/views.py
```diff
@@ -53,17 +53,34 @@ class UserList(DecoupledContextMixin, generics.ListAPIView):
...
-        else:
-            serializer = ShallowUserSerializer(user, context={'user': request.user})
-        return Response(serializer.data)
+        self.check_change_permission(request, user)
+        return super().update(request, *args, **kwargs)
+    
+    def partial_update(self, request, *args, **kwargs):
+        user = self.get_object()
+        self.check_change_permission(request, user)
+        return super().update(request, *args, **kwargs)
```
ssciolla, line 83: These are basically identical. I'm assuming we need both?

src/officehours_api/views.py
```diff
@@ -53,17 +53,34 @@ class UserList(DecoupledContextMixin, generics.ListAPIView):
...
+    def get_serializer(self, instance, data=None, many=None, partial=None):
+        ctx = self.get_serializer_context()
+        kwargs = {}
+        if data:
+            kwargs['data'] = data
+        return (
+            MyUserSerializer(instance, context=ctx, **kwargs)
+            if instance == ctx['user']
+            else ShallowUserSerializer(instance, context=ctx, **kwargs)
+        )
```
ssciolla, line 69: I saw this traceback when I was testing this with the Swagger UI. Looks like they offer a solution, but don't know exactly what's going on myself.
```
view's UserDetail raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
File "/usr/local/lib/python3.8/site-packages/drf_yasg/inspectors/base.py", line 50, in call_view_method
     return view_method()
TypeError: get_serializer() missing 1 required positional argument: 'instance'
```

src/officehours_api/views.py
```diff
@@ -53,17 +53,34 @@ class UserList(DecoupledContextMixin, generics.ListAPIView):
...
+
+    def check_change_permission(self, request, user):
+        if user != request.user:
+            self.permission_denied(request)
```
ssciolla, line 73: So you removed the `IsCurrentProfile` permission class, and this is kinda doing the work now of ensuring the current user is the user that's being changed. Would it make sense for this to be a `IsCurrentUser` permission class instead?

ssciolla (commented):
Looks pretty good to me, though there's some fancy DRF stuff I don't know that well. I had a couple questions and did come across a Traceback when using Swagger that would be nice to wipe out.

src/officehours_api/views.py
jlost: If I applied it as a permission class, it would apply to the `Retrieve` method as well, but I want that to remain callable (and return a `ShallowUser`).

src/officehours_api/views.py
jlost: Thanks for pointing that out, I noticed that in the latter one, it should be calling `super().partial_update`.

src/officehours_api/views.py
jlost: Good catch, I did fix this by setting `instance=None` in the args (made the arg optional).

src/officehours_api/views.py
ssciolla: Ah, that makes sense to me generally, even though I don't have a great handle on the permission classes.

src/officehours_api/views.py
ssciolla: Nice

ssciolla (approved):

PR #187, opened 2020-10-23
Lock pip dependencies in Dockerfile (#186)

### LINKED ISSUE #186: Lock pip requirements for hosting
As @ssciolla mentioned in #176, updates to packages in `Dockerfile` can break the build, so they should be locked to a compatible version either there or in a separate `host_requirements.txt`.

jlost:
Fixes #186


ssciolla (approved):
LGTM! Builds and runs locally. Versions are the latest available.

Issue #188, opened 2020-10-23
Refactor validateAndSetXResult wrapper functions as custom hooks


ssciolla:
In PR #173, I created a couple wrapper functions that would validate values and set `ValidationResult`s for preexisting `useState` hooks. However, the use of these functions requires passing variables repeatedly in at each use. In an OOP context, a class would make sense to store the shared variables and actions. However, @jlost appropriately suggested twice using a [React custom hook](https://reactjs.org/docs/hooks-custom.html). This makes sense to me after some review and experimentation, and I'd like to handle soon before the focus moves to other things on this project.

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/f7cf0ebcf1a9e1f641fca9009d001bcbd5d25865/src/assets/src/validation.ts#L117-L139

ssciolla:
For QA, we just need to make sure no regressions were introduced, i.e. that the forms and form validation still work as expected.

PR #189, opened 2020-10-23
Create custom validation hooks (#188)

ssciolla:
This PR refactors the existing `validateAndSet...` wrapper functions as the custom React hooks `useStringValidation` and `useMeetingTypesValidation` and updates usages in several components. The PR attempts to resolve issue #188.

### DIFF
--- a/src/assets/src/hooks/useValidation.ts
+++ b/src/assets/src/hooks/useValidation.ts
@@ -1,0 +1,34 @@
+import { useState } from "react";
+import { StringSchema } from "yup";
+
+import { QueueHost } from "../models";
+import { 
+    MeetingTypesValidationResult, validateMeetingTypes, validateString, ValidationResult
+} from "../validation";
+
+
+export function useStringValidation (schema: StringSchema, showRemaining?: boolean):
+    [undefined | ValidationResult, (value: string) => ValidationResult, () => void]
+{
+    const [validationResult, setValidationResult] = useState(undefined as undefined | ValidationResult);
+    const validateAndSetResult = (newValue: string) => {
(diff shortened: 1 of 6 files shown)


src/assets/src/hooks/useValidation.ts
```diff
@@ -0,0 +1,34 @@
...
+{
+    const [validationResult, setValidationResult] = useState(undefined as undefined | MeetingTypesValidationResult);
+    const validateAndSetResult = (newValue: Set<string>) => {
+        const result = validateMeetingTypes(newValue, queue);
+        setValidationResult(result);
+        return result;
+    }
+    const clearResult = () => setValidationResult(undefined);
+    return [validationResult, validateAndSetResult, clearResult];
+}
```
ssciolla, line 34: Is this what you were thinking, @jlost? Thanks for prodding me to do this, I like the result. I was previously thinking classes would also solve the parameter repetition, but this makes sense to me for this context.

jlost (approved):
Yeah, this looks good. It might be possible to take it even one step farther with generics (`useValidation<T>`) but there isn't that much repetition yet.

ssciolla:
> Yeah, this looks good. It might be possible to take it even one step farther with generics (`useValidation<T>`) but there isn't that much repetition yet.

Hm, yeah, this occurred to me, but I wasn't sure how to do it exactly. I started playing around with it, but was having a lot of trouble (probably partly cause I don't have a mastery of generics). The interior of the hooks are pretty similar, but the parameters are quite different. I think I'd need to maybe refactor the validation functions, too? I guess this is something we should consider in the future if the validation code gets built out in some way, as you suggested. I'm gonna focus on the other current project issues for now.

Issue #184, opened 2020-10-22
Prompt attendee to enable SMS notifications


mfldavidson:
We want to make sure that attendees know about the option to enable SMS notifications at the appropriate time.

Just below "Your number in line", IF the number is GREATER THAN 1, show this blurb:

> Get notified by SMS (text) message when it's your turn by adding your cell phone number and enabling attendee notifications in your [User Preferences](/preferences/).

Questions: is this the best UX? I do not want us to have to go out and check if someone has this set, that's too technically complex for right now, so does this wording confuse people and make them think that we're saying they don't have it set even if they do? Would this be better?

> If you have opted in to attendee notifications in your [User Preferences](/preferences/), you will get an SMS (text) message when it is your turn.

@teshsopher and Joel (will send to Joel separately as he isn't on GitHub) I'd appreciate your input on the wording here.

jlost:
How's this visually: 
![image](https://user-images.githubusercontent.com/11338926/97048118-e8f28000-1547-11eb-820d-8d8d4e7ff6e5.png)

teshsopher:
For wording, we could say, "Make sure to opt into attendee notifications in User Preferences to receive and SMS (text) when it's your turn."

For the layout, the link doesn't have enough contrast, but this is one of the better ways to make sure attendees see it. Another option could be to put it above the info box like:

<img width="519" alt="image" src="https://user-images.githubusercontent.com/63423382/97181262-74028e80-1771-11eb-9b03-e123333763af.png">

The problem with the second option is that attendees might skip over the text.

mfldavidson:
I agree with the concern that attendees will skip over it. I'd like to stick with the layout James has recommended if we can. Is it bad practice to change the color of just this one link hard-coded?

Similarly to the other one, I'd like to apply Joel's recommended verbage here:

> Did you know? You can receive a SMS (text) message when it's your turn by by adding your cell phone number and enabling attendee notifications in your [User Preferences](/preferences/).

jlost:
I believe Sam's fix here will solve the problem of link color contrast in light blue alerts:
https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/191/files#r512158694

jlost:
Here's how that looks: 
![image](https://user-images.githubusercontent.com/11338926/97454508-376b9a00-190d-11eb-9d94-b32962a97b5e.png)

jlost:
[...]

PR #190, opened 2020-10-23
Add prompts to enable SMS notifications #182, #184 

### LINKED ISSUE #182: Prompt host to enable SMS notifications
We want to make sure that hosts know about the option to enable SMS notifications at the appropriate time.

Where it currently says "No meetings in queue" in a gray box, change it to the following:

> No meetings in queue. Get notified by SMS (text) message when someone joins your empty queue by adding your cell phone number and enabling host notifications in your [User Preferences](/preferences/).

Questions: is this the best UX? I do not want us to have to go out and check if someone has this set, that's too technically complex for right now, so does this wording confuse people and make them think that we're saying they don't have it set even if they do? Would this be better?

> No meetings in queue. If you have opted in to host notifications in your [User Preferences](/preferences/), you will get an SMS (text) message when someone joins your empty queue.

@teshsopher and Joel (will send to Joel separately as he isn't on GitHub) I'd appreciate your input on the wording here.

jlost:
Adds text to remind users of the existence of notifications on the queue page and the edit page.

Will rely on @ssciolla's accessibility tweaks to fix the color contrast problem in the `info` alert.

Closes #182, #184

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -180,7 +180,7 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
         && (
             <Card.Text>
                 <Alert variant="info">
-                    Did you know? You can receive a SMS (text) message when 
+                    Did you know? You can receive an SMS (text) message when 
                     it's your turn by by adding your cell phone number and 
                     enabling attendee notifications in 
                     your <Link to="/preferences">User Preferences</Link>.
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -175,6 +175,17 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
         ? <small className="card-text-spacing meeting-type-message">A Host has been assigned to this meeting. Meeting Type can no longer be changed.</small>
         : <button disabled={props.disabled} onClick={props.onShowDialog} type="button" className="btn btn-link">Change</button>;
     const agendaText = props.queue.my_meeting!.agenda
+    const numberInLine = props.queue.my_meeting!.line_place + 1;
+    const notificationBlurb = numberInLine > 1
+        && (
+            <Card.Text>
+                <Alert variant="info">
+                    If you have opted in to attendee notifications in
+                    your <Link to="/preferences">User Preferences</Link>,
+                    you will get an SMS (text) message when it is your turn.
+                </Alert>
+            </Card.Text>
+        );
     return (
         <>
             {closedAlert}
@@ -182,7 +193,8 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
             <h3>You are currently in line.</h3>
             <Card className='card-middle card-width center-align'>
                 <Card.Body>
-                    <Card.Text>Your number in line: <strong>{props.queue.my_meeting!.line_place + 1}</strong></Card.Text>
+                    <Card.Text>Your number in line: <strong>{numberInLine}</strong></Card.Text>
+                    {notificationBlurb}
                     <Card.Text>Time Joined: <strong><DateTimeDisplay dateTime={props.queue.my_meeting!.created_at}/></strong></Card.Text>
                     <Card.Text>
                         Meeting via: <strong>{props.backends[props.queue.my_meeting!.backend_type]}</strong>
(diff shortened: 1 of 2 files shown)


src/assets/src/components/queueManager.tsx
```diff
@@ -212,7 +212,13 @@ function QueueManager(props: QueueManagerProps) {
             </Table>
         )
         : (
-            <Alert variant="dark">No meetings in queue.</Alert>
+            <>
+            <hr/>
+            <h5>No meetings in Queue</h5>
+            <p>
+                Did you know? You can get notified by SMS (text) message when someone joins your empty queue by adding your cell phone number and enabling host notifications in your <Link to="/preferences">User Preferences</Link>. 
+            </p>
```
ssciolla, line 220: I know @teshsopher suggested removing the Alert formatting, but semantically this feels to me like an alert; that would also be consistent with how we are doing it for the attendee. I also find the use of `<hr/>` here to be a little different from what we've been doing elsewhere, where it's kind of a footer separator.

src/assets/src/components/queueManager.tsx
```diff
@@ -212,7 +212,13 @@ function QueueManager(props: QueueManagerProps) {
             </Table>
         )
         : (
-            <Alert variant="dark">No meetings in queue.</Alert>
+            <>
+            <hr/>
+            <h5>No meetings in Queue</h5>
```
ssciolla, line 217:
```suggestion
            <h5>No Meetings in Queue</h5>
```

src/assets/src/components/queueManager.tsx
```diff
@@ -212,7 +212,13 @@ function QueueManager(props: QueueManagerProps) {
             </Table>
         )
         : (
-            <Alert variant="dark">No meetings in queue.</Alert>
+            <>
+            <hr/>
```
ssciolla, line 216: Nit-pick: you could indent a bit more here for readability.

src/assets/src/components/queue.tsx
```diff
@@ -175,14 +175,27 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
         ? <small className="card-text-spacing meeting-type-message">A Host has been assigned to this meeting. Meeting Type can no longer be changed.</small>
         : <button disabled={props.disabled} onClick={props.onShowDialog} type="button" className="btn btn-link">Change</button>;
     const agendaText = props.queue.my_meeting!.agenda
+    const numberInLine = props.queue.my_meeting!.line_place + 1;
+    const notificationBlurb = numberInLine > 1
+        && (
+            <Card.Text>
+                <Alert variant="info">
+                    Did you know? You can receive a SMS (text) message when
```
ssciolla, line 183: I think this should be "You can receive **an** SMS (text) message", since SMS phonetically starts with a vowel sound of "es". See https://www.merriam-webster.com/words-at-play/is-it-a-or-an

ssciolla (approved):
Technically, this all looks good. I made a couple suggestions/comments. I do think using `Alert` in both places makes more sense to me, especially if the link contrast issue is addressed elsewhere, but it's not my call ultimately : ) The grammar issue probably needs to be addressed, but otherwise, please feel free to merge when you think it's ready.

src/assets/src/components/queue.tsx
ssciolla: Also, we may want to consider wrapping the text here with `small`, since the Card text generally is kind of large. That would make this text the same size as the instructional text for the agenda field.

src/assets/src/components/queueManager.tsx
teshsopher: If we want to be more consistent, using the Alert formatting on the host side is fine for this release. However, I think one of my next goals is to reduce the number of Alert-style boxes, particularly on the attendee side, to improve overall clarity of ROHQ.

ssciolla: I will defer to @teshsopher. I guess, just because this is a conditionally rendered block, the Alert made more sense to me. I can understand not wanting to overuse that info presentation method.

src/assets/src/components/queueManager.tsx
ssciolla: Huh, thanks for explaining that. Your approach seems fine. I commented because in general I find it easier to read markup when the levels are clear. I've often used `<div>` instead of `<>` because I thought (I guess erroneously) that they were just converted into `div`s. I guess there's no harm in overusing `div`s, but I'm glad you pointed out these nuances.

jlost:
Moving forward and merging. Let's open a new issue if we want to make any more tweaks to wrt the host blurb.

Issue #155, opened 2020-09-10
Fix Color Contrast Issues


teshsopher:
There are some links and buttons that don't have enough contrast. Minimum contrast should be 4.5:1 (ideally 7:1 for WCAG AAA compliance).

Places this should be adjusted:
- Alert banners: Links are not a high enough contrast
- Breadcrumbs: Links in the breadcrumb don’t have high enough contrast
- Manage: “+ Add Queue button” does not have a sufficient contrast with the button text (Maybe a darker green?)
- Manage Queue: “+ Add Host” and “+ Add Attendee” buttons don’t have enough contrast (see previous point)
- Manage Queue: Disabled “Remove” button (trash icon) – Doesn’t have enough contrast; recommendation – Leave it enabled, but have an error message that tells why the user can’t click it instead
- Attendee View: “Manage it” link doesn’t have enough contrast

ssciolla:
I'm going to take notes as I try to address these:

- [] "Alert banners: Links are not a high enough contrast" - 
- [] "Breadcrumbs: Links in the breadcrumb don’t have high enough contrast"
- [] "Manage: “+ Add Queue button” does not have a sufficient contrast with the button text (Maybe a darker green?)" -
- [] "Manage Queue: “+ Add Host” and “+ Add Attendee” buttons don’t have enough contrast (see previous point)" -
- [] "Manage Queue: Disabled “Remove” button (trash icon) – Doesn’t have enough contrast; recommendation – Leave it enabled, but have an error message that tells why the user can’t click it instead"
- [] "Attendee View: “Manage it” link doesn’t have enough contrast" -

teshsopher:
Tested and working in QA

PR #191, opened 2020-10-26
Fix color contrast issues, make other misc. edits (#155)

ssciolla:
This PR tries to address the color contrast issues identified during the accessibility review. A couple other supporting or miscellaneous edits to the user interface and associated language are included. Comments will added requesting specific feedback on proposed hex values or changes. Other changes are described in a task list below. The PR aims to resolve issue #155.

- [x] Replace CSS class called `page-content-flow` which I previously introduced with the use of Bootstrap margin utilities (e.g. `mt-3`).
- [x] Remove link, but not button, from Tab order for Add Queue button, making only one action related to the function keyboard accessible.
- [x] Change label text for the `queue.created_at` value to "Created", as it is in the current release.

### DIFF
--- a/src/assets/src/components/queueEditors.tsx
+++ b/src/assets/src/components/queueEditors.tsx
@@ -93,14 +93,19 @@ export function ManageHostsEditor(props: ManageHostsEditorProps) {
     const hostsSoFar = props.hosts.map((host, key) => (
         <ListGroup.Item key={key}>
             <UserDisplay user={host} />
-            <div className='float-right'>
-                <RemoveButton
-                    onRemove={() => props.onRemoveHost(host)}
-                    size='sm'
-                    disabled={props.disabled || host.id === props.currentUser?.id}
-                    screenReaderLabel='Remove Host'
-                />
-            </div>
+            {
+                (host.id !== props.currentUser?.id)
+                    && (
+                        <div className='float-right'>
+                            <RemoveButton
+                                onRemove={() => props.onRemoveHost(host)}
+                                size='sm'
+                                disabled={props.disabled}
+                                screenReaderLabel='Remove Host'
+                            />
+                        </div>
+                    )
+            }
         </ListGroup.Item>
     ));
 
@@ -128,7 +133,10 @@ export function ManageHostsEditor(props: ManageHostsEditorProps) {
                 + Add Host
             </SingleInputField>
             <h3>Current Hosts</h3>
-            <p>To remove a host, select the trash icon to the right of the user's name. You cannot remove yourself as a host.</p>
+            <p>
+                To remove a host, select the trash icon to the right of the user's name.
+                <strong>You cannot remove yourself as a host.</strong>
+            </p>
             <ListGroup>{hostsSoFar}</ListGroup>
         </div>
     );
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -174,11 +174,11 @@ h3, .h3 {
 }
 
 a:not(.btn), a:link:not(.btn), a:visited:not(.btn) {
-	color: #004a8b;
+	color: #0060ba;
 }
 
 a:hover {
-	color: #004a8b;
+	color: #0060ba;
 }
 
 .bottom-right {
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -181,6 +181,14 @@ a:hover {
 	color: #0060ba;
 }
 
+.alert a {
+	text-decoration: underline;
+}
+
+p a {
+	text-decoration: underline;
+}
+
 .bottom-right {
 	position: fixed;
 	bottom: 0;
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -322,20 +322,20 @@ a.nav-link.disabled:hover, a.nav-link.disabled:active, a.nav-link.disabled:focus
 }
 
 .btn-success {
-    color: #ffffff;
-    background-color: #00820d;
-    border-color: #00820d;
+	color: #ffffff;
+	background-color: #00820d;
+	border-color: #00820d;
 }
 
 .btn-outline-success {
-    border-color: #00820d;
-    color: #00820d;
+	border-color: #00820d;
+	color: #00820d;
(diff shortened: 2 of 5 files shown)


src/assets/src/components/queueEditors.tsx
```diff
@@ -93,14 +93,19 @@ export function ManageHostsEditor(props: ManageHostsEditorProps) {
...
+                        <div className='float-right'>
+                            <RemoveButton
+                                onRemove={() => props.onRemoveHost(host)}
+                                size='sm'
+                                disabled={props.disabled}
+                                screenReaderLabel='Remove Host'
+                            />
+                        </div>
+                    )
+            }
```
ssciolla, line 108: @teshsopher, to address the contrast issues with the disabled "Remove Host" trash icon button, I decided to try simply removing that button when the host is the current user. To support this, I also made the phrase "You cannot remove yourself as a host" bold. What do you think about this solution? I wanted to avoid repeated language from the instructions, but I'm curious whether you think this works as part of your holistic vision for the UI.

src/officehours_ui/static/css/main.css
```diff
@@ -321,14 +321,20 @@ a.nav-link.disabled:hover, a.nav-link.disabled:active, a.nav-link.disabled:focus
 	color: #ffffff;
 }
 
+.btn-success {
+	color: #ffffff;
+	background-color: #00820d;
+	border-color: #00820d;
```
ssciolla, line 327: I modified the default Bootstrap button success color to be a darker shade of green (`#00820d`), resulting in a constrast ratio of 5 to 1 with white. I didn't want to too much darker, since I felt those shades look different enough to maybe be confusing or jarring to users' familiar with (or subconsciously familiar with) the Bootstrap palette. Thoughts, @teshsopher? I am open to feedback on this.

src/officehours_ui/static/css/main.css
```diff
@@ -174,11 +174,11 @@ h3, .h3 {
...
 
 a:not(.btn), a:link:not(.btn), a:visited:not(.btn) {
-	color: #0075bc;
+	color: #004a8b;
 }
 
 a:hover {
-	color: #0065a3;
+	color: #004a8b;
 }
```
ssciolla, line 182: @teshsopher, I modified the existing link color for non-button links to be a darker shade of blue, resulting in a 8.92 contrast ratio with white and a 7.21 contrast ratio with the light blue Bootstrap color seen in the background for `info` alerts (e.g. Message of the Day, "This is your queue, you can manage it"). Does this resolve those issues you raised? Do you want to tweak the hex value at all? I'm a little wary of making this so dark globally, since it may be getting too close to black and may make links not easily distinguishable (without the user hovering). It would be nice to keep the link color consistent across the board. Would AA level be enough in this situation?

jlost (approved):
LGTM

src/officehours_ui/static/css/main.css
jlost: Here's how it looks: 
image

A different approach would be to lighten the shade of blue in `info` alerts.

src/assets/src/components/queueEditors.tsx
teshsopher: I think this works, based on the code, but I'd like to see it visually too

src/officehours_ui/static/css/main.css
teshsopher: Meeting AAA requirements is just an ideal! In this case, AA is good enough, and having the link text be too close to black can definitely be a problem, like you mentioned. Unfortunately, it looks like, for the blue background, the darker color works best, or we would have to change the blue background to something lighter like F0F9F9, but then we risk it not being visible against the white. 

In the case of the darker text, it may be helpful to underline links that are in the alert/message boxes so that they stand out more and look more like clickable links.

src/officehours_ui/static/css/main.css
teshsopher: That's perfect!

ssciolla: Cool, thanks

src/assets/src/components/queueEditors.tsx
ssciolla: @teshsopher, it looks like this:
Screen Shot 2020-10-29 at 2 25 34 PM

src/officehours_ui/static/css/main.css
ssciolla: K, let me see what I can do. I'm gonna bump the link color down a bit so it's AA but still looks link-y. I'll also look into the underlining on alerts, though in an ideal world I'd prefer if links looked/behaved the same everywhere in the application.

ssciolla: Okay, I switched to `#0060ba` for links, which puts us at 6.2 contrast with white, and 5.02 with the `info` background color.

ssciolla: @teshsopher, do we think the color change will be sufficient? We can certainly add the underline with alerts using CSS, but I do think there is some benefit to link behavior being consistent throughout. I'm looking at how it would be if the  underline was everywhere, but that seems a little distracting in places like the breadcrumbs and queue list table.

teshsopher: The link color is very similar to the regular text, so it may be confusing with just the color change. For the sake of clarity, adding some other indicator would be good in this case. However, I do understand your concern with having the links be consistent throughout, and it may be worth underlining text when it's in a body of text, such as the Message of the Day or in the "Joining the meeting" box, and not underlining the breadcrumbs. For example, [the BlueJeans page](https://its.umich.edu/communication/videoconferencing/blue-jeans) on the ITS site uses a similar pattern to help with this type of issue. I agree that it can be a little overwhelming visually if we have too much text grouped together or several links in a small space, so it's definitely something that can use some refining.

ssciolla: Okay, let me give this a shot.

ssciolla: Okay, @teshsopher, I think I did what you requested here: https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/191/commits/fccc764128fd864c2bb4bb2dd71c7443dbca61d9

I think this is a good solution for now. I'm going to merge this so you can start taking a look in QA.

Issue #137, opened 2020-08-17
Warning Popup when Attendee clicks on "Leave the line"


teshsopher:
In order to prevent attendees from accidentally leaving the line, there should be a warning message that pops up letting users know what will happen so they can confirm if that's what they want to do. 

Title of popup: Leave the line
Popup message: You are about to leave the line. You will lose your place in queue.

mfldavidson:
Can this be pushed to QA @ssciolla @jlost I just tested and it isn't there

ssciolla:
Yeah, I'll do that. Just need to re-build.

teshsopher:
Tested and working in QA

PR #192, opened 2020-10-27
Modify showConfirmation, ensure confirmation pop-up when user leaves line (#137)

ssciolla:
This PR makes changes to the `queue.tsx` component so users are shown a pop-up confirmation (with language different whether the queue status is "open" or "closed") whenever they click the "Leave the line" button. I also refactored a little to minimize duplication and tweaked `showConfirmation` (perhaps arbitrarily) so that it doesn't prescribe a prefix or introductory language for the pop-up's body message. See comments on the code for explanations. The PR aims to resolve issue #137.

### DIFF
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -86,10 +86,10 @@ export const LoadingDisplay: React.FC<LoadingDisplayProps> = (props) => {
 }
 
 
-export const showConfirmation = (dialog: React.RefObject<Dialog>, action: () => void, title: string, actionDescription: string) => {
+export const showConfirmation = (dialog: React.RefObject<Dialog>, action: () => void, title: string, description: string) => {
     dialog.current!.show({
         title: title,
-        body: `Are you sure you want to ${actionDescription}?`,
+        body: description,
         actions: [
             Dialog.CancelAction(),
             Dialog.OKAction(action),
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -351,10 +351,11 @@ export function QueuePage(props: PageProps<QueuePageParams>) {
     }
     const [doLeaveQueue, leaveQueueLoading, leaveQueueError] = usePromise(leaveQueue);
     const confirmLeaveQueue = (queueStatus: 'open' | 'closed') => {
+        const loseMessage = 'By leaving the queue, you will lose your place in line'
         const description = 'Are you sure you want to leave the queue? ' + (
             queueStatus === 'open'
-                ? "You will lose your place in line."
-                : "The queue is currently closed. If you leave now, you will not be able to rejoin until the queue is reopened."
+                ? loseMessage + '.'
+                : `The queue is currently closed. ${loseMessage} and will not be able to rejoin until the queue is reopened.`
         );
         showConfirmation(dialogRef, () => doLeaveQueue(), 'Leave Queue?', description);
     }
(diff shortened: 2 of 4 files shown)


src/assets/src/components/queue.tsx
```diff
@@ -340,17 +350,13 @@ export function QueuePage(props: PageProps<QueuePageParams>) {
...
-                Dialog.OKAction(() => {
-                    doLeaveQueue();
-                }),
-            ],
-        });
+    const confirmLeaveQueue = (queueStatus: 'open' | 'closed') => {
+        const description = 'Are you sure you want to leave the queue? ' + (
+            queueStatus === 'open'
+                ? "You will lose your place in line."
+                : "The queue is currently closed. If you leave now, you will not be able to rejoin until the queue is reopened."
```
ssciolla, line 357: @mfldavidson or @teshsopher , I'm looking for feedback on the language here, since I tweaked it a bit. I can change it if desired.

When leaving queue with status open: "Are you sure you want to leave the queue? You will lose your place in line."
When leaving queue with status closed: "Are you sure you want to leave the queue? The queue is currently closed. If you leave now, you will not be able to rejoin until the queue is reopened."

src/assets/src/components/common.tsx
```diff
@@ -86,10 +86,10 @@ export const LoadingDisplay: React.FC<LoadingDisplayProps> = (props) => {
 }
 
 
-export const showConfirmation = (dialog: React.RefObject<Dialog>, action: () => void, title: string, actionDescription: string) => {
+export const showConfirmation = (dialog: React.RefObject<Dialog>, action: () => void, title: string, description: string) => {
     dialog.current!.show({
         title: title,
-        body: `Are you sure you want to ${actionDescription}?`,
+        body: description,
```
ssciolla, line 92: I elected to remove the "Are you sure you want to " prefix here and add it as part of the strings accompanying `showConfirmation` usages. I realize that introduces more repetition of language throughout, and I could see this being the wrong call. It just felt a little limiting for this function to require you to use that language every time, and also made it a little harder to understand what was happening when looking at any individual usage of "showConfirmation".

ssciolla: I guess I missed initially that it also included the trailing `?`. We definitely would need to modify that so that we could include multiple sentences.

src/assets/src/components/queue.tsx
teshsopher: For the queue with status open, it may be good to say, "By leaving the queue, you will lose your place in line."
For the queue with status closed: "The queue is currently closed. By leaving the queue, you will lose your place in line and will not be able to rejoin until the queue is reopened."

I'm okay with keeping the "Are you sure you want to leave the queue?" in either case. The message mostly should emphasize the action that's being taken and what will happen as a result of the action.

ssciolla: Okay, I will change the language to what you specified. I'm curious, do you think the button should be "Leave the queue" or "Leave the line"? Seems like the "Are you sure you want to **blank**?" should be aligned with the button action.

jlost (approved):

src/assets/src/components/queue.tsx
ssciolla: I updated the language as requested: https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/192/commits/30dc539f1da0c0353c39a7e87930ecfcf4eea8cb

ssciolla:
I'm gonna merge this for now, I'll raise the question about renaming the "Leave the line" button text separately.

Issue #156, opened 2020-09-10
Disambiguation of the "Edit" buttons


teshsopher:
The edit buttons on a queue's page need some disambiguation.

- When edit is selected, the opened fields should have a label (Edit Description does, but Edit Title doesn’t)
- When edit is clicked, move focus to text field so screen reader reads the name (NVDA doesn’t go into focus mode, so it’s not reading the field name; It works on the title edit, but not the description edit)

A11y Team Recommendation: Have the field exist, but hidden when the page is loaded. This may make it easier for screen readers to adjust for the field changes.

Note: This change may be unnecessary for when the Queue Settings page(s) are up.

jlost:
Looks like the edit buttons are no longer used after the Queue Settings page changes. Feel free to have a look in QA and reopen if I'm misunderstanding.

ssciolla:
@jlost, the Edit button is still used for the Agenda field in `queue.tsx`. That's the only remaining use I'm aware of.

teshsopher:
Since QA isn't really needed for this issue (I don't think anything was changed?), I'm moving this to done!

ssciolla:
@teshsopher , could you please test that the input field for the Agenda Edit button field receives focus at the right times (i.e. not on initial page load if if it starts in an editing state, but always focused when it appears after clicking the Edit button)?

ssciolla:
Also, probably worth checking that the Agenda field has the label you requested.

teshsopher:
Tested and working in QA. The field also has the label!

PR #194, opened 2020-10-28
Reintroduce self-focusing code to improve EditToggleField accessibility (#156)

ssciolla:
This PR adds code similar to a preexisting pattern in the project that allows `SingleInputField` components to focus their underlying input fields on initial render. This is then used by the `EditToggleField` to ensure the form input receives when the field is toggled to an `editing` state. Note: The `EditToggleField` is now only used by attendees when entering agenda values. The PR aims to resolve PR #156.

### DIFF
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -8,6 +8,7 @@ import Dialog from "react-bootstrap-dialog";
 import { StringSchema } from "yup";
 
 import { useStringValidation } from "../hooks/useValidation";
+import { useInitFocusRef } from "../hooks/useInitFocusRef";
 import { QueueAttendee, QueueBase, User } from "../models";
 import { ValidationResult } from "../validation";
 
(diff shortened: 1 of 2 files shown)


src/assets/src/components/common.tsx
```diff
@@ -247,21 +255,29 @@ export const StatelessInputGroupForm: React.FC<StatelessValidatedInputFormProps>
...
 }
 
 export const StatelessTextAreaForm: React.FC<StatelessValidatedInputFormProps> = (props) => {
+    const inputRef = createRef<FormControl<"textarea"> & HTMLTextAreaElement>();
+    useEffect(() => {
+        if (props.initFocus) {
+            inputRef.current!.focus();
+        }
+    }, []);
+
```
jlost, line 280: What do you think about making a hook that creates the ref, sets up the effect, and returns the ref?

jlost (approved):

src/assets/src/components/common.tsx
ssciolla: Let me see what it looks like.

ssciolla: @jlost, how about this? https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/194/commits/79d9792c37e66895fdc5fcf1329bebabc5a1e32e

I'm still getting comfortable with generics and these custom hooks. It seemed to work well to rely on passing a type to `<>` as part of the function calls, but I'm not entirely clear on whether that's a standard practice. It does seem to be how `createRef` works.

ssciolla: I made a couple other edits, so I'm going to re-request your review.

jlost (approved):
Looks good!

Issue #168, opened 2020-09-21
Websockets don't auto-reconnect on Chrome, or do but error doesn't clear


mfldavidson:
This is hard to reproduce, but on Chrome Desktop, if a page is idle for a long time causing the user connection to drop (and probably queue connection too but I did this on a page that doesn't have a queue connection), it is unclear whether it auto-reconnects or not. The error shows up and stays there, but it appears to still work...
![ezgif com-gif-maker](https://user-images.githubusercontent.com/31591389/93819548-de338b00-fc29-11ea-9245-94a1c6aeba82.gif)

On Chrome Mobile (also hard to reproduce) if a page is idle for a long time, both user and queue connection drop and the error just sits there. If I try to, say, add an agenda as an attendee on a queue where I am in line if it is showing these errors, it doesn't save, so it does seem that they aren't auto-reconnecting.

jlost:
I've noticed this before as well - by accident. I spent some time yesterday trying to repro this but wasn't able to. Would it be acceptable to "fix" (really, mitigate the confusion of) this by telling the user to refresh after *x* number of retries have elapsed?

PR #195, opened 2020-10-28
Prompt user to refresh after failed WebSocket reconnects (#168)

ssciolla:
This PR uses `ws.retryCount` to change the connection error messages to indicate "Unable to reconnect; please refresh the page". The PR aims to resolve issue #168.

Is this what you had in mind, @jlost? Feel free to push to or open a PR against the branch on my fork.

### DIFF
--- a/src/assets/src/hooks/useWebSocket.ts
+++ b/src/assets/src/hooks/useWebSocket.ts
@@ -8,7 +8,7 @@ interface OfficeHoursMessage<T> {
 
 export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onDelete?: (setError: (React.Dispatch<React.SetStateAction<Error | undefined>>)) => void) => {
     const [error, setError] = useState(undefined as Error | undefined);
-    const numRetries = 3;
+    const numRetries = 10;
     useEffect(() => {
         const ws = new ReconnectingWebSocket(
             url,
--- a/src/assets/src/hooks/useWebSocket.ts
+++ b/src/assets/src/hooks/useWebSocket.ts
@@ -8,13 +8,12 @@ interface OfficeHoursMessage<T> {
 
 export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onDelete?: (setError: (React.Dispatch<React.SetStateAction<Error | undefined>>)) => void) => {
     const [error, setError] = useState(undefined as Error | undefined);
+    const numRetries = 3;
     useEffect(() => {
         const ws = new ReconnectingWebSocket(
             url,
             undefined,
-            {
-                maxRetries: 3,
-            },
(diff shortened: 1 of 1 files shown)


src/assets/src/hooks/useWebSocket.ts
```diff
@@ -57,7 +56,9 @@ export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onD
             console.error(e);
             const errorName = e?.error?.name as string | undefined; // On Safari, Event is sometimes emitted instead of ErrorEvent
             const errorText = errorName ? ` (${errorName})` : "";
-            setError(new Error(`An unexpected error occurred${errorText}. Trying to reconnect...`));
+            const errorBeginning = `An unexpected error occurred${errorText}. `;
+            const errorEnding = ws.retryCount < numRetries ? 'Trying to reconnect...' : 'Unable to reconnect; please refresh the page.';
```
ssciolla, line 60: I'm not exactly sure when `ws.retryCount` is incremented, so this may need some tweaking to ensure all three attempts have been made before the message changes.

jlost:
Looks good! Were you able to test it?

ssciolla:
> Looks good! Were you able to test it?

I only tested it by shutting down the `web` service; I didn't try to reproduce the Chrome problem.

jlost:
We know that the error message was displayed when the Chrome problem happened, so we know that this fix should apply equally in this case. So I think your form of testing is good enough.

src/assets/src/hooks/useWebSocket.ts
jlost: Yes, it's hard to validate the number of retries as there isn't an "onretry" event to attach to. I get three errors from `reconnecting-websocket-mjs.js:518` (WebSocket connection failed: ERR_CONNECTION_REFUSED), which seems to indicate three failed retries. Let's trust the library; the exact number of retries isn't important to the functionality aiming for anyway.

src/assets/src/hooks/useWebSocket.ts
```diff
@@ -8,13 +8,12 @@ interface OfficeHoursMessage<T> {
 
 export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onDelete?: (setError: (React.Dispatch<React.SetStateAction<Error | undefined>>)) => void) => {
     const [error, setError] = useState(undefined as Error | undefined);
+    const numRetries = 3;
```
jlost, line 11:
```suggestion
    const numRetries = 10;
```

I noticed the three retries exhaust pretty quickly. It might be good to allot more time for network interruptions.

jlost (approved):
Suggested changing maxRetries from 3 to 10, but otherwise looks good.

PR #196, opened 2020-10-29
Use messaging_service_sid instead of phone number (#193)

### LINKED ISSUE #193: Improve Twilio Scalability
Outgoing SMS messages have rate limits that you can get around by distributing messages between phone numbers.

Message rate limits:
https://support.twilio.com/hc/en-us/articles/115002943027-Understanding-Twilio-Rate-Limits-and-Message-Queues

SMS Services have a Scaler option that autoselects the `From` number from your pool of phone numbers, distributing them evenly.

https://www.twilio.com/docs/sms/services

Twilio Proxy has a number pooling feature.

https://www.twilio.com/docs/proxy/understanding-phone-number-management

Short codes enable 100 messages per second but are prohibitively expensive:

https://support.twilio.com/hc/en-us/articles/226460288-How-much-does-a-Short-Code-cost-

jlost:
Use SMS Services so we can leverage message pooling.

Closes #193


src/officehours_api/notifications.py
```diff
@@ -14,7 +14,7 @@
 
 
 def notify_next_in_line(next_in_line: Meeting):
-    if not (settings.TWILIO_ACCOUNT_SID and settings.TWILIO_AUTH_TOKEN and settings.TWILIO_PHONE_FROM):
+    if not (settings.TWILIO_ACCOUNT_SID and settings.TWILIO_AUTH_TOKEN and settings.TWILIO_MESSAGING_SERVICE_SID):
```
ssciolla, line 17: I think I mentioned this in my review of the initial notifications PR, but I'll say it again in case I failed to communicate well before. While the computational cost of this is probably low, it would make more sense to me to have these variable checks determine (somewhere else) whether to look for the relevant signals in the first place, rather than have the signals still function and then bail out of these functions when they trigger. That would also reduce some duplication. Maybe you could just move this check to where you imported the `notifications` module in `models.ply`.

ssciolla (approved):
The test suite runs locally without error, and I tested both notifications cases in `dev`, with everything working great. Just the one comment suggesting a minor refactor.

src/officehours_api/notifications.py
ssciolla: Since this is only really relevant for local development or testing environments where Twilio isn't configured, it's also not a big deal, probably in the nit-pick category.

jlost: That's a good idea. That module will now assume those settings are configured. I also managed to bring `twilio` up to the module level. I could go even farther in reducing duplication, but I think it would add more complexity than it removes. Things still function & tests still pass with & without the settings configured.

ssciolla: Looks good to me!

PR #200, opened 2020-11-02
Use blue focus indicator for .btn-warning (#158)

### LINKED ISSUE #158: Increase visibility of "Leave the line" focus indicator
"Leave the line" focus treatment is very subtle: 

image

The outline/highlight should be more visible and have sufficient contrast

jlost:
Increases contrast from orange color for accessibility.
Closes #158

### DIFF
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -182,7 +182,7 @@ a:hover {
 }
 
 .btn-warning:focus {
-	box-shadow: 0 0 0 0.2rem rgba(12, 61, 222, 0.5);
+	box-shadow: 0 0 0 0.2rem #0c3dde7f;
 }
 
 .bottom-right {
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -181,6 +181,10 @@ a:hover {
 	color: #0065a3;
 }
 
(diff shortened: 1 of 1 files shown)


jlost:
Here's how it looks:
image

src/officehours_ui/static/css/main.css
```diff
@@ -181,6 +181,10 @@ a:hover {
 	color: #0065a3;
 }
 
+.btn-warning:focus {
+	box-shadow: 0 0 0 0.2rem rgba(12, 61, 222, 0.5);
```
ssciolla, line 185: Can you use a hex value here? It may make our lives easier in the future as we try to control all the different colors in the palette.

ssciolla: The blue focus indicator looks okay to me, but if you wanted to stay closer to yellow, you could use a brown? I wasn't privy to all the discussion about what color the button should be to begin with. A blue button seems like it would be okay to me, but maybe it's not noticeable enough.

jlost: I think RGBA is much easier to visualize at a glance and tweak, but I appreciate that the rest of the file is hex and it's better to be consistent. I encoded the 50% alpha as an extra hex digit at the end. 343f483

@teshsopher Any input on the color?

Issue #159, opened 2020-09-10
Add labels for form elements that don't have them


teshsopher:
3 form elements currently don't have a label. In each case, the text input has no label, but, instead, has placeholder which isn't accessible.

Labels (even if it's just screen reader only text) need to be added to:
- Add Host
- Add Attendee
- Add Queue

teshsopher:
Tested in QA: 
- As an attendee in queue, when the Agenda text field is opened, the text field doesn't have a label
- The Name and Description fields in the Queue Settings page don't have labels 
- The Add Host field in the Queue Settings page doesn't have a label

For these, I think, visually, these are labelled, it's just that screen readers don't "see" the label, so the fields don't appear to have one. Adding an aria-label attribute should work here

ssciolla:
`aria-label` needs to be moved from the containing `form` element/component to the specific `input` element.

teshsopher:
Tested and working in QA

PR #201, opened 2020-11-02
Tidy up CSS to use reds, greens, and other formatting consistently, eliminate custom CSS classes (#159)

ssciolla:
This PR makes several changes to `main.css` in order to make the use of `danger` and `success` colors and badge formatting consistent across the application. It also transforms existing custom CSS classes into classes that will override the Bootstrap defaults (`form-control.is-invalid` and `badge`), which allowed for the removal of `className` usages in `common.tsx`. More details will be provided in a task list. The PR aims to resolve issue #159.

- [x] Use `#CC0000` for all solid red UI features and standalone text (includes buttons, badges, validation formatting and feedback, button text; it does not include alert background or text).
- [x] Use `#00820D` for all solid green UI features (includes buttons and badges; it does not include alert background). Note: `text-success` was not specified, as it is not in use anywhere (that I'm aware of).
- [x] Make all badge text use `font-size: 90%`,  which affects the badges in the queue tables and the "In Person" badge in the meetings list on the Manage page.
- [x] Fix underline rules introduced in PR  #191 so they don't apply to buttons in alerts.

### DIFF
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -202,6 +202,8 @@ p a:not(.btn) {
 	padding: 1.25rem;
 }
 
+/* !important is needed to override another !important used in the text-emphasis Bootstrap mixin.
+   We should re-evaluate this in the future and potentially customize the danger value using Sass. */
 .text-danger {
 	color: var(--custom-red) !important;
 }
--- a/src/officehours_ui/static/css/main.css
+++ b/src/officehours_ui/static/css/main.css
@@ -181,11 +181,11 @@ a:hover {
 	color: #0060ba;
 }
 
-.alert a {
+.alert a:not(.btn) {
 	text-decoration: underline;
 }
 
-p a {
+p a:not(.btn) {
 	text-decoration: underline;
 }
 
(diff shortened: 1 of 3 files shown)


src/officehours_ui/static/css/main.css
```diff
@@ -181,11 +181,11 @@ a:hover {
 	color: #0060ba;
 }
 
-.alert a {
+.alert a:not(.btn) {
 	text-decoration: underline;
 }
 
-p a {
+p a:not(.btn) {
```
ssciolla, line 188: Because I did not previously add these `:not` lines, I noticed that I had inadvertently added an underline to the "Return to Previous Queue" Button text on the home page, since it was inside an alert.

src/officehours_ui/static/css/main.css
```diff
@@ -200,9 +200,13 @@ p a {
 	padding: 1.25rem;
 }
 
+.text-danger {
+	color: #CC0000 !important;
```
ssciolla, line 204: Without the `!important`, this wasn't applying, there was some `text-emphasis` Bootstrap asset that was taking precedence, I think.

jlost: Is it possible to make this a more specific selector (adding selectors before it), or to move this elsewhere in the file to have it take precedence without `!important`? I try to avoid using it where possible, but sometimes it makes sense.

jlost (changes requested):

src/officehours_ui/static/css/main.css
ssciolla: Yeah, let me see what I can do.

jlost (approved):
I see that Bootstrap is using `!important` already, so the only way to get around it is with `!important` or a different class. Either approach is fine by me, but I'd probably just use `!important` as the intent is actually to override that class. We can try switching to Sass down the line to avoid having to override Bootstrap.

Issue #212, opened 2020-11-03
Prevent people from adding non-US/Canada (+1) cell phone numbers


mfldavidson:
For now, we will only allow US/Canada numbers (i.e. country code +1). If we get requests to allow other countries we can re-evaluate but this should cover most use cases. Additionally, it is confusing for users to type in country codes which they aren't used to and easily lead to entering invalid international numbers.

Ideal scenario is to ONLY allow area code plus phone number, don't even allow entry of country code, to avoid the above mentioned confusion. If not 10 digits, reject input and provide this error:
> Please enter a valid US/Canada (+1 country code) phone number with area code that can receive SMS messages.

mfldavidson:
Well, this is what Twilio said:
> You would need to use an international number if you are communicating between two different countries. You can not text a China number using a USA number, it has to be an international number. 
So for now let's stick with just US and Canada, and if tons of requests come in to support numbers from a particular country (probably China), we can evaluate whether/how to purchase a China number and how much complexity it would add. I think it's likely a technical no-go.

PR #214, opened 2020-11-04
Modify PhoneInput to limit to US/Canada, simplify user input, improve styling (#212)

ssciolla:
This PR makes a few changes to `preferences.tsx` and `validation.ts` in order to limit phone numbers entered to those from the US and Canada, as well as simplify the input. More details are provided in a task list below. The PR aims to resolve issue #212.

- [x] Specified `onlyCountries={['us', 'ca']}` in `PhoneInput` to limit the possible country codes to just `+1`.
- [x] Specified `countryCodeEditable={false}` in `PhoneInput` so users must use the dropdown menu to change the country code (eleviating some of the awkwardness of entering phone numbers, noted in #212) .
- [x] Combined the existing invalid phone number message in `validatePhoneNumber` with the new text provided by @mfldavidson .
- [x] Changed the CSS style from the `style` variant to `bootstrap` because it has better focus treatment for the country code dropdown menu, and (at least theoretically) should be more consistent aesthetically with the rest of the application.

### DIFF
--- a/src/assets/src/components/preferences.tsx
+++ b/src/assets/src/components/preferences.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import { useState, useEffect } from "react";
 import { Alert, Button, Form } from "react-bootstrap";
 import PhoneInput from "react-phone-input-2";
-import 'react-phone-input-2/lib/style.css'
+import 'react-phone-input-2/lib/bootstrap.css'
 
 import * as api from "../services/api";
 import { MyUser } from "../models";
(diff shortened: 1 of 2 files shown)


mfldavidson (approved):
Given what I just put in #212 about how we can't text from a US number to international numbers, let's move forward as-is. LGTM.

ssciolla:
@jlost, do you want to take a look at this? Changes are pretty minor. Let me know if you think the `bootstrap` change makes sense to you; basically, I couldn't see the focus treatment for the country dropdown with the other style (it was in the Tab order, but nothing was highlighted).

jlost (approved):

Issue #215, opened 2020-11-05
Expose phone number and SMS preferences in admin


mfldavidson:
Make phone number and SMS preferences visible and configurable in the Users section in admin

PR #218, opened 2020-11-05
Add Profile fields to User page on admin site (#215)

ssciolla:
This PR follows the [pattern detailed in the Django docs](https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#extending-the-existing-user-model) to extend the existing `UserAdmin` to include the fields from the `Profile` model ("Phone number", "Notify me attendee", and "Notify me host"). The PR aims to resolve issue #215.


jlost (approved):
LGTM

PR #223, opened 2020-11-09
Fix admin by removing "reverse" from module scope (#224)

### LINKED ISSUE #224: Unable to access admin site because of module-level reverse
During testing in QA, we discovered a bug that prevented us from accessing the `admin` site. This traceback was visible in the backend:

```
django.urls.exceptions.NoReverseMatch: Reverse for 'app_list' not found. 'app_list' is not a valid view function or pattern name.
```

After investigating, @jlost and I concluded that the issue was caused by the usage of `reverse` in `officehours_api/notifications.py`, specifically the use of `reverse` at the module level in `PREF_URL`, introduced by PR #210.

jlost:
Closes #224

### DIFF
--- a/src/officehours_api/notifications.py
+++ b/src/officehours_api/notifications.py
@@ -15,11 +15,14 @@ logger = logging.getLogger(__name__)
 twilio = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
 
 DOMAIN = Site.objects.get_current().domain
-PREF_URL = f"{DOMAIN}{reverse('preferences')}"
-ADDENDUM = (
-    f"\n\nYou opted in to receive these texts from U-M. "
-    f"Opt out at {PREF_URL}"
-)
+
+
+def build_addendum():
+    pref_url = f"{DOMAIN}{reverse('preferences')}"
+    return (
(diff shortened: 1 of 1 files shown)


src/officehours_api/notifications.py
```diff
@@ -15,11 +15,14 @@
...
 
 DOMAIN = Site.objects.get_current().domain
-PREF_URL = f"{DOMAIN}{reverse('preferences')}"
-ADDENDUM = (
-    f"\n\nYou opted in to receive these texts from U-M. "
-    f"Opt out at {PREF_URL}"
-)
+
+
+def build_addendum():
```
ssciolla, line 20: Would it be worth adding a comment or docstring explaining why we did this?

ssciolla: We both missed this initially, so someone else might not understand pick this up either.

ssciolla (commented):
LGTM! Just the one suggestion about the comment/explanation. I reproduced the error locally, and then confirmed that this fixed it. Thanks for cleaning up a bug I introduced!

ssciolla (approved):
Sorry, I meant to approve.

src/officehours_api/notifications.py
jlost: Good idea, added a comment.

PR #225, opened 2020-11-11
Update push scripts

jlost:
Adds a push script for the QA environment and removes the "deployment patch hack" because imagestream triggers now exist on the deploymentconfigs, rendering it unnecessary.

I was of a mind to remove these scripts but apparently they're still useful, so they might as well be kept in working order.


scripts/push.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
+# Login with `oc login ...` before running.
```
ssciolla, line 4: Rename to `push-dev.sh`?

scripts/push-prod.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
+# Login with `oc login ...` before running.
+oc project officehours
 docker login -u `oc whoami` -p `oc whoami -t` docker-registry.webplatformsunpublished.umich.edu
```
ssciolla, line 6: Nit-pick: `docker-registry.webplatformsunpublished.umich.edu` could be made into a variable, but it's not critical given the limited purpose of these scripts.

scripts/README.md
```diff
@@ -0,0 +1,13 @@
+# Scripts
+
+## Push Scripts
+
+The push scripts in this directory will build an image locally and push it to OpenShift's registry.
+They should only be used if remote builds in OpenShift are failing.
```
ssciolla, line 6: An idea: There's probably a way to combine all these scripts into one, passing the project name at the command line. Something like this? Syntax may be a bit off.

```
REG_BASE = "docker-registry.webplatformsunpublished.umich.edu"
oc project "$1"
docker login -u `oc whoami` -p `oc whoami -t` "$REG_BASE"
docker build -t  "$REG_BASE"/"$1"/officehours .
docker push  "$REG_BASE"/"$1"/officehours
```

ssciolla (approved):
I left some re-factoring ideas, but I wouldn't rate them as high priority, of course.

Did you do some testing locally? I did not.

jlost:
I renamed `push.sh` to `push-dev.sh`, but will put off the deeper refactors for another time since these scripts are rarely-used hacks.

PR #238, opened 2020-11-23
Refactor Zoom and BlueJeansMeetingInfo into VideoMeetingInfo

ssciolla:


src/assets/src/components/meetingType.tsx
```diff
@@ -2,6 +2,13 @@ import * as React from "react";
 import { Form } from "react-bootstrap";
 
 
+export enum MeetingType {
+    inperson = "In Person",
+    bluejeans = "BlueJeans",
+    zoom = "Zoom"
+}
+
```
ssciolla, line 10: I'm honestly not sure if this is necessary, but for `VideoMeetingInfo`, it seemed nice not to have to use string comparison to check the backend type, while also being able to use the friendly name stored as a value.

jlost (approved):
Looks good. The state of backend-specific logic in the frontend has always bugged me, and this moves things in the right direction. It might make sense to refactor this even more in the future; it would be nice to have a design parallel between how the Django & React layers handle backends.

Issue #244, opened 2020-12-03
Filter out started meetings when calculating Meeting.line_place


ssciolla:
Right now, started/created/in-progress meetings seem to retain their place in line, which I believe gives waiting users a hazy sense of where they are in queue. We could remedy this by filtering out started meetings when calculating `line_place`. I'm thinking a change like the following:

```
    @property
    def line_place(self):
        if not self.queue:
            return None
        meetings = self.queue.meeting_set.order_by('id')
        unstarted_meetings: List[Meeting] = [meeting for meeting in meetings if meeting.status != MeetingStatus.STARTED]
        if self not in unstarted_meetings:
            return None
        for i in range(0, len(unstarted_meetings)):
            m = unstarted_meetings[i]
            if m == self:
                return i
```

PR #129, opened 2020-08-12
Zoom Backend

jlost:
Basic Zoom implementation. Includes three-legged OAuth integration. Includes just-in-time meeting creation.

PR #246, opened 2020-12-03
Modify Meeting.line_place to improve accuracy for waiting users (#244)

ssciolla:
Building on the introduction of a `MeetingStatus` `enum` introduced in PR #129, this PR attempts to improve the accuracy of `line_place` information by only assigning a place number if the meeting has not been started (consequently `Meeting.line_place` is left as `None`/`null` for started meetings). The PR aims to resolve issue #244.

### DIFF
--- a/src/officehours_api/models.py
+++ b/src/officehours_api/models.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from typing import List
+from typing import List, Optional
 
 from django.conf import settings
 from django.db import models
@@ -184,7 +184,7 @@ class Meeting(SafeDeleteModel):
         super().delete(*args, **kwargs)
 
     @property
-    def line_place(self):
+    def line_place(self) -> Optional[int]:
         if not self.queue:
             return None
         meetings = self.queue.meeting_set.order_by('id')
--- a/src/officehours_api/models.py
+++ b/src/officehours_api/models.py
@@ -191,12 +191,11 @@ class Meeting(SafeDeleteModel):
         unstarted_meetings: List[Meeting] = [
             meeting for meeting in meetings if meeting.status != MeetingStatus.STARTED
         ]
-        if self not in unstarted_meetings:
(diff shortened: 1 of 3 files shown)


ssciolla:
@jlost, I searched for `line_place` and I didn't see any much that this would break. `notifications.py` uses it, but those comparisons are `==`, so those should fail properly when it's `None`. I made quick fixes to the issues I saw in `queue.tsx`, but I"ll return to those after the rebase. Please let me know if I'm forgetting something...

src/officehours_api/models.py
```diff
@@ -183,12 +184,17 @@ def delete(self, *args, **kwargs):
...
         if not self.queue:
             return None
         meetings = self.queue.meeting_set.order_by('id')
-        for i in range(0, len(meetings)):
-            m = meetings[i]
+        unstarted_meetings: List[Meeting] = [
+            meeting for meeting in meetings if meeting.status != MeetingStatus.STARTED
+        ]
+        if self not in unstarted_meetings:
+            return None
```
jlost, line 195: It looks like these two lines wouldn't do anything except iterate through the list an extra time and removing them would retain the same behavior, since methods return `None` by default (i.e. when no `return` is specified). They do have the value of making it perfectly explicit what happens when `self` isn't there, which is good, and these lists would rarely be long, so there is probably no noticeable performance implication to iterating twice. Just something to be aware of.

jlost (approved):

src/officehours_api/models.py
ssciolla: Nice point, @jlost. If we want to be explicit but not iterate an extra time, could I delete these lines and put `return None` after the loop that was already there?

jlost: Yup, absolutely

ssciolla: Cool, done.

PR #249, opened 2020-12-03
Prompt 3-legged OAuth on host page only (#241)

### LINKED ISSUE #241: Prompt Zoom 3-legged auth only on Edit page
Instead of prompting for Zoom 3-legged oauth when any page is visited, prompt only when the Edit page is visited. That way, only hosts end up authorizing Zoom.

Added by Maggie: only prompt for Zoom 3-legged oauth when the queue has Zoom as an allowed meeting type.

jlost:
Instead of redirecting to authorize Zoom upon any page view, only redirect when they visit the `/manage/{id}` route, and only if Zoom is an enabled backend for that queue. This solution should gracefully handle situations where an existing queue has Zoom enabled while hosts are on the page. 

Closes #241

### DIFF
--- a/src/assets/src/components/queueManager.tsx
+++ b/src/assets/src/components/queueManager.tsx
@@ -21,7 +21,7 @@ import {
 } from "../models";
 import * as api from "../services/api";
 import { useQueueWebSocket, useUserWebSocket } from "../services/sockets";
-import { recordQueueManagementEvent, redirectToBackendAuth, redirectToLogin } from "../utils";
+import { checkBackendAuth, recordQueueManagementEvent, redirectToLogin } from "../utils";
 import { confirmUserExists, uniqnameSchema } from "../validation";
 
 
@@ -257,15 +257,6 @@ const MeetingInfoDialog = (props: MeetingInfoDialogProps) => {
     );
 }
 
-const backendAuthCheck = (user: MyUser, queue: QueueHost) => {
-    for (const backend of queue.allowed_backends) {
-        const authorized = user.authorized_backends[backend];
-        if (authorized === false) {
-            redirectToBackendAuth(backend);
-        }
-    }
-}
-
 interface QueueManagerPageParams {
     queue_id: string;
 }
@@ -301,7 +292,7 @@ export function QueueManagerPage(props: PageProps<QueueManagerPageParams>) {
     const userWebSocketError = useUserWebSocket(props.user!.id, (u) => setMyUser(u as MyUser));
 
     if (myUser && queue) {
-        backendAuthCheck(myUser, queue);
+        checkBackendAuth(myUser, queue);
     }
 
     // Set up API interactions
--- a/src/assets/src/components/queueManager.tsx
+++ b/src/assets/src/components/queueManager.tsx
@@ -17,11 +17,11 @@ import { PageProps } from "./page";
 import { usePromise } from "../hooks/usePromise";
 import { useStringValidation } from "../hooks/useValidation";
 import {
-    BluejeansMetadata, isQueueHost, Meeting, MeetingBackend, MeetingStatus, QueueAttendee, QueueHost, User, ZoomMetadata
+    BluejeansMetadata, isQueueHost, Meeting, MeetingBackend, MeetingStatus, QueueAttendee, QueueHost, User, MyUser, ZoomMetadata
 } from "../models";
 import * as api from "../services/api";
-import { useQueueWebSocket } from "../services/sockets";
-import { recordQueueManagementEvent, redirectToLogin } from "../utils";
+import { useQueueWebSocket, useUserWebSocket } from "../services/sockets";
+import { recordQueueManagementEvent, redirectToBackendAuth, redirectToLogin } from "../utils";
 import { confirmUserExists, uniqnameSchema } from "../validation";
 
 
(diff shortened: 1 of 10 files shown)


ssciolla:
@jlost, can you resolve the conflict? I can start looking at this.

src/officehours_api/backends/zoom.py
```diff
@@ -195,38 +196,29 @@ def auth_callback(cls, request):
...
         return {
-            'name': self.name,
-            'friendly_name': self.friendly_name,
-            'docs_url': self.docs_url,
-            'telephone_num': self.telephone_num
+            'name': cls.name,
+            'friendly_name': cls.friendly_name,
+            'docs_url': cls.docs_url,
+            'telephone_num': cls.telephone_num
         }
```
ssciolla, line 220: Nitpick: Do you prefer `cls`? I don't mind you making this consistent, but you didn't change the other `get_public_data` methods, and I'm seeing `self` and `cls` throughout these changes.

src/assets/src/components/queueManager.tsx
```diff
@@ -404,6 +404,13 @@ export function QueueManagerPage(props: PageProps<QueueManagerPageParams>) {
     const queueWebSocketError = useQueueWebSocket(queueIdParsed, setQueueChecked);
     const [visibleMeetingDialog, setVisibleMeetingDialog] = useState(undefined as Meeting | undefined);
 
+    const [myUser, setMyUser] = useState(undefined as MyUser | undefined);
+    const userWebSocketError = useUserWebSocket(props.user!.id, (u) => setMyUser(u as MyUser));
+
+    if (myUser && queue) {
+        checkBackendAuth(myUser, queue);
+    }
+
```
ssciolla, line 413: I think this is okay for now, but I wonder if adding another WebSocket connection to this page is a somewhat heavy solution. Did you explore passing `authorized_backends` through `globals`?

src/assets/src/utils.ts
```diff
@@ -18,6 +19,24 @@ export const redirectToSearch = (term: string) => {
...
+        action: `Redirected to ${backend} Auth`,
+        nonInteraction: true,
+    });
+    location.href = `/auth/${backend}/?state=${location.pathname}`;
+}
+
+export const checkBackendAuth = (user: MyUser, queue: QueueHost) => {
+    for (const backend of queue.allowed_backends) {
+        const authorized = user.authorized_backends[backend];
+        if (authorized === false) {
```
ssciolla, line 34: Nit-pick: This could be simplified to `!authorized`?

ssciolla (commented):
Looks pretty good. I will try running this, but here are some initial comments.

src/assets/src/components/queueManager.tsx
jlost: I had a few things in mind when I went this way...
* If you have multiple tabs open, your zoom auth wouldn't be noticed until you refresh the page, unless you're getting a fresh User via API or websocket.
* We might eventually want to move away from the current redirect and instead pop up a modal without leaving the SPA. The modal would have the same authorize link, except it would open in a new tab, and the tab would close when the flow is complete. This would handle users with multiple open tabs or queues with multiple auth-required backends better, because you could display a modal that has multiple authorize links, and it can vanish automatically when all of the backends are authorized.
* I've been thinking that, at a high level, it would be convenient to have a global User websocket always open no matter which page you're on. In which case, it would make sense to source authorized_backends from there. I mimicked that in a way that involved less refactoring.

In this case, if I replaced this websocket with a call to the User API, it would be functionally the same, except in an extreme edge case:  A non-Zoom auth'd user has two tabs open, `/manage/0` and `/manage/1`; `0` toggles Zoom to be an `allowed_backend`; the user is prompted and completes the auth; then `1` toggles Zoom to be an `allowed_backend`. If using websockets, `1` correctly updates itself with Zoom as a `authorized_backend`. If using the API, `1` prompts for Zoom auth again. (Not likely to ever happen).

IOW, given 3 choices that were only narrowly different, I picked the one that looked closest to where things might eventually end up.

ssciolla: I think I follow all of that; it does seem like we'd benefit from a WS in the multiple tabs situation. I guess I was a concerned about adding yet another way to get user info that could get out of sync, but `globals` doesn't give you much that would change anyway. I was also thinking you only need the update once, so why maintain a WS connection after that, but it may not matter much performance-wise if nothing changes.

src/assets/src/utils.ts
jlost: Depends what kind of behavior we want in the event queues are created with a backend, and then the backend is disabled for the application instance. This code would then get `const authorized = undefined`. If using `!authorized`, you get redirected to an error page every time you visit the queue. As I currently have it, the redirect is skipped, and the page might act funny in other ways, but at least you can get a better idea of what's going on.

src/officehours_api/backends/zoom.py
jlost: Nice catch, I updated those to make them all consistent. I don't feel strongly about it, but I originally went with the `cls` on `@classmethod` convention because it clarifies that you're not getting an instance.

src/assets/src/utils.ts
ssciolla: Oh, I missed some nuance here. We can keep this as you have it, though it may be worth talking about this edge case more in the future.

src/officehours_api/backends/zoom.py
ssciolla: TBH, I also missed some nuance here, I'm still learning about `@classmethod`. Your approach is sound.

jlost:
@ssciolla I manually rebased this on master.

src/officehours_api/models.py
```diff
@@ -41,6 +41,13 @@ class Profile(models.Model):
     notify_me_host = models.BooleanField(default=False)
     backend_metadata = JSONField(null=True, default=dict)
 
+    @property
+    def authorized_backends(self):
+        return {
+            backend.name: backend.is_authorized(self.user)
+            for backend in backend_instances.values()
```
ssciolla, line 48: Nit-pick: This could be `BACKEND_INSTANCES`.

ssciolla (approved):
Looks great to me! Tested locally.

PR #252, opened 2020-12-08
Auto-assign for single-host queues (#227)

### LINKED ISSUE #227: Auto-assign host on single-host queue
If a queue has only one host, auto-assign all meetings in the queue to the only host

jlost:
Auto-assigns new meetings in single-host queues; otherwise, retains old behavior.
Closes #227

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -18,7 +18,7 @@ import { PageProps } from "./page";
 import { usePromise } from "../hooks/usePromise";
 import * as api from "../services/api";
 import { useQueueWebSocket, useUserWebSocket } from "../services/sockets";
-import { redirectToLogin } from "../utils";
+import { addMeetingAutoAssigned, redirectToLogin } from "../utils";
 import { meetingAgendaSchema } from "../validation";
 
 
@@ -417,7 +417,7 @@ export function QueuePage(props: PageProps<QueuePageParams>) {
             category: "Attending",
             action: "Joined Queue",
         });
-        await api.addMeeting(queueIdParsed, props.user!.id, backendType);
+        await addMeetingAutoAssigned(queue!, props.user!.id, backendType);
     }
     const [doJoinQueue, joinQueueLoading, joinQueueError] = usePromise(joinQueue);
     const leaveQueue = async () => {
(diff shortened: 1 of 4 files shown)


src/assets/src/components/queue.tsx
```diff
@@ -417,7 +417,7 @@ export function QueuePage(props: PageProps<QueuePageParams>) {
             category: "Attending",
             action: "Joined Queue",
         });
-        await api.addMeeting(queueIdParsed, props.user!.id, backendType);
+        await addMeetingAutoAssigned(queue!, props.user!.id, backendType);
```
ssciolla, line 420: This isn't a quibble with your code, but just with the overall flow and logic of the user experience here: For the usages of `addMeetingAutoAssigned` in `queue.tsx`, auto-assigning the host will basically lock the attendee out of changing the meeting type after they join the queue (since `changeMeetingType` is hidden once the the Meeting has an `assignee`. I think this might be problematic.

We should discuss whether we want the auto-assignment on both sides of the UI, and also at what point it is appropriate to lock `changeMeetingType`. A change to one of these would probably resolve this issue. Sorry, I know I was one of the people who suggested this.

ssciolla (commented):
I think the code looks good, but I think we may need to pause to think about whether we want this to lock out attendee's from changing the meeting type (see comment).

src/assets/src/components/queue.tsx
jlost: I noticed that as well, but since Meeting Type is put prominently in front of you before you click Join Queue, I didn't think it very problematic to lock them in. I wonder how often users actually need to switch meeting types? And how often is it after they've waited in the queue for a while? (Because if they noticed they picked the wrong thing, they could just leave the queue and join again, and if they're already at the end of the line they're not losing anything)

ssciolla: Some good points -- I'm not sure how often folks would actually want to change the type. I do wonder though whether people in that situation would think to leave the line and restart; my gut says they would first want to change their existing preference, struggle, and then maybe realize they can they do what you suggested. @teshsopher, any thoughts?

teshsopher: From what I understand, most attendees will have the queue page open in the background of what they're doing and check on it periodically, so that can cause them to miss any mistakes they may have made in choosing a particular meeting type. It's hard to know how often changing the meeting type occurs. However, particularly for single host queues, I see this as a lower risk case, since attendees can re-join or contact the host to explain the situation. So I'm fine with locking the meeting type in this case.

ssciolla: Okay, let's keep it at is for now, then. Thanks for the discussion.

ssciolla (approved):
I'm approving, since we think locking the change meeting type action will be okay in this context.

Issue #233, opened 2020-11-19
JIT VC meetings - attendee information


mfldavidson:
Before the meeting is created, the Join Meeting button should be grayed out.

Before the meeting is created, show this message just above "Join Meeting" button (where it currently says "The host will join...")
> The host has not created the meeting yet. You will be able to join the meeting once it is created. We'll show a message in this window when it is created -- pay attention to the window so you don't miss it!

Once the meeting is created, show this message just above "Join Meeting" button (where it currently says "The host will join...")
> The host has created the meeting. Join it now! The host will join when they are ready for you.

![Joining the Meeting and Having Trouble with Video info boxes](https://user-images.githubusercontent.com/31591389/99694175-202a5300-2a5a-11eb-8f7a-cf81e023f91d.png)

Joining the Meeting box should be:
> #### Joining the Meeting
> Once the meeting is created, click Join Meeting to join the meeting and wait for the host. Download the app and test your audio now. Refer to [link] for additional help getting started.
The link and text for BlueJeans should be: [Getting Started with BlueJeans at U-M](https://its.umich.edu/communication/videoconferencing/blue-jeans/getting-started)
The link and text for Zoom should be: [Getting Started with Zoom at U-M](https://its.umich.edu/communication/videoconferencing/zoom/getting-started)

Having Trouble with Video? box should remain the same for BlueJeans.
The same box for Zoom should be (differences are the phone #, the link to international numbers, and terminology of meeting host and host key in last sentence):
> #### Having Trouble with Video?
> Having problems with video? As a back-up, you can call 1.312.626.6799 from the USA (or [find your international number to call in from outside the USA](https://umich.zoom.us/profile/setting?tab=telephony) - click See All Numbers under Toll Call to see all countries) from any phone and enter [id]#. You do not need a host key or participant ID.

Also add same having trouble with video blurb to host join info if meeting type is Zoom

ssciolla:
@mfldavidson, do we know the international number link for Zoom yet?

mfldavidson:
I apologize for not coming back to this @ssciolla thank you for the reminder. Unfortunately there is no one link we can use that shows all numbers for OUR account unless someone is logged in. So I think it will have to be as follows, which is ugly:

> #### Having Trouble with Video?
> Having problems with video? As a back-up, you can call 1.312.626.6799 from the USA* from any phone and enter [id]#. You are not a meeting host, so you do not need a host key.
[...]

Issue #247, opened 2020-12-03
Change leave the line link for attendees once meeting is in progress


mfldavidson:
When the meeting is in progress, "Leave the Line" on attendee view should change to "Cancel my meeting"

Issue #235, opened 2020-11-19
Remove join meeting blurb if meeting is in-person


mfldavidson:
The below message shows even when the meeting type is in person. We should just remove the entire message if the meeting is in-person. Note that this message is being changed for VC meetings in #233
> The host will join the meeting when it is your turn. We'll show a message in this window when your turn is coming up -- keep an eye on the window so you don't miss it!

PR #254, opened 2020-12-08
Refactor meeting info components, adjust language and conditional logic for JIT meeting creation (#233, #235, #247)

ssciolla:
This PR makes requested changes to the text and logic of the attendee-facing queue UI. It also adds an `intl_telephone_url` to backend objects served up to the frontend for use in the new `BlueJeansDialInMessage` and `ZoomDialInMessage`. The PR aims to resolve issues #233, #235, and #247.

### DIFF
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -443,7 +443,7 @@ export const LoginDialog = (props: LoginDialogProps) =>
 
 interface OneTouchDialLinkProps {
     phone: string; // "." delimited
-    meetingNumber?: string;
+    meetingNumber: string;
 }
 
 const OneTouchDialLink = (props: OneTouchDialLinkProps) => (
@@ -465,7 +465,6 @@ const IntlTelephoneLink = (props: IntlTelephoneLinkProps) =>  {
 }
 
 export type DialInMessageProps = OneTouchDialLinkProps & IntlTelephoneLinkProps;
-const meetingNumberPlaceholder = 'the meeting number (that will appear here once the meeting is created) followed by ';
 
 export const BlueJeansDialInMessage = (props: DialInMessageProps) => {
     const phoneLinkUsa = <OneTouchDialLink {...props} />;
@@ -473,7 +472,7 @@ export const BlueJeansDialInMessage = (props: DialInMessageProps) => {
         <span>
             Having problems with video? As a back-up, you can call {phoneLinkUsa} from the USA
             (or <IntlTelephoneLink {...props} />)
-            from any phone and enter {props.meetingNumber ?? meetingNumberPlaceholder}#.
+            from any phone and enter {props.meetingNumber}#.
             You are not a moderator, so you do not need a moderator passcode.
         </span>
     );
@@ -485,7 +484,7 @@ export const ZoomDialInMessage = (props: DialInMessageProps) => {
         <span>
             Having problems with video? As a back-up, you can call {phoneLinkUsa} from the USA
             (or <IntlTelephoneLink {...props} /> - click See All Numbers under Toll Call to see all countries)
-            from any phone and enter {props.meetingNumber ?? meetingNumberPlaceholder}#.
+            from any phone and enter {props.meetingNumber}#.
             You do not need a host key or participant ID.
         </span>
     );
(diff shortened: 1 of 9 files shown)


src/assets/src/components/common.tsx
```diff
@@ -441,29 +441,54 @@ export const LoginDialog = (props: LoginDialogProps) =>
...
+    const phoneLinkUsa = <OneTouchDialLink {...props} />;
+    return (
+        <span>
+            Having problems with video? As a back-up, you can call {phoneLinkUsa} from the USA
+            (or <IntlTelephoneLink {...props} /> - click See All Numbers under Toll Call to see all countries)
+            from any phone and enter {props.meetingNumber ?? meetingNumberPlaceholder}#.
+            You do not need a host key or participant ID.
+        </span>
+    );
 }
```
ssciolla, line 492: I don't know if this is the best way to do this, but I didn't really want to include a bunch of conditions inside a common `DialInMessage`. There is some repeated language, but we'll be ditching `BlueJeansDialInMessage` at some point I guess.

src/assets/src/components/common.tsx
```diff
@@ -441,29 +441,54 @@ export const LoginDialog = (props: LoginDialogProps) =>
...
+    const phoneLinkUsa = <OneTouchDialLink {...props} />;
     return (
         <span>
-            Having problems with video? As a back-up, you can call {phoneLinkUsa} from the USA 
-            (or <a target="_blank" href="https://www.bluejeans.com/premium-numbers"> find your international number to call in from outside the USA</a>) 
-            from any phone and enter {props.meetingNumber}#.
+            Having problems with video? As a back-up, you can call {phoneLinkUsa} from the USA
+            (or <IntlTelephoneLink {...props} />)
+            from any phone and enter {props.meetingNumber ?? meetingNumberPlaceholder}#.
+            You are not a moderator, so you do not need a moderator passcode.
```
ssciolla, line 477: We need to check whether hosts are BJ moderators, but I do not think they are. This is relevant because the same `DialInMessage` components are used on both the attendee and host sides (in the Join Info dialog).

src/assets/src/components/queue.tsx
```diff
@@ -139,61 +140,63 @@ const WaitingTurnAlert = (props: WaitingTurnAlertProps) => {
...
+            meetingNumber: props.metadata.numeric_meeting_id,
+            intlNumbersURL: props.backend.intl_telephone_url
+        } as DialInMessageProps;
+
+        dialInMessage = props.backend.name === 'zoom'
+            ? <ZoomDialInMessage {...dialInProps} />
+            : props.backend.name === 'bluejeans'
+                ? <BlueJeansDialInMessage {...dialInProps} />
+                : null;
+    }
```
ssciolla, line 171: I made a decision here, but I'm happy to walk it back. The "Having Trouble with Video?" card just doesn't seem that useful until the meeting has been created (i.e. it has a meeting number). Perhaps there is alternative text we could use before the meeting is created...

src/assets/src/components/common.tsx
jlost: It looks like the Meeting's `assignee` goes into the API call as `POST /v1/user/{assignee_bj_user_id}/scheduled_meeting`, and it looks like that does make them a moderator, because if I create a BJ meeting that way and then visit my BJ dashboard (https://bluejeans.com/scheduling/) I can see the Moderator Passcode.

jlost: It looks like the Meeting's assignee goes into the API call as POST /v1/user/{assignee_bj_user_id}/scheduled_meeting, and it looks like that does make them a moderator, because if I create a BJ meeting that way and then visit my BJ dashboard (https://bluejeans.com/scheduling/) I can see the Moderator Passcode.

ssciolla: Darn. But it doesn't require either the host or attendee to enter a moderator passcode when you call in, does it? Maybe we could have language similar to Zoom's "You do not need a host key or participant ID." Something like, "You do not need a passcode to enter the meeting"?

jlost: Right, it doesn't require a passcode. That language sounds fine.

src/assets/src/components/queue.tsx
jlost: Seems to make sense to me.

jlost (approved):
The code looks good, and I also poked around on the page and it seemed to work as intended.

src/assets/src/components/common.tsx
jlost: I think it's fine, sometimes a little copypaste can be better than adding abstractions.

ssciolla: Cool

src/assets/src/components/common.tsx
ssciolla: Done. I chose to use "join the meeting" instead.

src/assets/src/components/queue.tsx
ssciolla: Cool, I will ask for more feedback during the QA process.

Issue #157, opened 2020-09-10
Announce changes to the list on the "My Queues" page


teshsopher:
The `.list-group` where queues are displayed, async needs an aria-live, or some other means of notifying users that the list has changed (i.e. user has been added or removed from a particular queue while they are on this page)

ssciolla:
Hi @teshsopher, I started looking at this. Adding `aria-live='polite'` to the `div` containing the table seems to do some of the things we want, but doesn't do other things that seem important. Here's what I'm observing when testing with Brave and Mac's VoiceOver:

* New queue appears in list: The new row in the table is announced, and all the cells and their labels are read (which seems ideal from my perspective).
* Queue disappears from list: Nothing is announced. (not ideal).
* Contents of a queue table row change (e.g. status goes from "Closed" to "Open", or name changes): the cell that changed is announced, but it doesn't seem to provide the current value. From what I can tell, it's just announcing whatever the value of the changed cell was when it was first loaded. So if it was "closed", it just repeatedly announces "closed" whenever the status is toggled. (not ideal).

With all this in mind, do we want to add `aria-live` here, or do we want to look for a better, probably more complex solution that will announce exactly what changes are occurring?

teshsopher:
Given what you've found, it's probably better to look for a solution that gets us closer to the ideal state (where important updates are announced in a clear/accurate way). Given some of the complexity/difficulty of this (VoiceOver tends to be a bit of a wild card when it comes to screen readers, I've noticed), it doesn't have to be in this release, but it is something I think is important.

In the meantime, I'll reach out to the a11y/accessibility team to see if they have any suggestions or resources to help us with this!

ssciolla:
I meant to add the below comment to this issue, not #160.

Issue #163, opened 2020-09-11
Make page updates from WebSockets accessible


ssciolla:
In either case, we would need there to be an ARIA notification to tell users what information has changed when the table updates (e.g. if two queues close while the user is on this page)

_Originally posted by @teshsopher in https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/153#discussion_r486331226_

This is a good point, and something that we should have a application-wide strategy for. Seems like `aria-live` regions may be useful here, but it appears there may be some complexity doing this in React.

ssciolla:
This may be a duplicate of #157, though it could be considered more expansive, covering updates to all pages with WSs.

ssciolla:
My work in progress on this issue (and child issues #157 and #160) can be seen here: https://github.com/tl-its-umich-edu/remote-office-hours-queue/compare/master...ssciolla:issues-157-160-change-event-alerts

Issue #160, opened 2020-09-10
Announce success for actions on queue pages


teshsopher:
On the individual queue pages, certain actions should be announced (for all users).

Areas that need announcing:
- Copy needs to be announced (for screen readers). Keep the focus on the copy button, or add a role="alert" with the message (it need not be visually present)
- Edit in place fields (description and title)
- Add host
- Add attendee

Note: Some changes (such as the edit in place fields) may be unnecessary for when the Queue Settings page(s) are up.

ssciolla:
MDN mentions the existence of an `aria-relevant` attribute where you can specify `removals` as an action to be announced: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-relevant_attribute

However, this doesn't seem to work with VoiceOver. This site, which I don't think I've seen before, specifies that support for `aria-relevant` can be spotty: https://a11ysupport.io/tests/tech__aria__aria-relevant#assertion-aria-aria-relevant_attribute-convey_value_text-html-div_element-vo_ios-ios_saf

ssciolla:
@teshsopher, I could use some clarification on how you want "Add Attendee" and "Add Host" announced. I could add `aria-live='polite'` to their containing `div`s (which would lead to screen-readers reporting the new items after they are added, I believe), but when you say "all users", I get the feeling you want there to be an alert or some other message indicating something has happened.

I guess clarity on what you mean ideally by "announced" would help me determine how to address all these.

ssciolla:
>     * Copy needs to be announced (for screen readers). Keep the focus on the copy button, or add a role="alert" with the message (it need not be visually present)

Part of this looks like it is already done. @jlost added a screenreader alert a while ago that does announce "Copied" after the action. The focus does move to the URL after clicking the button. I could probably keep the focus on the button, if desired. Please let me know if you want that or anything else done to this field.

teshsopher:
"Add Host" is probably not necessary now, since that has moved to the Queue Settings, and isn't on the main page anymore. For "Add Attendee," it would be nice if, when an attendee joined, that was announced to the screen reader, either in the form of "[user] has joined the queue" or "[number] of attendees joined queue" (or something to that effect). The main concern is that folks who can't see the table won't know when someone joins the queue (but some of this work kind of reminds me of the request for an audio notification for when someone joins).

[...]

PR #258, opened 2020-12-09
Implement change alerts on Manage Queues and Queue Manager pages (#157, #160, #163)

ssciolla:
This PR introduces a front-end mechanism for tracking changes to certain entity types (so far, `QueueBase`, `Meeting`) that are updated by WebSocket connections. These changes seek to alert all users (including those with vision impairment) that contents of pages have changed without their doing anything, as well as provide confirmation in some cases that certain actions were successful (specifically, with meeting actions on the `QueueManagerPage`). Alerts are set to be visible for seven seconds before disappearing. The PR aims to resolve issues #157, #160, and #163.


ssciolla:
@jlost, I think this is functional, but it may be too ambitious to try to squeeze it in this release. I thought I'd start the conversation about the code though so we can keep it moving forward.

src/assets/src/changes.ts
```diff
@@ -0,0 +1,90 @@
...
+            return `The ${propName} changed from "${valueOne}" to "${valueTwo}".`;
+        }
+    }
+    return;
+}
+
+
+// https://lodash.com/docs/4.17.15#xorWith
+
+export function compareEntities<T extends Base> (oldOnes: T[], newOnes: T[]): string | undefined
```
jlost, line 45: `Base` can be a lot of things, but only the following types are supported in this function: `Meeting`, `Queue`. You could tighten up the typing here by further constraining your generic constraint to `T extends Manage|QueueBase`.

jlost: You could also say `type ComparableEntity = Meeting | QueueBase` then make this function non-generic, replacing `T` with `ComparableEntity`. The outcome would be the same. When I see a function that uses generic types, my assumption is that it doesn't make any attempt to figure out exactly what you passed in (IOW, as written I'd expect this function to care about properties in `Base` and nothing else).

jlost: The other direction you could go in would by making these functions even more generic, factoring out the type checks somehow. I find that recusively comparing JS objects to diff them is a problem I run into surprisingly often, which makes me wonder if there are libraries. A cursory google search shows there are some, but I haven't tried them yet. They might be worth considering. Otherwise, if they don't look appropriate or it would take too much work to pay off, I think this general approach is OK.

ssciolla: Nice points, let me think this over and try out the changes you suggest.

ssciolla: As far as packages, I've found:
* https://github.com/flitbit/diff
* https://github.com/cosmicanant/recursive-diff

Did you see any others? The first looks fairly reliable, though it hasn't been updated since 2018.
The second is much less widely used.

ssciolla: > You could also say `type ComparableEntity = Meeting | QueueBase` then make this function non-generic, replacing `T` with `ComparableEntity`. The outcome would be the same.

I might be betraying my hazy understanding of generics, but question: is there any merit to keeping the generic because otherwise the function could allow `oldOnes` and `newOnes` to be different types of objects if their types are each `ComparableEntity[]`?

Shatterbrain: Hard to say which lib is best without doing a deep dive, but I tend to gravitate toward things that have more uses, stars, contributors, etc. and `flitbit/diff` seems to have quite a lead there. Its types seem to be in `@types/deep-diff`. 

>is there any merit to keeping the generic because otherwise the function could allow oldOnes and newOnes to be different types

Absolutely, good point, keeping it generic makes sure both params are the same type.

jlost: Hard to say which lib is best without doing a deep dive, but I tend to gravitate toward things that have more uses, stars, contributors, etc. and `flitbit/diff` seems to have quite a lead there. Its types seem to be in `@types/deep-diff`.

> is there any merit to keeping the generic because otherwise the function could allow oldOnes and newOnes to be different types

Absolutely, good point, keeping it generic makes sure both params are the same type.

ssciolla: Ah, I must have missed something regarding TypeScript and `deep-diff`. That looks promising; I will investigate that further. `deep-object-diff` was a little unpredictable, and wasn't giving me enough info.

Regarding your `ComparableEntity` suggestion and generics, see commit https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/258/commits/1bd1d2049a12b02e0afe53df5f52c64325e85405

src/package.json
```diff
@@ -5,6 +5,8 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.28",
     "@fortawesome/free-solid-svg-icons": "^5.13.0",
     "@fortawesome/react-fontawesome": "^0.1.9",
+    "lodash.isequal": "^4.5.0",
+    "lodash.xorwith": "^4.5.0",
```
jlost, line 9: I'd also be OK with including all of `lodash`, since I happen to like it and we have no need to economize on bundle size.

src/assets/src/components/queueManager.tsx
```diff
@@ -159,14 +165,19 @@ function QueueManager(props: QueueManagerProps) {
...
             </Row>
             <Row noGutters className={spacingClass}>
                 <Col md={2}><div id='created'>Created</div></Col>
                 <Col md={6}><div aria-labelledby='created'><DateDisplay date={props.queue.created_at} /></div></Col>
             </Row>
+            <Row noGutters className={spacingClass}>
+                <Col md={12}>
+                    <ChangeLog changeEvents={props.meetingChangeEvents} deleteChangeEvent={props.deleteMeetingChangeEvent} />
+                </Col>
+            </Row>
```
jlost, line 180: Not really a technical comment but a stray observation - since these are positioned toward the top, the rest of the page shifts downward a bit when they appear, which might cause misclicks. Not sure what the best way to handle that would be.

src/assets/src/hooks/useEntityChanges.ts
```diff
@@ -0,0 +1,27 @@
...
+            const newChangeEvent = { eventID: nextID, text: changeText } as ChangeEvent;
+            setChangeEvents([...changeEvents, newChangeEvent]);
+            setNextID(nextID + 1);
+        }
+    };
+
+    // https://reactjs.org/docs/hooks-reference.html#functional-updates
+    const deleteChangeEvent = (id: number) => {
+        setChangeEvents((prevChangeEvents) => prevChangeEvents.filter((e) => id !== e.eventID));
+    };
```
jlost, line 24: What would you think about giving each event a `timestamp` and filtering everything >= `n` seconds old, instead of making events self-destruct? Seems like it could reduce some state, although conceivably the list could eventually grow pretty large...

src/assets/src/hooks/useEntityChanges.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ChangeEvent, compareEntities, ComparableEntity } from "../changes";
+
+
+export function useEntityChanges<T extends ComparableEntity>():
+    [ChangeEvent[], (oldEntities: readonly T[], newEntities: readonly T[]) => void, (key: number) => void]
+{
+    const [changeEvents, setChangeEvents] = useState([] as ChangeEvent[]);
+    const [nextID, setNextID] = useState(0);
+
+    const compareAndSetEventMap = (oldEntities: readonly T[], newEntities: readonly T[]): void => {
```
jlost, line 12: Weird question, but I'm not sure what's meant by `Map` in the name of this function.

ssciolla: Ah, nice catch, that is a vestige of a prior implementation that was using an object with IDs as the state variable instead of an array like it currently is. I need to update the function name, thanks!

src/assets/src/hooks/useEntityChanges.ts
ssciolla: Can you elaborate on how you think this would "reduce some state"? I would think that storing less data whenever possible would be preferable, though I'm sure there's a performance cost to all the updates (though I haven't seen this being a problem).

src/assets/src/hooks/useEntityChanges.ts
ssciolla: Should be resolved by https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/258/commits/2beee7b9579a0011cac4e7eb830a0e1be431ce76.

src/assets/src/hooks/useEntityChanges.ts
jlost: I mean it could remove the `useEffect` from `TimedChangeAlert` as well as the need for the deletion callback. Maybe "state" isn't the right word for it, but rather "side effect," although react hooks kinda muddle the difference. At any rate, no worries & this is good; now that I think about it, maybe the alternative of checking the timestamp against `now` on each render cycle has hidden complications.

src/package.json
ssciolla: Sure, I'll get to this.

src/assets/src/components/queueManager.tsx
ssciolla: Hm, yeah, tricky, it seemed like a simpler implementation to just always report them at the top, but they could be hidden from view or result in some page shifting if there are a lot of meetings in play.

PR #262, opened 2020-12-10
Fix notification tests (#255)

### LINKED ISSUE #255: Notification unit test failing
@jlost, I neglected to check the unit tests when I was reviewing your PR #251. I think deleting the `delete` trigger function broke this test. It shouldn't be a big deal to fix or remove it. It may also be worth looking at the other tests to see if they should be tweaked anywhere.

```
======================================================================
ERROR: test_first_meeting_removal_bad_phone_logs_exception (officehours_api.tests.NotificationTestCase)
----------------------------------------------------------------------
  File "/usr/local/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "/usr/src/app/officehours_api/tests.py", line 130, in test_first_meeting_removal_bad_phone_logs_exception
    self.assert_twilio_exception_logged(lambda: m1.delete())
  File "/usr/src/app/officehours_api/tests.py", line 67, in assert_twilio_exception_logged
    self.assertIn('Twilio returned the following information', cm.output[0])
IndexError: list index out of range
```

jlost:
Fixes notifications tests and improves business logic around notifications. The business logic changes aren't noticeable with the current UX, but might accommodate future UX changes, bulk jobs, etc.

Closes #255

### DIFF
--- a/src/officehours_api/models.py
+++ b/src/officehours_api/models.py
@@ -170,13 +170,15 @@ class Meeting(SafeDeleteModel):
             raise BackendException(self.backend_type) from ex
 
     def save(self, *args, **kwargs):
-        if self.backend_type != self._original_backend_type:
-            if self.status == MeetingStatus.STARTED:
+        if self._original_status.value >= MeetingStatus.STARTED.value:
+            if self.backend_type != self._original_backend_type:
                 raise Exception("Can't change backend_type once meeting is started!")
-        if self.assignee != self._original_assignee:
-            if self.status == MeetingStatus.STARTED:
+            if self.assignee != self._original_assignee:
                 raise Exception("Can't change assignee once meeting is started!")
         super().save(*args, **kwargs)
+        self._original_status = self.status
+        self._original_backend_type = self.backend_type
+        self._original_assignee = self.assignee
 
     def delete(self, *args, **kwargs):
         # Trigger m2m "remove" signals for attendees
--- a/src/officehours_api/models.py
+++ b/src/officehours_api/models.py
@@ -139,9 +139,9 @@ class Meeting(SafeDeleteModel):
 
     def __init__(self, *args, **kwargs):
         super(Meeting, self).__init__(*args, **kwargs)
-        self._original_backend_type = self.backend_type
-        self._original_assignee = self.assignee
-        self._original_status = self.status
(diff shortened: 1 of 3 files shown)


src/officehours_api/models.py
```diff
@@ -170,13 +170,15 @@ def start(self, assignee=None):
...
+            if self.backend_type != self._original_backend_type:
                 raise Exception("Can't change backend_type once meeting is started!")
-        if self.assignee != self._original_assignee:
-            if self.status == MeetingStatus.STARTED:
+            if self.assignee != self._original_assignee:
                 raise Exception("Can't change assignee once meeting is started!")
         super().save(*args, **kwargs)
+        self._original_status = self.status
+        self._original_backend_type = self.backend_type
+        self._original_assignee = self.assignee
```
ssciolla, line 181: If we do this, does `original` make sense anymore? Alternatives: `previous`, `saved`?

src/officehours_api/tests.py
```diff
@@ -136,33 +136,28 @@ def test_first_meeting_removal_bad_phone_logs_exception(self, mock_twilio: mock.
...
-        self.create_meeting([self.bar,])
-        self.create_meeting([self.baz,])
-        mock_twilio.reset_mock()
-        m1.delete()
+    def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
+        self.create_meeting([self.foo, self.bar,], True)
         receivers = self.get_receivers(mock_twilio)
-        self.assertFalse('+15555550002' in receivers)
+        self.assertTrue(
+            receivers >=
```
ssciolla, line 161: I'm forgetting now, what's the thinking behind using `>=` over strict equality? I may be not understanding something about set comparison.

ssciolla (approved):
LGTM! Tests ran fine. Just a couple nit-picky comments, take 'em or leave 'em.

src/officehours_api/tests.py
jlost: This is confirming that those values are "in" the set. It could allow there to be more values. Not sure if that even makes a practical difference in this instance, but I chose it because the name of the test implies we're asserting that the attendees were notified (`...notifies_attendees`)

Issue #269, opened 2020-12-10
WaitingTurnAlert needs to be fixed so first-in-line text is shown


ssciolla:
The specific alert text shown to those who are first in line is not showing up because the `WaitingTurnAlert` assumes that first in line is `0`, not `1`. We need to either modify this or how the rest of the attendee-facing queue page handles `line_place`.

Issue #271, opened 2020-12-10
Always show prompt to enable attendee SMS notifications if attendee is in queue


mfldavidson:
Show prompt to enable attendee SMS notifications if attendee is in queue, even if first in line (currently only shows if 2+ in line)

ssciolla:
Hm, okay, I'll take care of this also in PR #273.

PR #260, opened 2020-12-09
Give Badge and Join Info even spacing in MeetingDetails cell on QueueManagerPage

ssciolla:
This PR provides a minor tweak to spacing in the meeting tables on the `QueueManagerPage`. At smaller window widths, the meeting type badge and the "Join Info" button were overlapping, particularly with "BlueJeans" and "In Person". The `md` settings used here fix the spacing, though the spacing seems like a tad too much with the "Zoom" badge. The PR aims to resolve issue #263.

Issue #231, opened 2020-11-19
Attendee view info about place in line


mfldavidson:
# When attendee is 2+ in line (no longer just show this if <=5th--show all the time)
Yellow alert
### Zoom or BlueJeans:
> It's not your turn yet, but the host may be ready for you at any time. Pay attention to this page so you can join the meeting once it is created.

### In-Person:
> It's not your turn yet, but the host may be ready for you at any time. Pay attention to this page--we will tell you when the host is ready. Make sure you are nearby the in-person meeting location.

# When attendee is first in line and Create Meeting/Ready for Attendee has NOT been clicked (i.e. not in progress):
Green alert
### Zoom or BlueJeans:
> You're up next, but the host isn't quite ready for you. Pay attention to this page so you can join the meeting once it is created.

### In-Person:
> You're up next, but the host isn't quite ready for you. Pay attention to this page--we will tell you when the host is ready. Make sure you are nearby the in-person meeting location.

# When Create Meeting/Ready for Attendee HAS been clicked (Meeting moves to Meetings in Progress):
Green alert
### Zoom or BlueJeans:
> The host is ready for you! If you haven't already joined the meeting, follow the directions to join it now.

### In-Person:
> The host is ready for you! If you haven't already, go to the in-person meeting location to meet with the host.

ssciolla:
Because there is no way currently to know whether the host has joined the meeting, this workflow will likely need to be modified a bit. I will talk with @mfldavidson later today.

mfldavidson:
Tested and working in QA except for "when attendee is first in line" part has a bug--new issue #269 for the bug, so moving this to done.

PR #273, opened 2020-12-10
Fix usage of Meeting.line_place, text of alerts on attendee queue page (#260)

ssciolla:
This PR fixes a bug -- introduced during the reorganization of `QueueAttendingJoined` component -- where the `MeetingTurnAlert` component was expecting a zero-based numbering `line_place` but received one with one-based numbering. Rather than modify `MeetingTurnAlert`, I elected to change the other usages of `line_place`, since the backend code generally works with `line_place` as is. Instead, I just added `1` to the `line_place` in the place it is exposed in the UI. The PR also changes the `notificationBlurb` logic so it is always shown if the user is line. The PR aims to resolve issues #269 and #271.

For testing, ensure that attendees see alerts as spelled out in issue #231.

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -207,8 +207,8 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
     const meeting = props.queue.my_meeting!;
     const meetingBackend = getBackendByName(meeting.backend_type, props.backends);
     const isVideoMeeting = VideoBackendNames.includes(meetingBackend.name);
-    const numberInLine = meeting.line_place !== null ? meeting.line_place + 1 : null;
     const inProgress = meeting.status === MeetingStatus.STARTED;
+    const stillInLine = !inProgress && meeting.line_place !== null;
 
     // Alerts and head
     const closedAlert = props.queue.status === "closed"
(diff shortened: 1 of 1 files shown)


src/assets/src/components/queue.tsx
```diff
@@ -207,8 +207,7 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
     const meeting = props.queue.my_meeting!;
     const meetingBackend = getBackendByName(meeting.backend_type, props.backends);
     const isVideoMeeting = VideoBackendNames.includes(meetingBackend.name);
-    const numberInLine = meeting.line_place !== null ? meeting.line_place + 1 : null;
-    const inProgress = meeting.status === MeetingStatus.STARTED;
+    const inProgress = meeting.status === MeetingStatus.STARTED && meeting.line_place === null;
```
ssciolla, line 210: I've been struggling with whether or not we actually need to check what's going on with `line_place` here if we've already checked the `status` (since whether `line_place` is `null` or not is dependent on `status`). The WebSocket updates change both at the same time. I guess it doesn't hurt anything to keep it like this, maybe it protects us a little against changes to backend logic? I'd be interested to hear others' opinions on this.

jlost: My inclination is to try to keep most of the business logic in the backend so it doesn't have to be replicated in the frontend. This lets the frontend neatly reflect business logic changes without additional code changes in most circumstances. So I'd simplify this to just checking the `status`.
The variable naming `inProgress` also hints to me that `line_place` should have nothing to do with it -- or that the name should be adjusted. I could imagine changing `line_place` to be 0 or -1 or something if the meeting is started, in which case this code might stop working the way it should.

jlost (approved):
LGTM!

src/assets/src/components/queue.tsx
ssciolla: Makes sense. I will simplify and test.

jlost (approved):

src/assets/src/components/queue.tsx
ssciolla: Reworked a little bit after chatting with @jlost.

Issue #274, opened 2020-12-11
Only the assigned host can create meetings


mfldavidson:
Once a meeting has been assigned to a host in Meetings In Queue, only the assigned host can see the Ready for Attendee/Create Meeting buttons OR remove the meeting. Restrict via show/hide buttons as well as REST actions > only assigned host.

Once the host is assigned, other hosts (those NOT assigned) see this message in Meeting Actions where the buttons WOULD be:
> Only the assigned host can use meeting actions.

jlost:
REST actions & endpoints should also be restricted to assignees only (if assigned).

PR #277, opened 2020-12-11
Limit Meeting Actions to host assignee, only authorize start/ endpoint for assigned host (#274)

ssciolla:
This PR changes the UI workflow and underlying REST API action for starting meetings so that only the individual assigned to a meeting (`Meeting.assignee`) can start/create a meeting. While the meeting remove button is hidden once a host is assigned, the meeting deletion at the REST API level is still allowed by authenticated hosts and attendees. The PR aims to resolve issue #274.


jlost (approved):
LGTM

PR #278, opened 2020-12-11
Fix backend edge cases (#275)

### LINKED ISSUE #275: Handle host uninstalling Zoom app
If a host uninstalls the Zoom app, the only way I can figure out for them to reinstall it is:
1. Someone with superuser privileges goes in to the user's account in admin and replaces Backend Metadata with:
`{
    "zoom": {
    }
}` and saves
2. Host navigates to a queue that has Zoom as an allowed meeting type and re-authorizes

That's the only thing I can find that triggers them to re-authorize.

If they click Create Meeting on a Zoom meeting after they uninstall it, nothing happens and the generic Zoom error shows:
> Start Meeting: An unexpected error occurred in Zoom. You can check the ITS Status page (https://status.its.umich.edu/) to see if there is a known issue with Zoom, or contact the ITS Service Center (https://its.umich.edu/help) for help.

I'm not sure the correct fix for this...

jlost:
Closes #275

### DIFF
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -119,7 +119,6 @@ class Backend:
                 headers=cls._get_client_auth_headers(),
             )
             if resp.status_code == 401:
-                print("access token 401")
                 # The refresh_token was invalidated somehow.
                 # Maybe the user removed our Zoom app's auth.
                 # Force them to be prompted again.
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -31,6 +31,7 @@ class ZoomMeeting(TypedDict):
     created_at: str
     agenda: str
     start_url: str
+    join_url: str
 
 
 class ZoomUser(TypedDict):
@@ -155,6 +156,9 @@ class Backend:
                 'topic': 'Remote Office Hours Queue Meeting',
                 'start_time': datetime.now().strftime('%Y-%m-%dT%H:%M:%SZ'),
                 'timezone': 'America/Detroit',
+                "settings": {
+                    "join_before_host": False,
+                },
             },
         )
         resp.raise_for_status()
@@ -180,6 +184,7 @@ class Backend:
             'meeting_id': meeting['id'],
             'numeric_meeting_id': meeting['id'],
             'meeting_url': meeting['join_url'],
+            'host_meeting_url': meeting['start_url'],
         })
         return backend_metadata
 
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -118,6 +118,14 @@ class Backend:
                 },
                 headers=cls._get_client_auth_headers(),
             )
+            if resp.status_code == 401:
+                print("access token 401")
+                # The refresh_token was invalidated somehow.
+                # Maybe the user removed our Zoom app's auth.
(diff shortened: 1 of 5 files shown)


src/officehours_api/backends/zoom.py
```diff
@@ -167,9 +174,7 @@ def save_user_meeting(cls, backend_metadata: dict, assignee: User):
         if backend_metadata.get('meeting_id'):
             return backend_metadata
 
-        user_email = backend_metadata['user_email']
-        user = User.objects.get(email=user_email)
-        meeting = cls._create_meeting(user)
+        meeting = cls._create_meeting(assignee)
```
jlost, line 177: It looks like this was causing the meeting to always be created for the first host, which would often not be the assignee, but in testing scenarios, often *would* be.

src/officehours_api/models.py
```diff
@@ -159,8 +159,6 @@ def start(self, assignee=None):
         if not self.assignee:
             raise Exception("Can't start meeting before assignee is set!")
         backend = BACKEND_INSTANCES[self.backend_type]
-        user_email = self.queue.hosts.first().email
-        self.backend_metadata['user_email'] = user_email
```
jlost, line 163: No longer needed, not used anywhere.

src/officehours_api/backends/zoom.py
ssciolla: Might be missing something, but this sounds serious enough that it should be documented in a separate issue. Doesn't need to be a separate PR, in my view, though.

ssciolla: How would you reproduce this bug/case in QA?

jlost: * Create a queue with two hosts
* Create a meeting and assign it to the second host
* Start the meeting
* Correct behavior: underlying Zoom meeting is owned by the second host. Bad behavior: Is owned by first host.

#283

ssciolla: Awesome, thanks. I'll check to make sure it's working.

ssciolla: I might have to wait to test this until it's in QA, unless you've pushed this to `dev`. I don't have access to two umich accounts to use locally.

src/officehours_api/backends/zoom.py
```diff
@@ -118,6 +118,13 @@ def _get_access_token(cls, user: User) -> str:
                 },
                 headers=cls._get_client_auth_headers(),
             )
+            if resp.status_code == 401:
+                # The refresh_token was invalidated somehow.
+                # Maybe the user removed our Zoom app's auth.
+                # Force them to be prompted again.
+                zoom_meta = {}
+                user.profile.backend_metadata['zoom'] = zoom_meta
+                user.profile.save()
```
ssciolla, line 127: Yeah, so this seems to work when I remove the app integration manually. It is triggered when you hit "Create Meeting". There is still an error logged in the backend; that may not be desirable, but I don't expect this to happen very often.

ssciolla (approved):
All these changes make sense to me. I was only able to test the changes for issue #275 locally, but I don't foresee any further issues with #283 with these changes.

src/officehours_api/backends/zoom.py
jlost: Pushed it to dev.

Issue #264, opened 2020-12-10
Having Trouble with Video? Info before the meeting is created


mfldavidson:
Before the meeting has been created, show the Having Trouble with Video container on the attendee page, but show this text:

> Once the meeting is created, instructions for calling in to the meeting on a phone will appear here.

Show the same blurb where Having problems with video shows under Join Info on the host side

Once the meeting is created, show the text that is there now (i.e. the phone number, ID, etc) -- i.e. keep the way it shows up once the meeting is in progress the same as it is now, this change is only related to what shows before the meeting is created

PR #281, opened 2020-12-11
Create dialIn module, DialInContent component, add pre-meeting-creation text (#264)

ssciolla:
The PR aims to resolve issue #264.

### DIFF
--- a/src/assets/src/components/queueManager.tsx
+++ b/src/assets/src/components/queueManager.tsx
@@ -8,17 +8,17 @@ import Dialog from "react-bootstrap-dialog";
 
 import {
     UserDisplay, ErrorDisplay, FormError, checkForbiddenError, LoadingDisplay, DateDisplay,
-    CopyField, showConfirmation, LoginDialog, Breadcrumbs, DateTimeDisplay, BlueJeansDialInMessage,
-    userLoggedOnWarning, DialInMessageProps, ZoomDialInMessage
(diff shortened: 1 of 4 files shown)


src/assets/src/components/queueManager.tsx
```diff
@@ -243,17 +214,19 @@ const MeetingInfoDialog = (props: MeetingInfoDialogProps) => {
...
+        const meetingBackend = getBackendByName(meetingType, props.backends);
+        metadataInfo = VideoBackendNames.includes(meetingType)
+            ? (
+                <>
+                <p>This meeting will be via <strong>{meetingBackend.friendly_name}</strong>.</p>
+                <DialInContent metadata={props.meeting.backend_metadata} backend={meetingBackend} />
+                </>
+            )
+            : <div><p>This meeting will be <strong>In Person</strong>.</p></div>;
+    }
```
jlost, line 229: Not really related to your changes, but when I reread this whole block I wonder if it would make more sense to exit early if `!meeting` (thus assuring to TS that `meeting` is not undefined), rather than continuing along with a bunch of conditional checks and rendering that are ultimately pointless. Again, not for this PR, just something to think about in the future.

jlost (approved):
Pretty straightforward, and I like that you factored the dial-in components into their own module.

src/assets/src/components/queueManager.tsx
ssciolla: Nice idea. I'll keep that in mind for the future, but yeah, that'd require modifying this whole component.

Issue #270, opened 2020-12-10
Filter out meetings in progress when calculating number of people in line for attendees pre-queue-join


mfldavidson:
When an attendee is viewing a queue but has NOT joined yet, this info shows:
> Number of people currently in line: # 

That number should show the number of people in line, and NOT include number of meetings in progress. Currently it includes number of meetings in progress.

PR #285, opened 2020-12-14
Update get_line_length serializer method to count only un-started meetings (#270)

ssciolla:
The PR aims to resolve issue #270.

### DIFF
--- a/src/officehours_api/serializers.py
+++ b/src/officehours_api/serializers.py
@@ -154,7 +154,7 @@ class QueueAttendeeSerializer(serializers.ModelSerializer):
                   'allowed_backends']
 
     def get_line_length(self, obj):
-        return len([obj for obj in obj.meeting_set.all() if obj.status != MeetingStatus.STARTED])
+        return len([meeting for meeting in obj.meeting_set.all() if meeting.status != MeetingStatus.STARTED])
 
     def get_my_meeting(self, obj):
         user = self.context['user']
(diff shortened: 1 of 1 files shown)


src/officehours_api/serializers.py
```diff
@@ -154,7 +154,7 @@ class Meta:
                   'allowed_backends']
 
     def get_line_length(self, obj):
-        return obj.meeting_set.count()
+        return len([obj for obj in obj.meeting_set.all() if obj.status != MeetingStatus.STARTED])
```
jlost, line 157: You might be able to replace the `all` with a `filter`, which would be a bit more performant (offloads work to postgres).

jlost (approved):
LGTM!

src/officehours_api/serializers.py
ssciolla: I wanted to do that, but using a model method (`status`) makes it less straightforward. Looks like you can write a queryset manager that could do it nicely, but I'm not a huge fan of those (yet at least).

https://stackoverflow.com/questions/2276768/django-query-filtering-from-model-method

I also thought about defining a method on `Queue` that would get all the un-started meetings (which we could also use in `Meeting.line_place`), but `Meeting` and `MeetingStatus` are defined after `Queue`, so I'd need to mess with `models.py` more than I really wanted to.

ssciolla:
Thanks, I just modified the loop variable for clarity. Merging now...

Issue #303, opened 2021-01-05
BlueJeans -> Zoom Migration Script


jlost:
Some things that will need to be a part of this:
* Queues: If `queue.allowed_meetings` allows BlueJeans but not Zoom, remove BlueJeans and add Zoom. If it allows BlueJeans and Zoom, remove BlueJeans.
* Meetings: There are a couple directions we could go in...
    * Leave Meetings with `backend_type` of BlueJeans untouched for now and make sure the application code gracefully handles situations where a Meeting has a `backend_type` that isn't globally enabled. This has the advantage of not interrupting any final BlueJeans Meetings that were created just before BlueJeans was disabled (unlikely edge case, especially if we gave ample warning of the change).
    * Delete all Meetings with a `backend_type` of BlueJeans. This has the advantage of letting us remove BlueJeans code immediately. We will have given users ample warning that any BlueJeans meetings would be lost on (date) so it shouldn't come as a surprise.

(As I wrote the above, I think I talked myself into the "deleting the meetings" option, but lmk what you think...)

The script should return a log of all the effected queues and meetings. A dry run option would also be a good idea, just as a pre-run coherence check.

@ssciolla @mfldavidson  Thoughts?

ssciolla:
@jlost, @mfldavidson, what do you think about instead of deleting un-started meetings with BJ, simply changing the meeting type to the default (a concept that is pretty baked into the Django backend already, and will be Zoom in the case of deployment environments)? In combination with some kind of MOTD or Alert, I think this might be the least disruptive way to handle this case. Deleting the meetings in queue would need some more messaging to handle user impact.

jlost:
So if it's unstarted + bluejeans, change to the default, and if it's started + bluejeans, delete it. I like that and agree it could less disruptive, particularly if there are a lot of unstarted meetings of type BlueJeans when we do the migration. The only downside is that exactly what we're migrating becomes a little more conditional and thus could be harder to explain to users, or could generate some confusion if there happens to be started + bluejeans meetings.

ssciolla:
I was actually leaning toward retaining the existing started BJ meetings for this phase of the migration, since with a little finessing they're still functional. The downside is that we may also need a way to delete them later when the backend is removed, which is more code.

ssciolla:
[...]

PR #305, opened 2021-01-07
Introduce ability to phase out backends (#303)

ssciolla:
This PR makes modifications to backends and introduces a new `BackendPhaser` class in order to allow the application to more seamlessly phase out backends that are being retired. The modifications include:

* the addition of an `enabled` class variable to each implemented backend and the passing of all backend metadata via `context_processors.py` (as opposed to just metadata for the enabled ones, as previously);
* the use of the `enabled` property on `MeetingBackend` within the `meetingType` module to hide disabled backends as options in the UI
* a new class called `BackendPhaser`, which collects database operations needed to clear a backend name from queues and meetings
* a couple new model methods to facilitate the operations in `BackendPhaser`
* a Django management command `phase_out_backends` which makes use of `BackendPhaser`
* and other miscellaneous edits and exception handling to minimize the disruption of disabling a backend.

The PR aims to resolve issue #303.

### DIFF
--- a/src/assets/src/components/meetingType.tsx
+++ b/src/assets/src/components/meetingType.tsx
@@ -16,6 +16,8 @@ interface AllowedMeetingBackendsFormProps {
 }
 
 export function AllowedBackendsForm(props: AllowedMeetingBackendsFormProps) {
+    const enabledBackends = props.backends.filter((value) => value.enabled)
+
     const toggleAllowed = (backend_type: string) => {
         const newAllowed = new Set(props.allowed);
         if (newAllowed.has(backend_type)) {
@@ -25,7 +27,7 @@ export function AllowedBackendsForm(props: AllowedMeetingBackendsFormProps) {
         }
         props.onChange(newAllowed);
     }
-    const allowedMeetingTypeEditors = props.backends
+    const allowedMeetingTypeEditors = enabledBackends
         .map((b) =>
             <Form.Group key={b.name} controlId={b.name}>
                 <Form.Check
--- a/src/assets/src/components/meetingType.tsx
+++ b/src/assets/src/components/meetingType.tsx
@@ -16,7 +16,7 @@ interface AllowedMeetingBackendsFormProps {
 }
 
 export function AllowedBackendsForm(props: AllowedMeetingBackendsFormProps) {
-    const enabledBackends = props.backends.filter((value) => value.enabled)
+    const enabledBackends = props.backends.filter((value) => value.enabled);
 
     const toggleAllowed = (backend_type: string) => {
         const newAllowed = new Set(props.allowed);
@@ -54,14 +54,14 @@ interface BackendSelectorProps {
 }
 
 export const BackendSelector: React.FC<BackendSelectorProps> = (props) => {  
-    const options = Array.from(props.allowedBackends)
-        .map(
-            a => (
-                <option key={a} value={a}>
-                    {getBackendByName(a as EnabledBackendName, props.backends).friendly_name}
-                </option>
-            )
+    const enabledAllowedBackends = props.backends.filter(
+        (b) => b.enabled && props.allowedBackends.has(b.name)
     );
+
+    const options = enabledAllowedBackends.map(
+        (a) => <option key={a.name} value={a.name}>{a.friendly_name}</option>
+    );
+
     const handleChange = (event: React.FormEvent<HTMLSelectElement>) => {
         props.onChange(event.currentTarget.value);
     }
--- a/src/officehours/settings.py
+++ b/src/officehours/settings.py
@@ -11,7 +11,6 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 """
 
 import os
-from typing import Literal, Set
 
 import dj_database_url
 
@@ -217,6 +216,7 @@ def skip_auth_callback_requests(record):
         return False
     return True
 
+
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
@@ -334,10 +334,8 @@ BLUEJEANS_DOCS_URL = os.getenv('BLUEJEANS_DOCS_URL', DOCS_BASE_URL + 'blue-jeans
 BLUEJEANS_TELE_NUM = os.getenv('BLUEJEANS_TELE_NUM', '1.312.216.0325')
 BLUEJEANS_INTL_URL = os.getenv('BLUEJEANS_INTL_URL', 'https://www.bluejeans.com/premium-numbers')
 
-IMPLEMENTED_BACKEND = Literal['inperson', 'zoom', 'bluejeans']
-IMPLEMENTED_BACKENDS: Set[IMPLEMENTED_BACKEND] = {'inperson', 'zoom', 'bluejeans'}
-ENABLED_BACKENDS: Set[str] = {'inperson'}
-DEFAULT_BACKEND: IMPLEMENTED_BACKEND = "inperson"
+ENABLED_BACKENDS = {'inperson'}
+DEFAULT_BACKEND = "inperson"
 
 BLUEJEANS_CLIENT_ID = os.getenv('BLUEJEANS_CLIENT_ID', '').strip()
 BLUEJEANS_CLIENT_SECRET = os.getenv('BLUEJEANS_CLIENT_SECRET', '').strip()
--- a/src/officehours/settings.py
+++ b/src/officehours/settings.py
@@ -11,6 +11,7 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 """
 
 import os
+from typing import Literal, Set
 
 import dj_database_url
 
@@ -333,8 +334,10 @@ BLUEJEANS_DOCS_URL = os.getenv('BLUEJEANS_DOCS_URL', DOCS_BASE_URL + 'blue-jeans
 BLUEJEANS_TELE_NUM = os.getenv('BLUEJEANS_TELE_NUM', '1.312.216.0325')
 BLUEJEANS_INTL_URL = os.getenv('BLUEJEANS_INTL_URL', 'https://www.bluejeans.com/premium-numbers')
 
-ENABLED_BACKENDS = {'inperson'}
-DEFAULT_BACKEND = "inperson"
+IMPLEMENTED_BACKEND = Literal['inperson', 'zoom', 'bluejeans']
+IMPLEMENTED_BACKENDS: Set[IMPLEMENTED_BACKEND] = {'inperson', 'zoom', 'bluejeans'}
+ENABLED_BACKENDS: Set[str] = {'inperson'}
+DEFAULT_BACKEND: IMPLEMENTED_BACKEND = "inperson"
 
 BLUEJEANS_CLIENT_ID = os.getenv('BLUEJEANS_CLIENT_ID', '').strip()
 BLUEJEANS_CLIENT_SECRET = os.getenv('BLUEJEANS_CLIENT_SECRET', '').strip()
--- a/src/officehours_api/backends/backend_phaser.py
+++ b/src/officehours_api/backends/backend_phaser.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List, Optional
+from typing import List
 
 from officehours_api.backends.types import IMPLEMENTED_BACKEND_NAME
 from officehours_api.models import Meeting, MeetingStatus, Queue
@@ -18,6 +18,9 @@ class BackendPhaser:
 
     def get_queues_allowing_backend(self) -> List[Queue]:
         return list(Queue.objects.filter(allowed_backends__contains=[self.backend_name]))
+
+    def get_all_meetings_with_backend(self) -> List[Meeting]:
+        return list(Meeting.objects.filter(backend_type=self.backend_name))
 
     def get_meetings_with_backend_through_queues(self, queues: List[Queue]) -> List[Meeting]:
         meetings_with_backend = []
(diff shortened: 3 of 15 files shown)


src/officehours/settings.py
```diff
@@ -333,8 +334,10 @@ def skip_auth_callback_requests(record):
 BLUEJEANS_TELE_NUM = os.getenv('BLUEJEANS_TELE_NUM', '1.312.216.0325')
 BLUEJEANS_INTL_URL = os.getenv('BLUEJEANS_INTL_URL', 'https://www.bluejeans.com/premium-numbers')
 
-ENABLED_BACKENDS = {'inperson'}
-DEFAULT_BACKEND = "inperson"
+IMPLEMENTED_BACKEND = Literal['inperson', 'zoom', 'bluejeans']
```
ssciolla, line 337: I tried to use this type elsewhere via `settings`, but it doesn't seem like `mypy` can do much with because of the `settings` lazy-loading? Not exactly sure how all that works.

src/officehours_api/management/commands/phase_out_backends.py
```diff
@@ -0,0 +1,62 @@
...
+                started_meeting_with_old_backend: List[Meeting] = [
+                    meeting for meeting in meetings_with_old_backend
+                    if meeting.status == MeetingStatus.STARTED
+                ]
+                for started_meeting in started_meeting_with_old_backend:
+                    started_meeting.delete()
+                self.stdout.write(
+                    f'Deleted {len(unstarted_meetings_with_old_backend)} started meetings '
+                    f'with backend_type {old_backend} from queue {queue.id}.'
+                )
```
ssciolla, line 55: Hmm, this would work if you did this at the same time as the rest of the "phase out" work, but it wouldn't do anything if you had previously run this command, because of the traversing through `queue`. I think I need to modularize a bit more, so we could potentially run this code after a "phase out" but before we release without the BlueJeans code.

src/officehours_api/management/commands/phase_out_backends.py
```diff
@@ -0,0 +1,47 @@
...
+        old_backends: Set[settings.IMPLEMENTED_BACKEND] = settings.IMPLEMENTED_BACKENDS - settings.ENABLED_BACKENDS
+        self.stdout.write('Identified one or more old backends: ' + ', '.join(list(old_backends)))
+
+        for old_backend in old_backends:
+            phaser = BackendPhaser(old_backend)
+            phaser.phase_out(
+                options['remove_as_allowed'],
+                options['set_unstarted_to_default'],
+                options['delete_started']
+            )
```
jlost, line 47: I'm a big fan of the pattern of keeping the workhorse module (`BackendPhaser` in this case) separate from the CLI layer.

src/officehours_api/backends/backend_phaser.py
```diff
@@ -0,0 +1,67 @@
...
+        meetings_with_backend = self.get_meetings_with_backend()
+        unstarted_meetings_with_old_backend: List[Meeting] = [
+            meeting for meeting in meetings_with_backend if meeting.status != MeetingStatus.STARTED
+        ]
+        for meeting in unstarted_meetings_with_old_backend:
+            meeting.change_backend_type()
+        logger.info(
+            f'Set the backend_type for {len(unstarted_meetings_with_old_backend)} meeting(s) '
+            f'to the default {settings.DEFAULT_BACKEND}.'
+        )
```
jlost, line 45: This is all good, but an errant thought that popped into my head is that when I write code like this, I generally think logging is better suited to calling contexts (like the view layer or the CLI layer, or even just hoisting it into an aggregate function like `phase_out`), while deeper contexts are better off returning information about the work they did (for example, in this case you could return `unstarted_meetings_with_old_backend` wholesale, then the calling context can get its `len` and compose the log message). This offers better control over logging and makes your functions more functional, which in turn makes them more flexible.

That said, it wouldn't provide any major immediate benefit in this situation (we're unlikely to reuse this code elsewhere), but since it was on my mind I thought I'd offer it up as food for thought.

jlost (commented):
Looks great; I noticed it's still a draft so feel free to keep working if you like, but I made some comments on what you've done so far.

src/officehours_api/models.py
```diff
@@ -96,15 +92,17 @@ class Queue(SafeDeleteModel):
     def hosts_with_phone_numbers(self):
         return get_users_with_emails(self.hosts)
 
+    def remove_allowed_backend(self, old_backend: settings.IMPLEMENTED_BACKEND):
+        new_allowed_backends = list(filter(lambda x: x != old_backend, self.allowed_backends))
+        if len(new_allowed_backends) == 0:
+            new_allowed_backends.append(get_default_backend())
+        self.allowed_backends = new_allowed_backends
+        self.save()
```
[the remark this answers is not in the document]
ssciolla: So you are cool with the change happening here, but you want the `save` to be outside, or do you want the whole operation to not be a method and just be included in the external process? I was somewhat conflicted about whether these two new methods in `models.py` should be on the models at all. Some reading I did (https://medium.com/@jairvercosa/django-model-guideline-d48a96c9b38c) and the extracting of default conditional behavior from the external process made me lean towards this approach.

In Django, bulk operations don't usually require a `save`. I keep wanting to use querysets and bulk operations more for some of this PR, but the reliance on the calculated property `meeting.status` has made it tricky. I see the performance benefit of waiting to `save`, though it's not super relevant with the method uses so far.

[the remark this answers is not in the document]
ssciolla: Oh, I was kinda forgetting about how [`bulk_update`](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#bulk-update) works. Moving `save` out and then doing the bulk update with the modified objects (rather than saving individually) makes a lot of sense.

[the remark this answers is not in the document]
jlost: Fat vs. thin models is an eternal question of MVC; in Django I've been leaning toward fat models (not necessarily the right approach, but it feels like an intuitive way to organize things when you're starting out), while in .NET MVC I purposely use thin models and separate my business logic... So I'm inclined to say thin models might often or usually be better, but I'm not sure it would be worth refactoring this project.

And TBH, I didn't remember exactly how `bulk_update` works either, but usually there's a way to "mass save" (commit your transaction), and when it comes to batch jobs, often a record in the middle of your job will blow the whole thing up. In that case, if you were saving individually instead of in bulk at the very end you'll have a mess to clean up.

src/officehours_api/management/commands/phase_out_backends.py
```diff
@@ -0,0 +1,49 @@
+from typing import Set
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from officehours_api.backends import __all__ as IMPLEMENTED_BACKEND_NAMES
```
ssciolla, line 6: I'd be curious to know if this seems like a good idea or not. I realized that defining the list of implemented backends in `settings.py` actually didn't make that much sense, and this is probably a more straightforward approach that promotes keeping things in sync.

src/officehours_api/models.py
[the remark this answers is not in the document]
ssciolla: Yeah, it's tricky but interesting to think about; nice to hear your perspective. I think (?) I buy the idea in that article that model methods shouldn't be too dependent on external processes (like the email sending example) and just focus on transforming and modifying data associated with that model. I guess construing "dependency" broadly, you could say referencing `settings` indirectly is a dependency, but that was already something we were doing in `models.py`.

src/officehours/settings.py
ssciolla: Yeah, I'm using `mypy` and I tried setting up `django-stubs`, but it wasn't immediately clear to me whether it was working how I wanted it to/if I had it configured correctly. I decided to move this specific type out of `settings` to somewhat circumvent this, but also because details about what is implemented was (or probably should be) elsewhere, i.e. in `backends`.

src/officehours_api/models.py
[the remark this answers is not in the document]
ssciolla: K, I think this commit https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/305/commits/9a001989ad83413f282c2e9ba917e16cbae64884 is what you had in mind. There is only a bulk deletion method for querysets, so I'm not sure I can make that work here for the remaining operation I didn't modify.

src/officehours_api/models.py
```diff
@@ -137,6 +135,10 @@ class Meeting(SafeDeleteModel):
     def attendees_with_phone_numbers(self):
         return get_users_with_emails(self.attendees)
 
+    def change_backend_type(self, new_backend: Optional[settings.IMPLEMENTED_BACKEND] = None):
+        self.backend_type = new_backend if new_backend else get_default_backend()
+        self.save()
```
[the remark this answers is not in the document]
ssciolla: See https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/305/commits/9a001989ad83413f282c2e9ba917e16cbae64884

src/officehours_api/backends/backend_phaser.py
ssciolla: So in summary, you're saying return the data needed for logging and then log in `phase_out` or at the CLI level?

In terms of knowing exactly what the program is doing, I see some benefit to having some logging at the point of doing the work. You can have log messages tell you where you it's being called from (the method), so people can quickly find the relevant code if there's a problem. All of these could get downgraded to `debug`.

I think I follow your reasoning regarding having logging at a higher level. However, do you see any downside to separating the acting logic and the log creation? I wonder if returning fairly different data from all these functions to a higher-order method/function may make that code more difficult to parse for humans. I have seen people pass strings back and either log that or concatenate it with other strings to create a report at the end; I've never really loved that either.

I appreciate you raising this, it's good to think about these patterns more.

jlost: > So in summary, you're saying return the data needed for logging and then log in phase_out or at the CLI level?

Maybe; I realize this isn't a clear example of where it would be helpful. If you picture a method that looks at a data structure and conditionally changes some of its elements, you could imagine a lot of situations where calling contexts might want to do something other than just log the ones that were changed, or log them differently. So returning the changed records could potentially be more powerful, or returning a new data structure leaving the old one intact (immutability)... 

This idea also kinda relates to a concept in FP where you want to avoid "side effects" (like logging, network calls, etc.) but those "side effects" actually compose all of the useful stuff your program does (writing stuff to the screen, writing an http response...), so what you really do is hoist those parts up to the top layer of your app, making the lower layers more data driven.

> I see some benefit to having some logging at the point of doing the work

Agree, although if your functions are small and simple and pure enough, unit testing or even just REPL testing becomes a lot easier.

> I wonder if returning fairly different data from all these functions to a higher-order method/function may make that code more difficult to parse for humans

Yeah, this is particularly a problem in Python where typings aren't very good yet (being able to infer method types in your IDE helps immensely with following the shape of the data), so I agree that it's not a silver bullet, just something to think about.

ssciolla: Okay, does something like this seem better to you? https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/305/commits/fa24545b6ce21857a9163e2553dffe67aa0fba3f

src/assets/src/components/meetingType.tsx
```diff
@@ -52,14 +54,14 @@ interface BackendSelectorProps {
...
-                </option>
-            )
+    const enabledAllowedBackends = props.backends.filter(
+        (b) => b.enabled && props.allowedBackends.has(b.name)
+    );
+
+    const options = enabledAllowedBackends.map(
+        (a) => <option key={a.name} value={a.name}>{a.friendly_name}</option>
     );
+
```
ssciolla, line 64: I added this change recently. Prior to the `BackendPhaser.phase_out` being called, you could still see BlueJeans as an option in the dropdown, though the API still threw errors appropriately. I suppose we could debate whether all these changes to `meetingType.tsx` are necessary if we plan to use the DB migration code immediately.

ssciolla:
K, I think this is ready for a full review. Thanks for the feedback so far.

src/officehours_api/management/commands/phase_out_backends.py
jlost: Yeah, this makes the most sense.

src/officehours_api/backends/backend_phaser.py
jlost: Maybe ever so slightly; you didn't need to change anything, but LGTM!

jlost (approved):
Looks good & tested in dev!

jlost (approved):

src/officehours_api/backends/inperson.py
```diff
@@ -1,11 +1,13 @@
...
 
-from officehours_api.backends.backend_dict import BackendDict
+from officehours_api.backends.types import BackendDict, IMPLEMENTED_BACKEND_NAME
 
 
 class Backend:
-    name = 'inperson'
+    name: IMPLEMENTED_BACKEND_NAME = 'inperson'
     friendly_name = "In Person"
+    enabled = True
```
jlost, line 9: Nitpick, but it might make a bit more sense to handle this the same way as the other backends, since it would be more consistent and `inperson` could potentially be disabled by providing a different `settings.py`.

jlost (approved):
LGTM!

src/officehours_api/backends/inperson.py
ssciolla: Sure, no problem, I can do that.

ssciolla: Done. See https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/305/commits/6b9802a2dc5156ed5c41299439e05daf8d78866a

ssciolla:
Thanks for the reviews, @jlost. I'm going to merge now, so we can get this into QA soon.

Issue #304, opened 2021-01-06
Change homepage text to not mention BlueJeans


jlost:
![image](https://user-images.githubusercontent.com/11338926/103808984-d0faaa80-5026-11eb-8951-230463234831.png)

This text mentions BlueJeans but not Zoom. Ideally, it should probably be sourced from the `friendly_name` of `ALLOWED_BACKENDS`.

ssciolla:
@mfldavidson, I think you alluded to this in a recent meeting; this should probably not reference a specific VC technology at all. How about something like this?

"Join or host a queue for virtual or in-person office hours!"

ssciolla:
You could add a line about Zoom if that's important. Something like this:

"The tool currently supports {backend.friendly_name} videoconferencing."

mfldavidson:
How about "Join or host a queue for office hours." No mention of types?

mfldavidson:
You could also list out the types but I balk at hard coding any specific type of meeting (like in person)

ssciolla:
Yeah, we wouldn't need to hardcode anything if we listed the types, but I like your suggestion of not including any, as long as that's okay with you.

ssciolla:
I guess my thinking was that the pre-login landing page should give some more specific idea of what's being offered.

mfldavidson:
Tested in QA on 2/15 using Chrome Desktop, testing passed
![image](https://user-images.githubusercontent.com/31591389/107997518-ec0af380-6fb0-11eb-85b0-9714fa24065b.png)

PR #307, opened 2021-01-20
Remove mention of BlueJeans from login page

ssciolla:
This PR changes the text presented to the user before they have logged in, as requested by @mfldavidson, so that it does not mention backend or meeting types implemented. The PR aims to resolve issue #304.


jlost (approved):

mfldavidson (approved):
Looks good! Maybe someday in the future folks will want more clarity at this point but for now I think this is what we want and an easy resolution to the issue. Thanks for talking through it.

Issue #239, opened 2020-12-01
Co-locate meeting-related errors with Meetings in Queue/Add Attendee Form


ssciolla:
While we've improved error location and validation feedback elsewhere in the application, the `QueueManagerPage` still puts a number of meeting-related errors at the top of the page ('Add Meeting', 'Remove Meeting', 'Assignee', 'Start Meeting'). We could probably create a separate `ErrorDisplay` that lists these errors under "Meetings in Queue" next to the `AddAttendeeForm`.

mfldavidson:
I would prefer we not do this in 1.4--I do not think we will make it. I am pulling it and hopefully we can do it soon but not in this release.

mfldavidson:
Tested in QA on 2/15 using Chrome Desktop:

- [ ] As a host on the queue management page, try to add a user who is already in a meeting = [user] is already in a meeting message shows just above the entry field, very visible! ✅ 
- [ ] As a host on the queue management page, try to add an invalid "uniqname" between 3 and 8 characters = invalid user message shows just above the entry field, very visible! ✅ 
- [ ] As a host on the queue settings page, try to add an invalid "uniqname" between 3 and 8 characters = invalid user message shows just above the entry field, very visible! ✅ 

The other errors aren't as easy to replicate as they'll only show if something goes wrong (fail to remove meeting, fail to start meeting, etc) but based on above testing, they should display in correct location as well

PR #308, opened 2021-01-20
Co-locate Add Meeting errors with AddAttendeeForm (#239)

ssciolla:
The separates out the "Add Meeting" error source from the other errors and passes it down the component tree to `QueueManager` so it can report errors usually triggered by the `AddAttendeeForm` (specifically those thrown when a user not in the system is entered, or when the user is already in a meeting) next to that form, rather than at the top of the page. The PR aims to resolve issue #239.

Note: Other errors could be reported on different parts of the page (e.g. "Start Meeting" or "Remove Meeting"), but too much customization could complicate the UI and code and not offer that much benefit. Of all these, "Add Meeting" error handling seems to me to offer the greatest benefit to the user experience.

### DIFF
--- a/src/assets/src/components/queueManager.tsx
+++ b/src/assets/src/components/queueManager.tsx
@@ -104,6 +104,7 @@ interface QueueManagerProps {
     backends: MeetingBackend[];
     defaultBackend: string;
     onAddMeeting: (uniqname: string, backend: string) => void;
+    addMeetingError: FormError | undefined;
     onRemoveMeeting: (m: Meeting) => void;
     onSetStatus: (open: boolean) => void;
     onShowMeetingInfo: (m: Meeting) => void;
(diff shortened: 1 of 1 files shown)


jlost (approved):
LGTM!

src/assets/src/components/queueManager.tsx
```diff
@@ -323,19 +325,18 @@ export function QueueManagerPage(props: PageProps<QueueManagerPageParams>) {
...
     const isChanging = removeMeetingLoading || addMeetingLoading || setStatusLoading || changeAssigneeLoading || startMeetingLoading;
-    const errorSources = [
+    const globalErrorSources = [
         {source: 'Access Denied', error: authError},
         {source: 'Queue Connection', error: queueWebSocketError},
         {source: 'User Connection', error: userWebSocketError},
         {source: 'Remove Meeting', error: removeMeetingError},
-        {source: 'Add Meeting', error: addMeetingError},
         {source: 'Queue Status', error: setStatusError},
-        {source: 'Assignee', error: changeAssigneeError},
```
ssciolla, line 333: Looks like I accidentally deleted this when I was playing around with it. Going to add it back.

ssciolla: Done (https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/308/commits/6d1a3eb471772492095dd404cb5ec8a054ec8b88)

Issue #301, opened 2020-12-16
Info for host on host key


mfldavidson:
If the current user is the host of the meeting, when they click Join Info, instead of showing "You do not need a host key or participant ID" they should see "You will need to enter a host key. Find your host key at the bottom of [your U-M Zoom profile](https://umich.zoom.us/profile). DO NOT share your host key with anyone! You do not need a participant ID."

ssciolla:
Okay, wait, to be clear, are we changing things such that a host key is required? My impression from the conversation today was that this would only be needed if you were calling in. (I just want to make sure what we're trying to accomplish, how to organize the related content.)

mfldavidson:
No, we are not making any changes. The host key is already required if the host is only calling in on the phone, and not a computer. Maybe we should make that clear, here's adjusted language:

> You will need to enter a host key to start the meeting if you only join by calling in on a phone. Find your host key at the bottom of [your U-M Zoom profile](https://umich.zoom.us/profile). DO NOT share your host key with anyone! You do not need a participant ID.

ssciolla:
Thanks for the clarification, @mfldavidson. I modified the language/formatting a little bit to accommodate the conditional logic. I also felt like the "if you only join by calling in on a phone" might be a bit heavy-handed given that this is already in a dial-in blurb. Can you take a look at the images below and critique my changes?
<img width="744" alt="Screen Shot 2021-01-25 at 12 37 43 PM" src="https://user-images.githubusercontent.com/35741256/105744335-f5052980-5f0a-11eb-9bec-e6e5a11b6fdd.png">
<img width="511" alt="Screen Shot 2021-01-25 at 12 38 24 PM" src="https://user-images.githubusercontent.com/35741256/105744337-f6365680-5f0a-11eb-9484-453c0908b880.png">

mfldavidson:
Participant view looks good and I love your edits. Great work @ssciolla , let's open a PR.

mfldavidson:
Tested in QA on 2/15 using Chrome Desktop:
- [x] confirmed that the blurb only shows up for hosts and not attendees
- [x] confirmed that the blurb matches the instructions people hear when they call in
- [x] confirmed that clicking the link to your Zoom profile works properly if currently logged into Zoom in the browser, or if not logged in; if user is logged in to a DIFFERENT Zoom account, it will show the account they are logged into and not their U-M account, but nothing we can do about that

PR #309, opened 2021-01-25
Give host info about host key for dial in (#301)

ssciolla:
This PR adds a nullable `profile_url` to the backend configuration workflow and leverages it in the `dialIn` module to add information for Zoom users about how to find their host key to use when calling in by phone. The PR aims to resolve issue #301.

### DIFF
--- a/src/officehours_api/backends/bluejeans.py
+++ b/src/officehours_api/backends/bluejeans.py
@@ -134,6 +134,7 @@ class Backend:
     enabled: bool = name in settings.ENABLED_BACKENDS
 
     docs_url: str = settings.BLUEJEANS_DOCS_URL
+    profile_url = None
     telephone_num: str = settings.BLUEJEANS_TELE_NUM
     intl_telephone_url: str = settings.BLUEJEANS_INTL_URL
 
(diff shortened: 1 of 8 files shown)


src/officehours_api/backends/bluejeans.py
```diff
@@ -193,6 +194,7 @@ def get_public_data(cls) -> BackendDict:
             'friendly_name': cls.friendly_name,
             'enabled': cls.enabled,
             'docs_url': cls.docs_url,
+            'profile_url': cls.profile_url,
             'telephone_num': cls.telephone_num,
             'intl_telephone_url': cls.intl_telephone_url
         }
```
ssciolla, line 200: @jlost, this method is identical across the three backends, which begs the question whether it should be inherited from a parent class. I introduced this method, but I'm fairly certain you've had occasion in the past to consider something like this (and thus that you have some opinion about this). Any thoughts you'd like to share?

jlost: My thoughts on this aren't very nuanced; I think a base class makes sense to share methods like this and also show which method signatures need to exist in all `Backend`s.

jlost (approved):
Looks good.

src/officehours_api/backends/bluejeans.py
ssciolla: Cool, this isn't directly related to the issue being addressed here, so I've created a separate one: https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/311

Issue #311, opened 2021-01-27
Create parent class for Backends


ssciolla:
@jlost, this method is identical across the three backends, which begs the question whether it should be inherited from a parent class. I introduced this method, but I'm fairly certain you've had occasion in the past to consider something like this (and thus that you have some opinion about this). Any thoughts you'd like to share?

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/309#discussion_r564849401_

I'm thinking the parent class should at least have the public class variables and basic implementations of `save_user_meeting` and `get_public_data`.

ssciolla:
This can be marked as QAed/Done after a regression test.

PR #312, opened 2021-01-27
Implement BackendBase abstract base class (#311)

ssciolla:


src/officehours_api/backends/backend_base.py
```diff
@@ -0,0 +1,38 @@
...
+        }
+    
+    @classmethod
+    @abstractclassmethod
+    def is_authorized(cls, user: User) -> bool:
+        pass
+
+    @abstractmethod
+    def save_user_meeting(self, backend_metadata: dict, assignee: User):
+        pass
```
ssciolla, line 38: @jlost, this isn't the highest priority issue, but I was curious how difficult it would be. When you get a chance, let me know what you think. I think there might be a way to enforce the class variables, but get_public_data would fail anyway if not all those were defined. We could remove the defaults, but I read somewhere those were recommended when possible.

Maybe relevant reading: https://stackoverflow.com/questions/22046369/enforcing-class-variables-in-a-subclass

ssciolla:
@jlost, I think it might be kinda nice to get this in soon. Seems pretty low-risk, and will give people (us or future developers) direction if they need to add another backend.

jlost (approved):
Looks great!

Issue #261, opened 2020-12-10
Cancel my meeting link > popup changes


mfldavidson:
This is not urgent but should probably be addressed eventually.

Now that the "Leave line" link changes to "Cancel my meeting" once the meeting is in progress, we probably need to adjust the language in the popup. The popup still shows:
![image](https://user-images.githubusercontent.com/31591389/101806784-5f404700-3ae2-11eb-8035-e46926703536.png)

Better language would be something like:

> #### Cancel Meeting?
> Are you sure you want to cancel your meeting? By cancelling your meeting, you will no longer meet with the queue host(s), and will have to rejoin at the end of the line if you change your mind.

mfldavidson:
Tested and working in QA 2/18/21

PR #316, opened 2021-02-15
Adjust leave queue dialog text based on meeting status (#261)

ssciolla:
This PR modifies the "Leave Queue" dialog to respond to the meeting status, as requested by @mfldavidson. I modified some of the text so the conditionals would be a bit simpler. The PR aims to resolve issue #261.

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -6,8 +6,8 @@ import { Alert, Button, Card, Col, Modal, Row } from "react-bootstrap";
 import Dialog from "react-bootstrap-dialog";
 
 import {
-    BluejeansMetadata, EnabledBackendName, MeetingBackend, MeetingStatus, MyUser, QueueAttendee,
-    User, VideoBackendNames, ZoomMetadata
+    BluejeansMetadata, EnabledBackendName, Meeting, MeetingBackend, MeetingStatus, MyUser,
+    QueueAttendee, User, VideoBackendNames, ZoomMetadata
 } from "../models";
 import {
     checkForbiddenError, Breadcrumbs, DateTimeDisplay, DisabledMessage, EditToggleField, ErrorDisplay,
@@ -59,7 +59,7 @@ interface QueueAttendingProps {
     joinedQueue?: QueueAttendee | null;
     disabled: boolean;
     onJoinQueue: (backend: string) => void;
-    onLeaveQueue: () => void;
+    onLeaveQueue: (myMeeting: Meeting) => void;
     onLeaveAndJoinQueue: (backend: string) => void;
     onChangeAgenda: (agenda: string) => void;
     onShowDialog: () => void;
@@ -256,7 +256,7 @@ function QueueAttendingJoined(props: QueueAttendingProps) {
         <Button
             variant='link'
             type='button'
-            onClick={() => props.onLeaveQueue()}
+            onClick={() => props.onLeaveQueue(meeting)}
             disabled={props.disabled}
             aria-label={leaveButtonText}
         >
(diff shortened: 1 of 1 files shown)


src/assets/src/components/queue.tsx
```diff
@@ -446,14 +446,27 @@ export function QueuePage(props: PageProps<QueuePageParams>) {
         await api.removeMeeting(queue!.my_meeting!.id);
     }
     const [doLeaveQueue, leaveQueueLoading, leaveQueueError] = usePromise(leaveQueue);
-    const confirmLeaveQueue = (queueStatus: 'open' | 'closed') => {
-        const loseMessage = 'By leaving the queue, you will lose your place in line'
-        const description = 'Are you sure you want to leave the queue? ' + (
-            queueStatus === 'open'
-                ? loseMessage + '.'
-                : `The queue is currently closed. ${loseMessage} and will not be able to rejoin until the queue is reopened.`
+    const confirmLeaveQueue = (queueStatus: 'open' | 'closed', meetingStatus: MeetingStatus | undefined) => {
```
ssciolla, line 449: It might be nice to move this out of `queue.tsx`, but I wasn't really sure where else to put it.

src/assets/src/components/queue.tsx
```diff
@@ -446,14 +446,27 @@ export function QueuePage(props: PageProps<QueuePageParams>) {
...
+                action: 'leave the queue',
+                gerund: 'leaving the queue',
+                consequences: 'you will lose your place in line'
+            };
+
+        const description = (
+            `Are you sure you want to ${dialogParts.action}? ` +
+            `By ${dialogParts.gerund}, ${dialogParts.consequences}. ` +
+            'If you change your mind, you will have to re-join at the end of the line.' +
+            (queueStatus === 'closed' ? ' The queue is currently closed. You will not be able to re-join until the queue is re-opened.' : '')
```
mfldavidson, line 467: I like the way you adjusted this to work for all the possible scenarios. The only thing I think we should change is that the flow of saying "If you change your mind, you'll have to rejoin. The queue is closed. You can't rejoin." feels a little weird to me in that you're saying, "you have to do this thing but actually you can't and it's slightly weird that I'm not addressing that in the first statement." I think that simply adding "Note:" in front of "The queue is currently closed" might help that feel less weird. We might even make that part Italic. So something like this:

```suggestion
            (queueStatus === 'closed' ? ' <em>Note: The queue is currently closed. You will not be able to re-join until the queue is re-opened.</em>' : '')
```

mfldavidson (changes requested):
Approved with requested change. Nice work figuring out how to make this work for so many complicated scenarios, @ssciolla 🎉

jlost (approved):
Looks good, I like the factoring.

src/assets/src/components/queue.tsx
ssciolla: @mfldavidson, are the italics important to you? I'm not absolutely sure if that's possible here. The package we're using might not support HTML or other mark-up within the dialog body. https://www.npmjs.com/package/react-bootstrap-dialog#showoptions

ssciolla: What you suggested just ends up being literal tags, since it's part of a string.

mfldavidson: I'm fine either way, it was just an example.

mfldavidson (approved):

src/assets/src/components/queue.tsx
ssciolla: Might be possible to do this, but I'm going to put that off for now, since it'd require some more extensive changes.

ssciolla:
@jlost, sorry, upon reviewing my work, I think I found a more appropriate way to invoke `confirmLeaveQueue` to ensure `meeting` is defined. Can you take a look at these changes? https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/316/commits/cc9fb30188b9b1f65a2e7e7264ec2ce0a308bfef

jlost (approved):

Issue #317, opened 2021-02-15
When migrating, replace disabled backend with the default backend, ignoring other enabled backends


ssciolla:
During discussion of the migration logic, we determined that it in the case of `['inperson', 'bluejeans']` in `Queue.allowed_backends`, it would be preferable to replace `bluejeans` with `zoom` (or some default), rather than removing the disabled backend (i.e. `['inperson']`), which is what `BackendPhaser` is doing now.

I proposed the following on Slack:

"We could replace `remove_allowed_backend` with `replace_allowed_backend_with_default` (and the implied logic, which is similar to what we have, just always add `DEFAULT_BACKEND` if it’s not present) and I think that fixes things.

Meetings would then use the default backend,  `Meeting.change_backend_type` prefers `DEFAULT_BACKEND` if it’s available."

mfldavidson:
Tested and working in QA 2/18/21

PR #318, opened 2021-02-16
Replace disabled backend with default in Queue.allowed_backends (#317)

ssciolla:
This PR modifies the `BackendPhaser` migration logic and the `Queue` model method referenced therein to better handle the case where a `Queue` has `['inperson', 'bluejeans']` for `allowed_backends`. The model method `replace_allowed_backend_with_default` was modified so that it 1) removes the disabled backend specified, 2) checks for the default backend in the `allowed_backends` and 3) adds it to `allowed_backends` if not present. This results in `['inperson', 'zoom']`, instead of `['inperson']` as with the old logic. Unstarted meetings will subsequently be changed to the default backend, because `Meeting.change_backend_type` prefers that when it's in `meeting.queue.allowed_backends`. The PR aims to resolve issue #317. 

`Queue.allowed_backends` cases that need to be handled (expected outcome included):
* `['inperson', 'bluejeans']` -> `['inperson', 'zoom']`
* `['bluejeans']` -> `['zoom']`
* `['inperson', 'zoom', 'bluejeans']` -> `['inperson', 'zoom']`
* `['bluejeans', 'zoom']` -> `['zoom']`

### DIFF
--- a/src/officehours_api/backends/backend_phaser.py
+++ b/src/officehours_api/backends/backend_phaser.py
@@ -28,9 +28,9 @@ class BackendPhaser:
             meetings_with_backend += list(queue.meeting_set.filter(backend_type=self.backend_name))
         return meetings_with_backend
 
-    def remove_backend_from_queue_allowed_backends(self, queues_allowing_backend: List[Queue]) -> List[Queue]:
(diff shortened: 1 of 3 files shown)


jlost (approved):
Looks good!

Issue #315, opened 2021-02-15
Make profile link in host key blurb target=_blank


mfldavidson:
In this section for hosts calling in to meetings:

> When calling in by phone, you will need to enter a host key to start the meeting. Find your host key at the bottom of your Zoom profile. DO NOT share your host key with anyone!

We should make "your Zoom profile" link target=_blank so that they aren't taken away from the page in ROHQ they're currently working on

mfldavidson:
Tested and working in QA 2/18/21

PR #319, opened 2021-02-17
Add target="_blank" to link tag for Zoom profile (#315)

ssciolla:
This PR makes a requested change to the `ZoomDialInMessage` so that the Zoom profile link opens in a new tab rather than sending users away from ROHQ. The PR aims to resolve issue #315.


mfldavidson:
FYI if you use the word "issue" I believe it prevents it from linking the PR to the issue. I think you should just say "resolve #315". I linked it already though

mfldavidson (approved):

ssciolla:
> FYI if you use the word "issue" I believe it prevents it from linking the PR to the issue. I think you should just say "resolve #315". I linked it already though

Thanks, @mfldavidson. good to be reminded. I'm kinda being stubborn here 'cause I prefer the language I used, but I know it adds an extra step for us. I'll do it differently going forward.

PR #327, opened 2021-04-22
#204 Added hover functionality for nav-link buttons and dropdown menu buttons

### LINKED ISSUE #204: Hover indicator for Home and Resources & Support
When hovering over your uniqname in the upper right, the color changes to show that you're hovering over it. This does not occur for Home nor for Resources & Support. Should it?

Hovering over my uniqname:
image

Hovering over Home or Resources & Support:
image

vjcao:
This PR closes #204. When hovering over the nav-link buttons, they will change their colors to a lighter shade of blue (#004881, same as the original dropdown color). When hovering over the dropdown menu buttons, the white buttons will turn to #C0C0C0.


mfldavidson:
Closes #204

src/officehours_ui/static/css/main.css
```diff
@@ -687,3 +687,16 @@ div[role=main] {
...
+	border-radius: 3px;
+}
+
+a.dropdown-item:hover {
+	background-color: #C0C0C0 !important;
+}
+
+input.dropdown-item:hover {
+	background-color: #C0C0C0 !important;
+}
```
ssciolla, line 702: @mfldavidson or @teshsopher, we may need some input here. @vjcao also added a gray hover background, which was out of scope for the issue, but a logical addition. Do we like this? Do we need to tweak the color for adequate contrast/consistency?
Screen Shot 2021-05-05 at 12 26 13 PM

ssciolla, line 702: @vjcao, are both the `a` and `input` blocks necessary for this? I tried commenting out the `input` block, nothing seems to have changed?

ssciolla (commented):
Solution looks good when I tested it, just a couple questions about the choice of color and whether all the blocks are necessary.

src/officehours_ui/static/css/main.css
vjcao: Hi @ssciolla. Both of the blocks are necessary for the dropdown menu. The `a` block is for the "Manage Queues" and "Preferences" buttons while the `input` block is for the "Logout" button. Sometimes changes to static files won't be reflected on local host so I sometimes have to use this command to refresh the static files: `python manage.py collectstatic --noinput --clear`.

ssciolla: Ah, you are correct. Logout is not highlighting on hover. I'll resolve this comment.

teshsopher: @ssciolla I think it looks good, and the color contrast is great too! 

My only question is: does this menu show up when the user clicks the "hamburger" menu in the top right? If it is, then I would suggest having the hamburger menu appear as an "X" or some other "close menu" icon while the menu is open so that users can more easily see that they can close the menu.

ssciolla: Thanks for the input on highlight colors and hamburger-menu suggestion, @teshsopher. Note that this picture is the mobile (small viewport) appearance; there is no hamburger menu with a large viewport. While related to these changes, I feel this is a separate issue that would require additional work and research. Would you mind taking a moment to document this as a new issue? If you're pretty busy, I can, just let me know.

teshsopher: I can do that!

ssciolla (approved):
Changes look good when I tested them, did not see any disruptions to other drop-down menus or inputs. I wonder if it's possible to make use of existing classes with some of this, but CSS could use a re-org anyway and we've talked about moving the navigation bar into the SPA. I'm good with this for now!

src/officehours_ui/static/css/main.css
ssciolla: Thank you!

Issue #234, opened 2020-11-19
Field for in-person meeting location on queue if queue has in-person meeting type option


mfldavidson:
If In-Person is selected as a meeting type option, an additional field should become available just below Meeting Types in Add Queue and Settings.

> #### In-Person Meeting Location
> Attendees who select to meet in-person will be instructed to meet at this location. Enter all information an attendee would need to know, such as a street address, building name, and/or room number.

The field should be a text field with a character limit of 100 and the same validation used in the queue name field.

In the attendee view, BEFORE joining the line, the location should show as an additional bullet below "You are not in the meeting queue yet" but only if In Person is an allowed meeting type:
> - In-person meetings will take place at: [location]

AFTER joining the line, if the meeting type is In Person, the location should show below "Meeting Via" above "Agenda":
> Meet at: [location]

vjcao:
May I work on this issue as well? Thank you.

mfldavidson:
Please do! Thank you @vjcao

vikaschanduri:
Hi, would be great to work on this issue as well!

ssciolla:
> Hi, would be great to work on this issue as well!

@vikaschanduri, I think we're good on this issue now, @vjcao has it pretty much done.

mfldavidson:
![image](https://user-images.githubusercontent.com/31591389/117322345-ed371880-ae5b-11eb-9bf7-6983c7c407fc.png)

Tested and passed:
- [x] Adding this feature to a queue that already allowed in-person meetings didn't break anything since it isn't required
- [x] Checking in-person as a meeting type in queue settings shows the field, unchecking it hides it
- [ ] Checking in-person as a meeting type in add queue shows the field, unchecking hides it
- [ ] Prior to joining the queue, attendees see "The host(s) have not specified an in-person meeting location." if the field is blank AND in-person is enabled
- [ ] Prior to joining the queue, attendees do NOT see "The host(s) have not specified an in-person meeting location." if in-person is not enabled
- [ ] Prior to joining the queue, attendees see the in-person meeting location if the field has any characters other than a space in it
- [ ] Prior to joining the queue, attendees see "The host(s) have not specified an in-person meeting location." if the field only has blank spaces in it AND in-person is enabled
- [ ] After joining the queue for a Zoom meeting, in-person meeting location does not show at all
[...]

Issue #219, opened 2020-11-05
Align Manage Queues language


mfldavidson:
On homepage, link is "Manage Your Own Queues", but in dropdown, it is "Manage Queues." Change link on homepage to "Manage Queues" to align language and make it clear they are the same page.

HemanginiShah:
5/31/2023 - QA test completed successfully.  Now both places the same text is visible. On homepage, link is "Manage Queues", and also in dropdown, it is "Manage Queues."

PR #326, opened 2021-04-22
Issue #234 Adding Text Field for In-Person Meeting Location

vjcao:
This PR addresses issue #234. The text input uses the same validation as the queue name field and only appears after the in-person checkbox is checked. This change only works on the front end though. The backend doesn't save the input from the field for some reason. Any help would be much appreciated.

PR #328, opened 2021-04-25
Issue #219 Align Manage Queues Language and Issue #234 "Meeting Location" Textfield 

vikaschanduri:
This PR addresses issue #219 and issue #234. 

#219 The Manage Queues link on the homepage now uses the same language as the dropdown menu link. (only 1 line changed)

#234 There is now a text field that accepts "Meeting Location" information (200 character limit) in the "Add Queue" and "Queue Settings/Manage Queue" pages. Performed a database migration to add the "meeting_location" field to the schema so the meeting location information persists across pages correctly. Rest of the behavior follows specifications from issue description exactly. 

Other: added queue "Description" and "Meeting Location" text to "Manage" queues page. Goal is to make accessing that information easier for the Host since it is no longer required to go into the queue "Settings" page to see it.

This pull request is part of a final project for a UMich CS class. Our team realized there was already work being done on this issue as we were finishing up our implementation, so we understand if this PR cannot be accepted. Regardless, any feedback would be greatly appreciated!


src/assets/src/components/queueEditors.tsx
```diff
@@ -74,6 +79,27 @@ export function GeneralEditor(props: GeneralEditorProps) {
                 onChange={props.onChangeAllowed}
                 disabled={props.disabled}
             />
+            {/* our code */}
+            {
+                props.allowedMeetingTypes.has("inperson") &&
+                <div>
```
ssciolla, line 85: You can also use `<>` and `</>` here instaed of an enclosing `div`. I recently saw a tip in an article suggesting this was preferred, see 4 in https://betterprogramming.pub/21-best-practices-for-a-clean-react-project-df788a682fb. Certainly it would keep the HTML structure more consistent, since none of the blocks associated with a field are grouped in `div`s.

src/assets/src/components/home.tsx
```diff
@@ -47,7 +47,7 @@ export function HomePage(props: PageProps) {
             <QueueLookup/>
             <hr/>
             <Link to="/manage">
-                Manage Your Own Queues
+                Manage Queues
             </Link>
```
ssciolla, line 51: While it's not always possible, keeping a PR focused on one issue is preferred. This change for #219 involves different files/functionality, so it's a good candidate to be separated out. This would help us keep track of and QA changes as we prepare for a release. If you'd like to open a separate PR with this change, I will review it. If you aren't able to, I can also cherry-pick your commit here and submit it separately.

src/assets/src/components/queueManager.tsx
```diff
@@ -168,6 +168,14 @@ function QueueManager(props: QueueManagerProps) {
                 <Col md={2}><div id='created'>Created</div></Col>
                 <Col md={6}><div aria-labelledby='created'><DateDisplay date={props.queue.created_at} /></div></Col>
             </Row>
+            <Row noGutters className={spacingClass}>
+                <Col md={2}><div id='meetingLocation'>Meeting Location</div></Col>
+                <Col md={6}><div aria-labelledby='meetingLocation'>{props.queue.meeting_location}</div></Col>
+            </Row>
```
ssciolla, line 174: Nice idea, I thought about this, we may want to add something like this later on.

ssciolla (commented):
@vikaschanduri, thank you for your work and interest in the project. Because #326 was opened earlier than this (and has since been merged), I'm going to be closing this PR. I did read through your changes, and they looked solid -- nice work figuring out how all the pieces fit together. I left some minor feedback for you. If you'd like to open that separate PR for your work on #219, I can review it.

PR #332, opened 2021-05-05
Pin Twisted to <21 to fix prod startup

### LINKED ISSUE #331: Exception related to workers occurs with latest build
Though the project images build fine, upon startup in the QA environment, the application throws an exception during the worker setup process, finishing with `TypeError: SelectorEventLoop required, instead got: <uvloop.Loop running=False closed=False debug=False>`.

This appears to be an issue related to `django/channels` and its dependency `twisted`:
https://github.com/django/channels/issues/1639

jlost:
Fixes #331 
A better long term fix would be to switch to `poetry` instead of `pip`. Having a lockfile would prevent these types of surprises.

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -16,3 +16,4 @@ django-filter~=2.2.0
 channels~=2.4.0
 channels-redis~=2.4.2
 twilio~=6.45.4
+twisted<21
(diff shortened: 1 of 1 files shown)


ssciolla (approved):
I ran this locally by commenting out the default `docker-compose` command, as you suggested, and it worked for me; all the workers finished their setup. I'd say merge it.

It's kinda unfortunate that we have to specify a library we're not using explicitly, but doesn't sound like we have other easy options. Maybe it merits a comment in `requirements.txt`?

jlost:
> Maybe it merits a comment in `requirements.txt`?
Very good idea, I went ahead and did this.

PR #347, opened 2021-12-01
Specify "use_pmi: False" to circumvent setting problem

### LINKED ISSUE #346: Create Meeting with PMI Zoom setting disrupts meeting joining
Due to recent Zoom changes, if you have the "Use Personal Meeting ID (PMI) when scheduling a meeting" setting enabled on your Zoom account, when you launch the meeting as a host, you see the following error message from Zoom:
Screen Shot 2021-12-01 at 12 48 16 PM

It looks like we may be able to avoid this problem if we add a `settings` parameter, `use_pmi: False`.

ssciolla:
This PR aims to resolve #346.

### DIFF
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -163,6 +163,7 @@ class Backend(BackendBase):
                     "join_before_host": False,
                     "waiting_room": True,
                     "meeting_authentication": False,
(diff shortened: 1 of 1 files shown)


mfldavidson (approved):
Looks good, let's merge and test in dev or QA

PR #364, opened 2022-08-29
Remove Zoom backend metadata if any client error occurs during refresh (#363)

### LINKED ISSUE #363: Http status code for invalid token during refresh seems to have changed
We discovered that the error code thrown when a Zoom refresh token is invalid seems to have changed at some point. It was previously 401, but now seems to be 400. The code below needs to be changed to accept both or a range so that we can avoid issues in the future when a user's authorization with the tool is somehow removed. It also may make sense to add a username to log message so we can better assist if problems arise in the future.

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/b65aa19a75c29b38d9847c635e0dd8bd79c75d39/src/officehours_api/backends/zoom.py#L125-L131

ssciolla:
This PR aims to resolve #363.

### DIFF
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -131,6 +131,7 @@ class Backend(BackendBase):
                 zoom_meta = {}
                 user.profile.backend_metadata['zoom'] = zoom_meta
                 user.profile.save()
+                logger.info(f'Removed Zoom metadata for user {user.id} to prompt re-authorization.')
             resp.raise_for_status()
             token = resp.json()
             zoom_meta.update({
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -126,6 +126,8 @@ class Backend(BackendBase):
                 # The refresh_token was invalidated somehow.
                 # Maybe the user removed our Zoom app's auth.
                 # Force them to be prompted again.
+                # The specific code sent by Zoom has changed before,
+                # so it now checks for any client error during refresh.
                 zoom_meta = {}
                 user.profile.backend_metadata['zoom'] = zoom_meta
                 user.profile.save()
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -122,7 +122,7 @@ class Backend(BackendBase):
                 },
                 headers=cls._get_client_auth_headers(),
             )
-            if resp.status_code == 401:
+            if resp.status_code >= 400 and resp.status_code < 500:
                 # The refresh_token was invalidated somehow.
                 # Maybe the user removed our Zoom app's auth.
                 # Force them to be prompted again.


src/officehours_api/backends/zoom.py
```diff
@@ -122,7 +122,7 @@ def _get_access_token(cls, user: User) -> str:
                 },
                 headers=cls._get_client_auth_headers(),
             )
-            if resp.status_code == 401:
```
zqian, line 125: Could you please add a comment line here on why this change is needed, e.g. Zoom changed the API return status, etc.?

ssciolla: Sure, if you'd like me to.

ssciolla: How is this? https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/364/commits/a0596d19bf2fb08b8a7cdea8cbdc14fe6c938699

jlost (approved):
Looks good!

src/officehours_api/backends/zoom.py
zqian: looks good there.

mfldavidson (approved):
Perfect, I like the approach of "any 400 level error"

src/officehours_api/backends/zoom.py
```diff
@@ -122,10 +122,12 @@ def _get_access_token(cls, user: User) -> str:
...
+            if resp.status_code >= 400 and resp.status_code < 500:
                 # The refresh_token was invalidated somehow.
                 # Maybe the user removed our Zoom app's auth.
                 # Force them to be prompted again.
+                # The specific code sent by Zoom has changed before,
+                # so it now checks for any client error during refresh.
                 zoom_meta = {}
```
jonespm, line 131: This looks fine. I might log a info message when this happens with the `resp.status_code` and the `user.get_username().` just incase there are unexpected errors that come up , but this might be hard to test so I'm fine with it as-is.

jonespm (approved):

src/officehours_api/backends/zoom.py
ssciolla: Thanks! I was just working on this. I added a log message here (https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/364/commits/99ed7cc5b9508bccbab3a72fb4bbadecf159a07f) that includes the user ID (to keep the username itself out of the log) so we know when the reset has actually happened. I was discussing with @zqian this morning, the way the code is written, an exception will still be raised, so we'll see in the logs and in email what the status code was from Zoom.

jonespm: Okay that makes sense that it still hits the `raise_for_status`. We just didn't know the id of the user affected.

This is similar to Student Explorer where the user ID is logged when an exception occurs but even there there's a request to add it to the email too. I think for this app having it in the logs is fine as we want to reduce review and reduce the email frequency.

ssciolla: Yeah, handling exception and broken URL emails will have to be a separate issue, hopefully we can do that at some point.

PR #366, opened 2022-08-30
Modify expiration calculation, extract repeated code (#365)

### LINKED ISSUE #365: Calls to Zoom API don't use unexpired access token
In looking at and testing #363, at some point I realized that Zoom API calls always seem to first refresh the token; access tokens do not seem to be re-used, even though they are stored in `User.profile.backend_metadata`. Upon inspection, this seem to be because the math for the expiration time seems to be off. It appears it should be adding Zoom's `expires_in` value, not subtracting it.

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/374d8fd5346dfcbe28120ec27fd121c2881d467c/src/officehours_api/backends/zoom.py#L137-L141

There would be a consequence to adding this, in that authorization removed in the Zoom marketplace would need to be handled for both expired and unexpired tokens -- right now everything goes through the same path, so it's always handled. Doing it in both circumstances would require some refactoring.

ssciolla:
This PR aims to resolve #365.

### DIFF
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -122,7 +122,7 @@ class Backend(BackendBase):
 
     @classmethod
     def _get_access_token(cls, user: User) -> str:
-        zoom_meta = user.profile.backend_metadata['zoom']
+        zoom_meta = user.profile.backend_metadata['zoom'].copy()
         if time() > zoom_meta['access_token_expires']:
             logger.debug('Refreshing token')
             resp = requests.post(
@@ -142,13 +142,14 @@ class Backend(BackendBase):
                 cls._clear_backend_metadata(user)
             resp.raise_for_status()
             token = resp.json()
-            zoom_meta.update({
+            new_zoom_data = {
                 'refresh_token': token['refresh_token'],
                 'access_token': token['access_token'],
                 'access_token_expires': cls._calculate_expires_at(token['expires_in']),
-            })
+            }
+            user.profile.backend_metadata['zoom'].update(new_zoom_data)
             user.profile.save()
-        return zoom_meta['access_token']
+        return user.profile.backend_metadata['zoom']['access_token']
 
     @classmethod
     def _get_session(cls, user: User) -> requests.Session:
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -111,9 +111,20 @@ class Backend(BackendBase):
         return resp.json()
 
     @classmethod
+    def _clear_backend_metadata(cls, user: User) -> None:
+        user.profile.backend_metadata['zoom'] = {}
+        user.profile.save()
+        logger.info(f'Removed Zoom metadata for user {user.id} to prompt re-authorization.')
+
+    @classmethod
+    def _calculate_expires_at(cls, expires_in: int) -> float:
+        return time() + expires_in - cls.expiry_buffer_seconds
+
+    @classmethod
     def _get_access_token(cls, user: User) -> str:
         zoom_meta = user.profile.backend_metadata['zoom']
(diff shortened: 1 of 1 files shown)


src/officehours_api/backends/zoom.py
```diff
@@ -128,19 +139,17 @@ def _get_access_token(cls, user: User) -> str:
...
             resp.raise_for_status()
             token = resp.json()
-            zoom_meta.update({
+            new_zoom_data = {
                 'refresh_token': token['refresh_token'],
                 'access_token': token['access_token'],
-                'access_token_expires': time() - token['expires_in'] - cls.expiry_buffer_seconds,
-            })
+                'access_token_expires': cls._calculate_expires_at(token['expires_in']),
+            }
```
ssciolla, line 149: @jlost, let me know if this resolves your concerns with `zoom_meta`.

src/officehours_api/backends/zoom.py
```diff
@@ -170,6 +179,9 @@ def _create_meeting(cls, user: User) -> ZoomMeeting:
                 },
             },
         )
+        if resp.status_code == 401:
```
ssciolla, line 182: I think we're only likely to see `401` here if the token has been invalidated, but we could also use the client range if we were concerned. It's working like this for now.

jonespm:
@ssciolla Do you have a test plan for this? Is there something in the log to look for when a new authorization key is retrieved? How did you originally notice this a problem.

Logically this seems correct in calculating the expiration time and it works but I'm not sure how to tell that the token is persisting for the full time.

src/officehours_api/backends/zoom.py
```diff
@@ -110,10 +110,21 @@ def _spend_authorization_code(cls, code: str, request) -> ZoomAccessToken:
...
+        logger.info(f'Removed Zoom metadata for user {user.id} to prompt re-authorization.')
+
+    @classmethod
+    def _calculate_expires_at(cls, expires_in: int) -> float:
+        return time() + expires_in - cls.expiry_buffer_seconds
+
     @classmethod
     def _get_access_token(cls, user: User) -> str:
-        zoom_meta = user.profile.backend_metadata['zoom']
+        zoom_meta = user.profile.backend_metadata['zoom'].copy()
```
jonespm, line 125: Are you doing this just to better protect this from backend_metadata changes? I don't see `zoom_meta` actually modified anywhere in this method any longer and the only two uses of it is before `user.profile.backend_metadata` is modified. 

It could just be removed and replaced in those two cases with `user.profile.backend_metadata['zoom']`? I don't really see any preferences for preferred style patterns out there in Python.

ssciolla: The `copy` may not longer be necessary. @jlost raised a concern in DM that the reference to `zoom_meta` and subsequent updating could unintentionally persist values that we meant to clear using `_clear_backend_metadata`. I don't think it was likely to ever come up, because of the exception throwing, but I think we wanted to do some future proofing and stay away from this kind of possibility. I think I've avoided that by not updating `zoom_meta`, so maybe the copy can be removed, but I was doing it out of an abundance of caution.

jonespm (approved):
I ran through the test plan on the issue and it passes all the steps. Not sure if @jlost has any comments but it looks good to me.

src/officehours_api/backends/zoom.py
jlost: Since you're no longer mutating `zoom_meta`, I think it would be OK to remove the `.copy()`. It doesn't cause any harm though.

jlost (approved):

src/officehours_api/backends/zoom.py
ssciolla: I removed `.copy()`, thanks!

PR #368, opened 2022-09-01
Fixes #367 - Dockerfile does not build on M1/ARM

### LINKED ISSUE #367: Dockerfile does not build on M1/ARM
On M1 Mac this fails building with the error.

```
#0 2.411       Error: pg_config executable not found.
#0 2.411       
#0 2.411       pg_config is required to build psycopg2 from source.  Please add the directory
#0 2.411       containing pg_config to the $PATH or specify the full executable path with the
#0 2.411       option:
```
Looks fixed in this issue https://github.com/psycopg/psycopg2/issues/1286

And in the latest version of psycopg2-binary

`+    'psycopg2-binary~=2.9.3'
`

jonespm:
These are the minimal changes needed to get this running. There are
probably more dependency updates that would be nice to have.

Leaving as draft as it shouldn't make the emergency release but we'll want to look at this and other dependency updates.


src/requirements.txt
```diff
@@ -1,4 +1,4 @@
-django~=3.0.5
+django==3.2
```
ssciolla, line 1: @jonespm, could you please rebase this and then pin to a more specific release here? I'd like to get this in today so M1 folks can contribute.

ssciolla (changes requested):

src/requirements.txt
jonespm:
```suggestion
django~=3.2.16
```

jonespm: Yup, done.

ssciolla:
Thanks, I'll try running this soon.

ssciolla:
@jonespm, can you quickly add `DEFAULT_AUTO_FIELD` as in CAE like so? https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/e6fc3ac1167f86e7574ded62304fdb406cd1aee3/backend/settings.py#L149

src/requirements.txt
```diff
@@ -12,7 +12,7 @@ django-safedelete~=0.5.4
 jsonfield~=3.1.0
 django-redirect-to-non-www~=0.1.1
 drf-api-tracking~=1.6.0
-django-filter~=2.2.0
+django-filter~=2.4.0
```
ssciolla, line 15: Please bump to latest `22.1` (switched to CalVer I guess).

src/Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN apt-get -y install gcc
 RUN pip install \
     'gunicorn~=20.0.4' \
     'uvicorn[standard]~=0.12.2' \
-    'psycopg2-binary~=2.8.6'
+    'psycopg2-binary~=2.9.3'
```
ssciolla, line 30: Please bump to `2.9.5`.

ssciolla (approved):
Couple minor things it'd be good to do now, but otherwise looks good and I didn't see any issues in testing. You can merge once we've addressed those.

PR #382, opened 2023-02-22
Upgrade to Postgres 14, change client library setup (#381)

### LINKED ISSUE #381: Upgrade Postgres in docker-compose from 10 to 14.4
The operations team is going to upgrade the database to Postgres 14, they said specifically 14.4. We should upgrade our local docker-compose to match and start it up to see if anything obvious doesn't work. 

I asked if this was the latest version and it was commented that this was the latest version they saw available in AWS. Though newer minor versions may also be available soon as the most recently released is 14.7. 

Version 14 has support until November 2026. 

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/c9a05f25c3edcc0381be424b34df0519476f798d/docker-compose.yml#L25

ssciolla:
This PR aims to resolve #381.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     environment:
       - ALLOWED_HOSTS=*
       - DEBUG=True
-      - DATABASE_URL=postgresql://admin:admin@database/postgres
+      - DATABASE_URL=postgresql://admin:admin_pw@database/admin
       - FEEDBACK_EMAIL=office-hours-devs@umich.edu
     ports:
       - 8003:8001
@@ -22,10 +22,11 @@ services:
       - webpack_watcher
 
   database:
-    image: postgres:10.4-alpine
+    image: postgres:14.4-alpine
     environment:
       - POSTGRES_USER=admin
       - POSTGRES_DB=admin
+      - POSTGRES_PASSWORD=admin_pw
     volumes:
       - officehours-postgres-data:/var/lib/postgresql/data
 
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -23,11 +23,11 @@ WORKDIR /usr/src/app
 ENV WEB_CONCURRENCY=5
 
 RUN apt-get update
-RUN apt-get -y install gcc
+RUN apt-get -y install gcc libpq-dev
 RUN pip install \
     'gunicorn~=20.0.4' \
     'uvicorn[standard]~=0.12.2' \
-    'psycopg2-binary~=2.9.5'
+    'psycopg2~=2.9.5'
 COPY requirements.txt /tmp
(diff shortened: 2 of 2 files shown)


jonespm:
This looks good. Did you change from the binary to source [because of the installation notes](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) or some other reason? Does it add much time to the compilation?

Would you know why these packages are separate and not just in the requirements.txt? Only reason I can _think_ of is so they'll be cached in a separate layer.

jonespm (approved):

docker-compose.yml
```diff
@@ -22,10 +22,11 @@ services:
       - webpack_watcher
 
   database:
-    image: postgres:10.4-alpine
+    image: postgres:14.4-alpine
```
jonespm, line 25: I'd like to move off of alpine but we aren't changing anything and it isn't used in the deployment, I think it's _fine_ to stay on alpine here. But if we ever upgrade the node dependency in the Dockerfile, I'd switch that over to the `node:##-slim` tag.

ssciolla:
> This looks good. Did you change from the binary to source [because of the installation notes](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) or some other reason? Does it add much time to the compilation?

@jonespm, I was having issues with the binary file after the upgrade. I thought it might be because of M1, but I'm not sure. Do you have time to change it back and remove the the linux package? I'd be curious if you also see issues. It does seem to be recommended that you use this approach, but this isn't something I'm very knowledgeable about.

> Would you know why these packages are separate and not just in the requirements.txt? Only reason I can _think_ of is so they'll be cached in a separate layer.

I asked James about this a while ago; I think his rationale was that the database and server libraries shouldn't be coupled to the rest of the application, since some of them could be swapped out or a different `Dockerfile` could be written. At least that's my recollection.

docker-compose.yml
ssciolla: What's the downside to `alpine`? Just curious.

jonespm:
That makes sense the binary may not work.

I also noticed that the [psycopg](https://pypi.org/project/psycopg/) package has a newer version released at 3.1.8. This 2.9.5 is from 10/2022.

They have a slightly different way of installing this, either as
`pip install "psycopg[binary]"`
or
`pip install "psycopg[c]"`

ssciolla:
@jonespm, Django says to use `psycopg2`: https://docs.djangoproject.com/en/3.2/ref/databases/#postgresql-notes

jonespm:
I see, looks like it's going to be supported in Django 4.2, but there was a lot of work involved. 

https://github.com/django/django/pull/15687

PR #385, opened 2023-03-03
Update backend dependencies (#383)

### LINKED ISSUE #383: Update backend dependencies
The backend is due for an update. We're behind multiple minor versions on Python (see also #370), multiple major versions of Django Channels, and multiple major versions of Redis. Most of the other pip packages also have minor or patch versions that can be picked up, but some may need to be held back until the frontend is updated or removed/replaced due to Python/Django ecosystem changes.

ssciolla:
This PR aims to resolve #383.

To Do
- [x] Update Python (3.10, 3.11 came out less than 6 months ago)
- [x] Update Channels
  - [x] Update `redis`
    - [x] Need to update `kustomize` YAMLs
  - [x] Update channels libraries, make needed changes
  - [x] ~~Figure out event loop issue in `channels_redis` (issue 332 there)~~ (no longer an issue with 4.1.0)
  - [x] Use `uvicorn` with development server over installing `daphne` (was baked into channels before)
- [x] Replace `drf-rest-swagger` with `drf-spectacular`
  - [x] Resolve most errors/warnings with `extend_schema` and `extend_schema_field`
  - [x] Add documentation to `README` about using Swagger UI
- [x] Replace `jsonfield` with `django.models.JSONField`
  - Seems like [this](https://forum.djangoproject.com/t/deprecation-warning-in-unmaintained-3rd-party-app-used-in-old-migrations-django-jsonfield/10819/4) will work?
- [x] Update and fix `django-safedelete`
  - [x] Figure out bug

Resources
- https://channels.readthedocs.io/en/latest/index.html


src/officehours_api/serializers.py
```diff
@@ -83,6 +88,39 @@ class Meta:
...
+    def get_my_meeting(self, obj):
+        user = self.context['user']
+        my_meeting = (
+            obj.meeting_set.filter(attendees__in=[self.context['user']]).first()
+            if user.is_authenticated else None
+        )
+        if not my_meeting:
+            return None
+        serializer = NestedMyMeetingSerializer(my_meeting, context=self.context)
+        return serializer.data
```
ssciolla, line 121: This was moved because it was referenced in `MyUserSerializer` before it was defined.

src/requirements.txt
```diff
@@ -1,21 +1,17 @@
...
-jsonfield~=3.1.0
+django==3.2.18
+djangorestframework==3.14.0
+drf-spectacular==0.25.1
+mozilla-django-oidc==3.0.0
+whitenoise==6.3.0
+dj-database-url==1.2.0
+requests==2.28.2
+django-watchman==1.3.0
+django-webpack-loader==0.6.0
```
ssciolla, line 9: This package will need an update when we work on the frontend. It needs to stay in sync with the companion `npm` package.

src/officehours/settings.py
```diff
@@ -72,7 +72,7 @@ def str_to_bool(val):
 
 if DEBUG:
     INSTALLED_APPS += [
-        'drf_yasg',
+        'drf_spectacular',
```
ssciolla, line 75: `drf_spectacular` was recommended by `drf_yasg` maintainers for OpenAPI 3 schema generation. We've also used it with Canvas App Explorer.

src/officehours_api/migrations/0009_auto_20200325_1220.py
```diff
@@ -14,7 +15,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='meeting',
             name='backend_metadata',
-            field=jsonfield.fields.JSONField(null=True),
+            field=models.JSONField(null=True),
```
ssciolla, line 18: While this doesn't seem advisable, I did find a suggestion in the Django forms that editing the migration file would be an appropriate solution when migrating from a third-party implementation of a JSON field to the native `JSONField` released in Django 3.1: https://forum.djangoproject.com/t/deprecation-warning-in-unmaintained-3rd-party-app-used-in-old-migrations-django-jsonfield/10819

I tested locally and Django doesn't complain about unapplied changes after editng `models.py` to use the new one. And preexisting data loads correctly.

src/officehours_api/models.py
```diff
@@ -74,6 +73,8 @@ def get_users_with_emails(manager: models.Manager):
 
 class Queue(SafeDeleteModel):
     _safedelete_policy = SOFT_DELETE_CASCADE
+    deleted_by_cascade = None
```
ssciolla, line 76: The `1.2.0` release of `django-safedelete` introduced a field/column called `deleted_by_cascade`. However, it didn't supply defaults. The docs recommend and an issue I found suggest either adding this line or writing a migration to provide data for existing records. Since this field wouldn't do much for us (we have three fields using `SafeDeleteModel`, and only `Queue` has a soft delete policy), I think it's fine to opt out of this behavior. I'm open to adding a migration if someone thinks that's better.

src/requirements.txt
```diff
@@ -1,21 +1,16 @@
...
 django-redirect-to-non-www~=0.1.1
-drf-api-tracking~=1.6.0
-django-filter~=22.1
-channels~=2.4.0
-channels-redis~=2.4.2
-twilio~=6.45.4
-
-# Transitive dependency, pinned to fix an error. See: https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/331
-twisted<21
+drf-api-tracking==1.8.0
```
ssciolla, line 12: We'll need to keep an eye on this library. It hasn't been updated for a couple years, and doesn't explicitly support Django 3.1 or 3.2.

src/requirements.txt
```diff
@@ -1,21 +1,16 @@
...
-mozilla-django-oidc~=1.2.3
-whitenoise~=5.0.1
-dj-database-url~=0.5.0
-requests~=2.23.0
-django-watchman~=1.1.0
-django-webpack-loader~=0.6.0
-django-safedelete~=0.5.4
-jsonfield~=3.1.0
+django==3.2.18
+djangorestframework==3.14.0
```
ssciolla, line 2: You'll notice I changed all these dependencies to hard pins from allowing patch versions. I can be convinced to go back, but it seems preferable to have reproducible builds and avoid possible bugs in these patches. Most of T&L's projects use pins.

ssciolla:
Found a bug here related to `django-safedelete`. If you delete a queue with existing meetings, it throws an internal server error. I may have to walk back the version update or add a data migration.

src/officehours_api/models.py
```diff
@@ -206,7 +209,7 @@ def delete(self, *args, **kwargs):
         # Trigger m2m "remove" signals for attendees
         self.attendees.remove(*self.attendees.all())
         self.save()
-        super().delete(*args, **kwargs)
+        return super().delete(*args, **kwargs)
```
ssciolla, line 212: I fixed a bug where an internal server error was thrown when deleting a queue with meetings. This change fixed it. See https://github.com/makinacorpus/django-safedelete/issues/212 and https://docs.djangoproject.com/en/3.2/ref/models/instances/#deleting-objects

docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
       - 8003:8001
     volumes:
       - ./src:/usr/src/app:z
-    command: "python manage.py runserver 0.0.0.0:8001"
+    command: "python -m uvicorn officehours.asgi:application --host=0.0.0.0 --port=8001 --reload"
```
ssciolla, line 18: I decided to bump to `4.0.0` for Django Channels, since the issue that was blocking us seems to have been resolved in `channels_redis` 4.1.0. In Channels 4, they decoupled the implementation from the `daphne` 4.0.0. You can still use `daphne` for development, but it's a separate package and it needs to be added to `APPS` in Django. I figured rather than having two ASGI server implementations installed, we may as well use `uvicorn` for local development. It has a `--reload` feature and looks to work well, with really just some logging changes that I can see. I don't think there will be logging changes to production since we're already using `uvicorn` -- unless there was a change in `uvicorn` itself picked up by the version change.

jonespm: This sounds good, I've switched to gunicorn for development for other projects that aren't upgraded to use uvicorn yet.

src/officehours/settings.py
jonespm: Looks like a good switch

src/officehours_api/models.py
jonespm: Looks good from the change in safedelete and probably good practice to do anyway.

src/requirements.txt
jonespm: This is probably better now with dependabot watching for these. It's not going to be able to tell as well if we need an upgrade with the ~= versions.

src/requirements.txt
jonespm: Can we add this as a comment here? Or is it mentioned on the issue for the update for the frontned?

src/requirements.txt
jonespm: Funny that this was a fork of https://github.com/aschn/drf-tracking. Abandoned projects/forks are still tough to deal with. 

I don't see any forks of this or the other that are more active for this action. So yeah, we'll have to see if anyone else forks or if this gets updates.

src/requirements.txt
ssciolla: It's changed in this commit from the FE update (https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/405/commits/4ebecb2511d1377afc21ed62d835a77ea55808e1), I'm keeping track of it, I dunno if a comment is really needed.

src/officehours/routing.py
```diff
@@ -1,12 +1,22 @@
...
+
+asgi_application = get_asgi_application()
+
 application = ProtocolTypeRouter({
     # (http->django views is added by default)
-    'websocket': AuthMiddlewareStack(
-        URLRouter(
-            officehours_api.routing.websocket_urlpatterns
+    'http': asgi_application,
+    'websocket': AllowedHostsOriginValidator(
```
jonespm, line 15: Was this just done for additional security or was there some other reason here?

ssciolla: Yeah, I saw it in some of the Channels tutorials and wondered why it wasn't included here. Seemed like an easy thing to add. Here's the documentation on it: https://channels.readthedocs.io/en/stable/topics/security.html

ssciolla:
Thanks for the review, @jonespm! Much appreciated. I'm gonna merge this and try to get it running in QA. If you come across any bugs from these changes, please feel free to open an issue and assign me.

PR #393, opened 2023-03-26
Align queue language- problem 219

jzenas:
Changed line 50 of home.tsx to align the language to be "Manage Queues" so that the home page matches the drop down menu.


ssciolla (approved):
Looks good! We've gone back and forth on whether to indent with the `<>...</>`, but standard linting configuration seems to suggest it. Other spacing additions/fixes are good.

PR #395, opened 2023-03-28
Bump postgres to 14.7 (#394)

### LINKED ISSUE #394: Bump `postgres` image version to 14.7
We were alerted by the database team that they'll actually be using Postgres 14.7 initially. We should bump what is in `docker-compose.yml` to be consistent for local testing.

ssciolla:
This PR aims to resolve #394.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,7 +22,7 @@ services:
       - webpack_watcher
 
   database:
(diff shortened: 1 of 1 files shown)


zqian (approved):
Test was a success with localhost and docker compose

PR #396, opened 2023-03-29
Fixes #336 Allow admins to edit users who haven't authorized any backends

### LINKED ISSUE #336: Allow admins to edit users who haven't authorized any backends
Not super high priority but should probably fix eventually.

When attempting to edit a user in admin, such as to add or remove staff status, the form will not save if the user has not authorized any backends. The `Backend metadata` field is auto-populated with an empty dictionary `{}` and is a required field, but the empty dictionary doesn't appear to be considered content and therefore is counted as an empty field, preventing saving.

Inserting the following text into the field resolves the error and allows the form to be saved, and the user is still prompted to authorize Zoom as they normally would if the field was set as `{}`
```
{
    "zoom": {}
}
```

image

As an admin, I need to be able to edit users who haven't authorized any backends without making adjustments to the contents of the `Backend metadata` field.

Tombow1:
This PR fixes #336, an issue that occurs when an admin attempts to edit a user and the backend_metadata field considers `{}` as empty.

The issue arises because JSONField treats the following as empty values:

`EMPTY_VALUES = (None, '', [], (), {})`

As a result, the auto-filled `{}` cannot be directly saved when the field is required. This PR addresses the issue by adding the `blank="True"` attribute to `backend_metadata`, allowing `{}` to be saved.

Since the original `backend_metadata` field permits null, it should be safe to make the field not required.

### DIFF
--- a/src/officehours_api/models.py
+++ b/src/officehours_api/models.py
@@ -53,7 +53,7 @@ class Profile(models.Model):
     phone_number = models.CharField(max_length=20, default="", blank=True, null=False)
     notify_me_attendee = models.BooleanField(default=False)
     notify_me_host = models.BooleanField(default=False)
-    backend_metadata = JSONField(null=True, default=dict)
+    backend_metadata = JSONField(null=True, default=dict, blank=True)
 
     @property
     def authorized_backends(self):


ssciolla:
Hi, @Tombow1, I'm going to test this now.

FYI, it's best practice to open your PRs from an issue branch on your fork (e.g. `issue-336-blank-metadata`), rather than `master`. That allows you to keep your fork's `master` in sync with the main repository and simultaneously work on multiple issues (each in a separate branch) all based on the current version of the application.

PR #399, opened 2023-04-05
Refactor search form and URL to use query (#398)

### LINKED ISSUE #398: Search queries with `/` in the name fall through React Router, end in 404
**Expected behavior**

Search queries should be able to include special characters like `/` so that queues using these characters are properly found.

**Actual behavior**

When you search for a queue with a `/` in the name, the client-side routing doesn't catch this and it falls through to the Django app, which then throws a 404.

**Commentary**

This seems to happen even if you add `encodeURIComponent` to the lookup on the homepage (see `assets/components/home.tsx`).

ssciolla:
This PR aims to resolve #398. It replaces the URL parameter with a query parameter, which seems to be both better fitted to the use case (search term is dynamic, not a permanent resource) and to the HTML form element which uses it by default.

Sorry, @jonespm, I had already started working on this. Hopefully you're willing to test/review this instead.

Test setup

- Create a queue with no special characters or punctuation in the name.
- Create a queue with a "/" in the name.
- Create a queue with a "#" in the name.
- Create a queue with other special characters in the name.

Test cases

From the home page:
- Click "Search Queues" without any input. A pop-up should indicate the field is required.
- Enter a fragment of the name with the "/", including the "/". Check the queues listed to ensure the right queues are present.
- Enter a fragment of the name with the "#", including the "#". Check the queues listed to ensure the right queues are present.
- Enter a fragment of the name with special characters. Check the queues listed to ensure the right queues are present.

In the URL bar:
- Navigate to "https://localhost:8003/search/". Note that no results are returned and the search term is an empty string (`undefined`).
- Navigate to "https://localhost:8003/search/?term={something}" where {something} is an encoded search term. Check the queues listed to ensure the right queues are present.

Resource(s)
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form

### DIFF
--- a/src/assets/src/components/search.tsx
+++ b/src/assets/src/components/search.tsx
@@ -20,7 +20,7 @@ export function SearchPage(props: PageProps) {
         setSearchResults
     );
     useEffect(() => {
-        if (term) {
+        if (term !== undefined) {
             doSearch(term);
         } else {
             setSearchResults([]);
--- a/src/assets/src/components/search.tsx
+++ b/src/assets/src/components/search.tsx
@@ -13,7 +13,7 @@ export function SearchPage(props: PageProps) {
     if (!props.user) {
         redirectToLogin(props.loginUrl);
     }
-    let term = (new URLSearchParams(props.location.search)).get('term') ?? undefined;
+    const term = (new URLSearchParams(props.location.search)).get('term') ?? undefined;
     const [searchResults, setSearchResults] = useState(undefined as ReadonlyArray<QueueBase> | undefined);
     const [doSearch, searchLoading, searchError] = usePromise(
         (term: string) => apiSearchQueue(term),
--- a/src/assets/src/components/search.tsx
+++ b/src/assets/src/components/search.tsx
@@ -9,22 +9,25 @@ import { redirectToLogin } from "../utils";
 import { PageProps } from "./page";
 
 
-interface SearchPageParams {
-    term: string;
-}
-
-export function SearchPage(props: PageProps<SearchPageParams>) {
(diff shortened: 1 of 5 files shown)


jonespm (approved):
Looks to work fine to me.

src/assets/src/components/search.tsx
```diff
@@ -9,22 +9,25 @@ import { redirectToLogin } from "../utils";
...
-export function SearchPage(props: PageProps<SearchPageParams>) {
+export function SearchPage(props: PageProps) {
     if (!props.user) {
         redirectToLogin(props.loginUrl);
     }
-    const term = props.match.params.term;
+    let term = (new URLSearchParams(props.location.search)).get('term') ?? undefined;
+    if (term !== undefined) {
+        term = decodeURIComponent(term);
+    }
```
ssciolla, line 19: I'm not really sure this is necessary anymore. I think since I changed this to a search query and am using `URLSearchParams`, things seem to be decoded already before this logic.

ssciolla: Yeah, seems like `URLSearchParams` does decode: https://stackoverflow.com/a/45516812

ssciolla:
@jonespm, could you take another quick look at this? I modified a couple things, mainly removing the decoding in `search.tsx` since `URLSearchParams` does that for you. The `encodeURIComponent` in `api.tsx` seems necessary since without it characters like "#" and "&" are interpreted as part of the query string.

jonespm (approved):
I tested this out again and it still looks to work fine. It seems like there was some confusion whether or not decoding was needed with URlSearchParams but it does decode the value.

Issue #306, opened 2021-01-15
Enforce Meeting.backend_type in Meeting.queue.allowed_backends via API


ssciolla:
We don’t seem to be enforcing that `Meeting.backend_type` needs to be present in `Queue.allowed_backends` when a meeting is created via the API. Seems like we could add another `validate` function to `MeetingSerializer` to do this. We don't allow people to do this in the UI.

There is also a question of whether we would want to prevent changes to `Queue.allowed_backends` if it involves removing a type that meetings currently use. We enforce this explicitly in the UI, but it doesn't appear to be done for the API. Enforcing this also in the API would make sense, but we would want to be careful not to disallow this globally in the backend, as that would conflict with the phasing out logic (and desire to let started meetings hang around) presented in PR #305.

This is related to #289.

ssciolla:
Test plan

@zqian, best way to test this is probably locally via the Swagger UI. I can help get you set up, but basically you'd create a queue that only allows Zoom or in person, and then try to create a meeting via Swagger with the backend type you did NOT specify as being allowed.

zqian:
Tested on localhost with an in-person queue. Trying to add a meeting via swagger UI for Zoom type, 
`{
  "queue": 1,
  "attendee_ids": [
    1
  ],
  "assignee_id": 1,
  "backend_type": "zoom"
}`

and got the following error in response:

`{
  "non_field_errors": [
    "zoom is not one of the queue's allowed backend types (['inperson'])"
  ]
}`

Testing is a success.

PR #400, opened 2023-04-06
Fixed issue #306: Enforce Meeting.backend_type in Meeting.queue.allowed_backends via API

Frank-duuuu:
Added conditions to `validate()` in `MeetingSerializer`, checking if the `backend_type` of this meeting is in the `allowed_backends` of this queue. Additionally, added unit tests to test the validation logic in `MeetingSerializerTestCase`.

This change ensures that the backend type of a meeting is only allowed if it matches the allowed backends specified for the queue. This prevents incorrect backend types from being associated with meetings and causing issues downstream.

Please review and merge this pull request if it meets the project's standards.

### DIFF
--- a/src/officehours_api/serializers.py
+++ b/src/officehours_api/serializers.py
@@ -260,7 +260,8 @@ class MeetingSerializer(serializers.ModelSerializer):
 
     def validate(self, attrs):
         '''
-        Ensure the assignee is a host.
+        Ensure the assignee is a host,
+        and the meeting's backend type is in the queue's allowed backend type.
         '''
         queue = self.instance.queue if self.instance else attrs["queue"]
         hosts = queue.hosts.all()
--- a/src/officehours_api/serializers.py
+++ b/src/officehours_api/serializers.py
@@ -268,5 +268,5 @@ class MeetingSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError("Assignee must be a host!")
         backend_type = attrs["backend_type"]
         if backend_type not in queue.allowed_backends:
-            raise serializers.ValidationError("Invalid backend type!")
+            raise serializers.ValidationError(f"{backend_type} is not one of the queue's allowed backend types ({queue.allowed_backends})")
         return attrs
--- a/src/officehours_api/serializers.py
+++ b/src/officehours_api/serializers.py
@@ -266,4 +266,7 @@ class MeetingSerializer(serializers.ModelSerializer):
         hosts = queue.hosts.all()
         if attrs.get("assignee") and attrs["assignee"] not in hosts:
             raise serializers.ValidationError("Assignee must be a host!")
+        backend_type = attrs["backend_type"]
+        if backend_type not in queue.allowed_backends:
+            raise serializers.ValidationError("Invalid backend type!")
         return attrs
--- a/src/officehours_api/serializers.py
+++ b/src/officehours_api/serializers.py
@@ -267,7 +267,6 @@ class MeetingSerializer(serializers.ModelSerializer):
         hosts = queue.hosts.all()
         if attrs.get("assignee") and attrs["assignee"] not in hosts:
             raise serializers.ValidationError("Assignee must be a host!")
-        backend_type = attrs["backend_type"]
-        if backend_type not in queue.allowed_backends:
-            raise serializers.ValidationError(f"{backend_type} is not one of the queue's allowed backend types ({queue.allowed_backends})")
+        if attrs.get("backend_type") and attrs["backend_type"] not in queue.allowed_backends:
+            raise serializers.ValidationError(f"{attrs['backend_type']} is not one of the queue's allowed backend types ({queue.allowed_backends})")
         return attrs
--- a/src/officehours_api/tests.py
+++ b/src/officehours_api/tests.py
@@ -169,7 +169,7 @@ class NotificationTestCase(TestCase):
 
 class MeetingSerializerTestCase(TestCase):
     """
-    Test the scenarios where the backend type is invalid, valid,
+    Test the scenarios where the backend type is invalid, is valid,
     is not updated but something else is, and is updated to be invalid.
     """
 
--- a/src/officehours_api/tests.py
+++ b/src/officehours_api/tests.py
@@ -170,7 +170,7 @@ class NotificationTestCase(TestCase):
 class MeetingSerializerTestCase(TestCase):
     """
     Test the scenarios where the backend type is invalid, valid,
-    patched to none, and patched to invalid.
+    is not updated but something else is, and is updated to be invalid.
     """
 
     def setUp(self):
--- a/src/officehours_api/tests.py
+++ b/src/officehours_api/tests.py
@@ -210,13 +210,14 @@ class MeetingSerializerTestCase(TestCase):
     def test_patch_backend_type_none(self):
         meeting = Meeting(**{
             'queue': self.inperson_queue,
+            'attendee': self.user2,
             'agenda': 'test agenda',
-            'assignee_id': self.user1.id,
+            'assignee': self.user1,
             'backend_type': 'inperson'
         })
         # Patch the meeting without a backend type
         data = {
-            'attendee_ids': [self.user2.id],
+            'agenda': 'patch the meeting without a backend type',
         }
         serializer = MeetingSerializer(meeting, data=data, partial=True)
         valid = serializer.is_valid(raise_exception=False)
@@ -226,13 +227,13 @@ class MeetingSerializerTestCase(TestCase):
     def test_patch_backend_type_invalid(self):
         meeting = Meeting(**{
             'queue': self.inperson_queue,
+            'attendee': self.user2,
             'agenda': 'test agenda',
-            'assignee_id': self.user1.id,
+            'assignee': self.user1,
             'backend_type': 'inperson'
         })
         # Patch the meeting to an invalid backend type
         data = {
-            'attendee_ids': [self.user2.id],
             'backend_type': 'zoom'
         }
         serializer = MeetingSerializer(meeting, data=data, partial=True)
--- a/src/officehours_api/tests.py
+++ b/src/officehours_api/tests.py
@@ -6,6 +6,10 @@ from twilio.base.exceptions import TwilioRestException
 
 from officehours_api.models import User, Queue, Meeting
 
+from officehours_api.serializers import MeetingSerializer
+
+from rest_framework.exceptions import ValidationError
+
 
 @override_settings(TWILIO_ACCOUNT_SID='aaa', TWILIO_AUTH_TOKEN='bbb', TWILIO_MESSAGING_SERVICE_SID='ccc')
 class NotificationTestCase(TestCase):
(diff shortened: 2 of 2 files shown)


src/officehours_api/serializers.py
```diff
@@ -266,4 +266,7 @@ def validate(self, attrs):
         hosts = queue.hosts.all()
         if attrs.get("assignee") and attrs["assignee"] not in hosts:
             raise serializers.ValidationError("Assignee must be a host!")
+        backend_type = attrs["backend_type"]
+        if backend_type not in queue.allowed_backends:
+            raise serializers.ValidationError("Invalid backend type!")
```
ssciolla, line 271: Nice work. I'll review this more thoroughly next week. Before that, could you update this validation error message to be more detailed? Something like `f"{backend_type} is not one of the queue's allowed backend types (f{queue.allowed_backends})"`

Frank-duuuu: Of course! The error message has been updated in the commit ([Changed Error Output](https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/400/commits/9a04f254686ff5dd59356252b8d904599a31ebb3)).

src/officehours_api/serializers.py
```diff
@@ -260,10 +260,14 @@ def validate_queue(self, queue):
...
+        Ensure the assignee is a host,
+        and the meeting's backend type is in the queue's allowed backend type.
         '''
         queue = self.instance.queue if self.instance else attrs["queue"]
         hosts = queue.hosts.all()
         if attrs.get("assignee") and attrs["assignee"] not in hosts:
             raise serializers.ValidationError("Assignee must be a host!")
+        backend_type = attrs["backend_type"]
+        if backend_type not in queue.allowed_backends:
+            raise serializers.ValidationError(f"{backend_type} is not one of the queue's allowed backend types ({queue.allowed_backends})")
```
ssciolla, line 272: So I did a little testing, and I found some issues with this. I think we need to use the same syntax we did for `assignee`, where we check whether it exists using `attrs.get(...)` before we use it. There are a couple ways in the application that a meeting can be `PATCH`ed or updated, and it's not a given that the `backend_type` will be included. Once this is modified, it probably makes sense to add a couple test cases for the patching as well.

Let me know if you need assistance formulating those test cases. There is a PATCH endpoint for meetings in the Swagger UI, and you can get the meeting IDs from the Django admin site or the database.

ssciolla (changes requested):
So I did a little testing, and this needs some modification. Specifically, see my last comment about a bug when the meeting is updated without the `backend_type`. The other comments are mostly about style and code organization -- nice to have but not as important as the bug of course.

src/officehours_api/serializers.py
Frank-duuuu: We have modified the code and added a couple of tests in the test file. We are considering adding a test case (`test_patch_backend_type_invalid()`) in which the meeting is valid first (in person), then we patch it to an invalid type (zoom), and we would get the expected error message. However, since we don't have the credential of zoom type, the error message wouldn't be the one we expected, so the test would not pass until the credential is included. Should we leave the test here or just delete it?

ssciolla: Looks like you made the right change to the validation. Let me take a look at the tests and I'll let you know if I think they are reasonable. Yes, it's probably a bit weird to test without both backend types enabled.

Frank-duuuu: I figured out one naive way to solve this is to run the test and set the environment variable at runtime. So it could pass all tests if we just run `docker-compose run -e ZOOM_CLIENT_ID="test" -e ZOOM_CLIENT_SECRET="test" web python manage.py test officehours_api.tests.MeetingSerializerTestCase` by setting the ZOOM environment variables on the go.

ssciolla (commented):
Thanks for fixing the bug and adding tests. It looks good. Can you add the `skipIf` decorator to the `zoom` tests and consider moving some of the queue set up to `setUp`? Once that is done, I will approve!

src/officehours_api/tests.py
```diff
@@ -163,3 +167,95 @@ def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
...
+    Test the scenarios where the backend type is invalid, valid,
+    patched to none, and patched to invalid.
+    """
+
+    def setUp(self):
+        self.user1 = User.objects.create(username='user1')
+        self.user2 = User.objects.create(username='user2')
+
+    def test_backend_type_invalid(self):
+        queue = Queue.objects.create(name='test queue', status='open', allowed_backends=['zoom'])
```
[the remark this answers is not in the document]
Frank-duuuu: Yeah, using skipIf seems like a valid solution. And thanks for reminding us of the duplicate code issue. We have fixed that in the latest commit (eb34d24).

src/officehours_api/tests.py
```diff
@@ -163,3 +169,92 @@ def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
...
+            'attendee_ids': [self.user2.id],
+            'agenda': 'test agenda',
+            'assignee_id': self.user1.id,
+            'backend_type': 'zoom'
+        }
+        serializer = MeetingSerializer(data=data)
+        with self.assertRaises(ValidationError) as cm:
+            serializer.is_valid(raise_exception=True)
+        error = str(cm.exception.detail['non_field_errors'][0])
+        self.assertEqual(error, "zoom is not one of the queue's allowed backend types (['inperson'])")
```
ssciolla, line 260: I don't know if this is really testing what happens a meeting is patched. I think we need to something more like what is described here: https://www.django-rest-framework.org/api-guide/serializers/#partial-updates

I think it could be something like this, where you create a model instance and then only pass the updated data.
```
        meeting = Meeting(**{
            'queue': self.inperson_queue.id,
            'attendee_ids': [self.user2.id],
            'agenda': 'test agenda',
            'assignee_id': self.user1.id,
            'backend_type': 'inperson'
        })
        # Patch the meeting to an invalid backend type
        data = {
            'backend_type': 'zoom'
        }
        serializer = MeetingSerializer(meeting, data=data, partial=True)
```
Can we use this approach for the two patch cases?

ssciolla (commented):
Thanks for the updates. I think we need to tweak how we're testing partial updates, but otherwise we're good to go.

src/officehours_api/tests.py
Frank-duuuu: We tried to achieve something like this but failed to find this partial-updates documents. Thank you so much for the updates! We have added it to our latest commit `d0e34cc`!
For the patch test cases, now we would only assert once after the meeting is patched. So the flow would be to create a new meeting, patch the meeting with new info, and run the validation.

src/officehours_api/tests.py
```diff
@@ -163,3 +165,78 @@ def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
...
+        meeting = Meeting(**{
+            'queue': self.inperson_queue,
+            'agenda': 'test agenda',
+            'assignee_id': self.user1.id,
+            'backend_type': 'inperson'
+        })
+        # Patch the meeting without a backend type
+        data = {
+            'attendee_ids': [self.user2.id],
+        }
```
ssciolla, line 220: Sorry about all the back and forth. Can you specify the attendee in the setup in both these cases? Then maybe just patch the `agenda` in this one, and then the `backend_type` in the other? In the UI, the attendee is basically always set on the initial meeting creation and then never changed. We may as well try to make these realistic or follow real-world usage.

Frank-duuuu: Of course! I thought we cannot assign attendees when initialization because when I tried to include `attendee_ids` it would throw an error `Meeting() got an unexpected keyword argument 'attendee_ids'`. But then I realized I forgot to reference the definition of Meeting() in models.py and I just need to include `attendee` from the setup. So now it would work I think!

ssciolla: You're right, I wasn't thinking this through completely. Also apparently I'm a little rusty with the Django ORM. What your doing isn't throwing an error, but it's also not creating an attendee. We have to use special methods to establish this relationship with the other models (`Attendee`, `User`) since it's a [Many-to-Many Relationship](https://docs.djangoproject.com/en/4.2/topics/db/examples/many_to_many/).

I think this is probably the appropriate way to create attendees.

```
        meeting = Meeting(**{
            'queue': self.inperson_queue,
            'agenda': 'test agenda',
            'assignee': self.user1,
            'backend_type': 'inperson'
        })
        meeting.save()
        meeting.attendees.add(self.user2)
        meeting.save()
        logger.info(meeting)
        logger.info(meeting.attendees.all())
```

When debugging this, I recommend adding a logger to the file using `logging.getLogger` and then using `logger.info` to look at the state of the `Meeting` or attendees, like I am above.

This doesn't really affect your tests, but we may as well get the attendee adding right. Could you update these cases one last time please?

Frank-duuuu: Of course, should I define the logger object `logger = logging.getLogger(__name__)` at the beginning of this file or just within those two test case functions?

ssciolla: Beginning of the file. You don't have to commit that though, it's just for your debugging/exploration.

Frank-duuuu: Noted! So I just need to change them into this and it should be fine for commit right?
```
meeting = Meeting(**{
            'queue': self.inperson_queue,
            'agenda': 'test agenda',
            'assignee': self.user1,
            'backend_type': 'inperson'
        })
        meeting.save()
        meeting.attendees.add(self.user2)
        meeting.save()
```

ssciolla: Yep, looks good.

Frank-duuuu: I have pushed the latest commit([52d435a](https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/400/commits/52d435af99e863f2804a646e6d8627cee024ac23)) to resolve this.

src/officehours_api/tests.py
```diff
@@ -163,3 +165,83 @@ def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
             receivers >=
             {'+15555550000', '+15555550001'}
         )
+
+
+class MeetingSerializerTestCase(TestCase):
+    """
+    Test the scenarios where the backend type is invalid, valid,
+    patched to none, and patched to invalid.
```
ssciolla, line 173: I think we should update the language here. The `test_patch_backend_type_none` is not really about setting the backend type to `None`, it's about updating another attribute and making sure everything still works. This was the bug I found before; we were changing the assignee or the agenda on the meeting, but because the backend type was always validated (we didn't have the `if attrs.get("backend_type")` check), we were accessing a key that didn't exist in the partial data we were passing and getting internal server errors.

```suggestion
    is not updated but something else is, and is updated to be invalid.
```

src/officehours_api/tests.py
```diff
@@ -163,3 +165,83 @@ def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
...
+            'attendee_ids': [self.user2.id],
+            'agenda': 'test agenda',
+            'assignee_id': self.user1.id,
+            'backend_type': 'inperson'
+        }
+        serializer = MeetingSerializer(data=data)
+        valid = serializer.is_valid(raise_exception=False)
+        self.assertTrue(valid)
+
+    def test_patch_backend_type_none(self):
```
ssciolla, line 210: In line with the other comment, I think this makes it a bit clearer.
```suggestion
    def test_patch_without_backend_type(self):
```

ssciolla (approved):
Hi @Frank-duuuu, thanks! I'm approving. Can you review/accept the language changes to the docstring and method name? I think those changes make things a bit clearer. After that, you can click "Rebase and merge" (if it's not available, let me know, I can do it).

src/officehours_api/tests.py
```diff
@@ -163,3 +165,83 @@ def test_meeting_start_notifies_attendees(self, mock_twilio: mock.MagicMock):
             receivers >=
             {'+15555550000', '+15555550001'}
         )
+
+
+class MeetingSerializerTestCase(TestCase):
+    """
+    Test the scenarios where the backend type is invalid, valid,
```
ssciolla, line 172:
```suggestion
    Test the scenarios where the backend type is invalid, is valid,
```

ssciolla:
@Frank-duuuu, done, thanks again for the contribution!

Issue #298, opened 2020-12-15
Fix non-obvious hortizontal scrolling on mobile


ksofa2:
On phone browsers the “Meeting Actions” column in the queue manage page requires a non-obvious horizontal scroll

ksofa2:
For example,
![Remote Office Hours Queue - Office Hours](https://user-images.githubusercontent.com/152920/102272693-27287200-3eef-11eb-839c-0ce0563f46cf.jpeg)
 on Safari on an iPhone 11

vjcao:
May I work on this issue?

mfldavidson:
Oh hey @vjcao ! Yes, please go ahead and work on this issue and open a PR with your work when ready. We are not actively developing right now but we can still potentially incorporate your work into a future release. Thank you so much for your interest in contributing! We really appreciate it.

HemanginiShah:
Test Successful.  Now a gray bar shows up at the bottom window when moved horizontally

ssciolla:
Hi @HemanginiShah, test condition should be that the scroll bar is always visible on the management page when the screen is small. This should be the case on Chrome and Safari browsers.

HemanginiShah:
Ok, thanks for the clarification.



*With Warm Regards,**  <https://www.name-coach.com/hemangini-shah>*

 *Hemangini Shah <https://www.name-coach.com/hemangini-shah>, PMP , *BI
Analyst Senior, ITS

Issue #369, opened 2022-09-19
Broken INTERNAL link: Add missing favicon.ico and robots.txt


jonespm:
Probably get a dozen or so messages a day about this missing file

> [ROHQ] Broken INTERNAL link on officehours.it.umich.edu
> 
> Referrer: https://officehours.it.umich.edu/
> Requested URL: /favicon.ico

jonespm:
Some other broken links that seem to be common and are not really an error

 /static/css/srs-banner.png
/robots.txt

Frank-duuuu:
Hi! I am reaching out on behalf of a group of students from EECS 481 Software Engineering, and we are looking for an open-source project to contribute to for our final project in the next three weeks. We are interested in your project and would love to know if we can assist in any way.

I saw this issue with the label "good first issue" and was wondering if we could provide any help. We would appreciate it if you could give us some first steps on handling this issue. Alternatively, if there is any other way we can contribute to your project, please let us know.

ssciolla:
I see the following items needing to be resolved to close this issue:

- The missing `favicon.ico`: There is actually a `favicon.png` already in the project, so I think we just need to add a `.ico` version to make alerts about this go away. It's probably worth doing some research about whether both formats are required or recommended. I found this in a quick search: https://stackoverflow.com/questions/1344122/favicon-png-vs-favicon-ico-why-should-i-use-png-instead-of-ico
- The missing `srs-banner.png`: This file is referenced in our CSS, but I don't think it's actually present or being served. This should likely be resolved by removing the CSS that mentions it (of course, we should make sure there are no changes once it is removed).
- The missing `robots.txt`: We should add a `robots.txt`, likely "disallow"-ing or instructing bots to not scrape anything but the home page. Most of the site is behind SSO. Looks like there are some articles like this out there that would help: https://www.cloudflare.com/learning/bots/what-is-robots-txt/

Frank-duuuu:
> Hi @Frank-duuuu, thanks for your interest in contributing to the project. I'm in contact with another student group about this issue; I'm not sure if you're part of that group or not.
> 
> Would you mind emailing our group at [its.tl.dev@umich.edu](mailto:its.tl.dev@umich.edu) and starting a conversation with us there? We can assist in finding an issue for you to work on if this one is already being taken.

Thank you so much for your offering. I have sent out the email. Hope to see you guys soon!

ssciolla:
Test Plan

[...]

PR #402, opened 2023-04-08
fixed issues 369 and 298

lialex123:
I fixed the issues [#369](https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/369) and [#298](https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/298). I added css to fix up #298. I added the necessary files and linked them to the website for #369. I also created tests for #369 under officehours_ui folder.


src/officehours_ui/static/css/main.css
```diff
@@ -560,6 +559,22 @@ h3.alert-heading {
...
+}
+.table-responsive::-webkit-scrollbar-thumb {
+    border-radius: 8px;
+    border: 2px solid white;
+    background-color: rgba(0, 0, 0, .5);
+}
+.table-responsive::-webkit-scrollbar-track { 
+    background-color: #fff; 
+    border-radius: 8px; 
+}
```
ssciolla, line 576: Couple things here.

1) This seems to work on Chrome/Safari on Mac OS. It looks like we may need to modify the `scrollbar-color` and `scrollbar-width` if we want the styling to be consistent on Firefox. See https://css-tricks.com/the-current-state-of-styling-scrollbars-in-css/ I think you have to change a setting on Mac OS to get the scrollbar to always show in Firefox however.

2) Typically we try to open one PR for each issue. Could you possibly move this to a separate branch and open a separate PR?

src/officehours_ui/urls.py
```diff
@@ -16,4 +17,6 @@
     path('add_queue/', SpaView.as_view(), name='add_queue'),
     path('auth/<backend_name>/', AuthPromptView.as_view(), name='auth_prompt'),
     path('callback/<backend_name>/', auth_callback_view, name='auth_callback'),
+    path("robots.txt", SpaView.as_view(template_name="robots.txt", content_type="text/plain")),
```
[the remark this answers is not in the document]
jonespm: `robots.txt` could also probably also be defined in views.py without a separate file. If you didn't want to do that I'd _probably_ use the TemplateView as SpaView is just a special view to redirect to the SPA front page. 

```
def robots_txt(request):
    content = "User-agent: *\nDisallow: /"
    return HttpResponse(content, content_type="text/plain")
```

And then referenced directly.
`    path('robots.txt', views.robots_txt, name='robots_txt'),`

src/officehours_ui/templates/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-Agent: *
+Disallow: /
```
jonespm, line 2: @ssciolla Do you think we should have a Disallow here? I guess this site really has nothing to index anyway and it's not something for the public, so seems like it would be fine, but this is a new file.

src/officehours_ui/urls.py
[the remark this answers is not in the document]
ssciolla: That kind of view would work, or I think I'd just serve a static file using `RedirectView` like `favicon.ico`. It seems like since `robots.txt` implies it's a file, it's a bit more straightforward to provide one.

src/officehours_ui/static/css/main.css
jonespm: I looked into this and added some styles for firefox. But I've been testing and it doesn't 100% seem to work so I'm not sure.

I think the Firefox usage on mobile is going to be very low and on a desktop this really isn't a problem. So I'm inclined to leave it as it is. It seems to look good on mobile Chrome.

src/officehours_ui/urls.py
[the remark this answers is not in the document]
jonespm: Yeah, that seems like a good idea too. Looks to work.

jonespm:
I believe I've resolved the issues on here and this PR is all good for review. I tested it out and it looks good. 

Are you good with the image that was selected here? We could probably downscale it. I'm not sure the source of this.


We could also use the more generic Michigan one that ITS uses.

jonespm:
Okay, I committed this to the source

src/officehours_ui/tests.py
```diff
@@ -0,0 +1,12 @@
...
+class UITestCase(TestCase):
+    def test_robots_txt(self):
+        c = Client()
+        response = c.get('/robots.txt')
+        self.assertEqual(response.status_code, 200)
+
+    def test_favicon_ico(self):
+        c = Client()
+        response = c.get('/static/favicon.ico')
+        self.assertEqual(response.status_code, 200)
```
ssciolla, line 12: I think these tests need to be modified. The first one is failing because of the redirect (getting status code 301). The other works, but it's not testing the root, which is I think where we typically see requests to. I guess both should be change to hit the root plus the filename, and they should expect 301?

src/officehours_ui/static/css/main.css
```diff
@@ -554,6 +553,35 @@ h3.alert-heading {
...
+    scrollbar-color: rgba(0, 0, 0, .5) white;
+}
+
+/* Firefox */
+@-moz-document url-prefix() {
+    .table-responsive {
+        scrollbar-width: thin;
+        scrollbar-color: rgba(0, 0, 0, .5) white;
+    }
+}
```
ssciolla, line 568: Do you get this to work in Firefox? I didn't seem to, even if I mess around with the system settings on my Mac. Also, it looks like this is double applying styles, are both necessary?

I guess my first thought would be to not try to use these if we can't get them to work. Ideally we would, but I'd rather not commit something that is potentially not helping.

ssciolla (commented):
@jonespm, I had a couple thoughts here. Can we at least fix the test cases?

Thanks for changing to the `RedirectView`. I wasn't necessarily thinking you'd try to resolve #298, but getting in the Safari/Chrome fix would be some improvement. Let me know what your current thinking is about Firefox/why you added what you did.

src/officehours_ui/static/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-Agent: *
```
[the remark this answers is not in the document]
jonespm: Yeah, I've seen it both ways as well as all lower case. It's not case sensitive. Can commit this though. Seems a little nit-picky.

src/officehours_ui/tests.py
jonespm: Okay I didn't know about the tests, I made the updates to have them both get a 301. It looks like this is the correct return.

src/officehours_ui/static/css/main.css
jonespm: No, I couldn't get it to work either, I can remove the classes for Firefox that I couldn't get to work.

src/officehours_ui/static/css/main.css
```diff
@@ -554,6 +553,27 @@ h3.alert-heading {
 	color: #cecbc9;
 }
 
+
+.table-responsive {
+    scrollbar-width: thin;
+    scrollbar-color: rgba(0, 0, 0, .5) white;
+}
```
ssciolla, line 560: Looks like this does get applied in Firefox (I see the styles with the inspector, and checking boxes). It just doesn't make it show all the time. I couldn't figure out how to do this, even adjusting the OS settings. Who knows!

ssciolla (approved):
Looks good! One CSS simplification (removal) we might make; feel free to do that if you think it's worth it.

src/officehours_ui/static/css/main.css
```diff
@@ -554,6 +553,27 @@ h3.alert-heading {
...
+.table-responsive::-webkit-scrollbar-thumb {
+    border-radius: 8px;
+    border: 2px solid white;
+    background-color: rgba(0, 0, 0, .5);
+}
+.table-responsive::-webkit-scrollbar-track { 
+    background-color: #fff; 
+    border-radius: 8px; 
+}
```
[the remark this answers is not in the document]
jonespm: Okay!

PR #403, opened 2023-04-11
Fixes #336, originally #396  Allow admins to edit users who haven't authorized any backends

### LINKED ISSUE #336: Allow admins to edit users who haven't authorized any backends

Tombow1:
Hi @ssciolla, thanks for the comments and I have generated a migration for my changes and removed the unwanted part. This is a new PR for #396 since it was originally made on the master repo.


src/officehours_api/migrations/0023_auto_20230411_0302.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.2.18 on 2023-04-11 07:02
+
+import django.contrib.postgres.fields
+from django.db import migrations, models
+import jsonfield.fields
+import officehours_api.models
```
ssciolla, line 6: Can you remove any imports that are not being used?

ssciolla (commented):
Thanks for opening the new PR and making the changes. I'll do some more testing on this today. Once it's approved, I think we can probably merge this before #383, and I can update that PR.

jonespm (approved):
This looks good to me

src/officehours_api/migrations/0023_auto_20230411_0302.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.18 on 2023-04-11 07:02
+from django.db import migrations, models
```
ssciolla, line 2:
```suggestion
from django.db import migrations
```

ssciolla (approved):
Aside from that last unused import, this looks good! If you could remove that (which can be done in the GitHub UI), that would be great.

PR #406, opened 2023-04-21
Add way to verify phone numbers

alielwart:
Adding way to verify phone numbers from Twilio. This included a file called: test.py which includes the plain implementation and shows how this verification works. For future it can be easily added that Twilio sends an SMS notification to verify with the user.

### DIFF
--- a/src/officehours_api/views.py
+++ b/src/officehours_api/views.py
@@ -92,7 +92,7 @@ class UserDetail(DecoupledContextMixin, LoggingMixin, generics.RetrieveUpdateAPI
 
 
     def update(self, request, *args, **kwargs):
-        user = self.get_object
+        user = self.get_object()
         self.check_change_permission(request, user)
         self.verify_phone(request, user)
         return super().update(request, *args, **kwargs)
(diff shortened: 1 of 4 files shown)


src/officehours_api/views.py
```diff
@@ -73,9 +77,24 @@ def check_change_permission(self, request, user):
         if user != request.user:
             self.permission_denied(request)
 
+    def verify_phone(self, request, user):
+         client = Client()
+         if user != request.user:
+             if User.phone_number != None:
```
ssciolla, line 83: I'm not sure I understand these checks yet. For `user != request.user`, are we just checking to see if the user has been changed? I think we probably want to check specifically for if the phone number has changed and then run the verification. `User.phone_number` puzzles me, cause `User` is the model, not the model instance. We probably do need a check like this to make sure we're not passing `None` into this verification process.

ssciolla (commented):
Thanks for the contribution! This looks like a very good start. I'm planning to do some additional testing to check my understanding of your work, but I'm releasing some initial thoughts/questions.

I know the semester is now over, so I'm not sure how much you'll want to or be able to finish this work. If you could at least let me know if you'll have time, that would be helpful so we can think about how to get this completed.

jonespm:
Closing as this was incomplete and continued in #474

Issue #357, opened 2022-05-06
"Exception: Can't change assignee once meeting is started!" message should be shown to host


ssciolla:
When one host goes to assign themselves as a host when another host has simultaneously or previously created a meeting, an error is thrown in the backend: 
```
Exception: Can't change assignee once meeting is started!
```
However, the frontend only shows the user "Attendee: Internal Server Error". It's apparent that what has happened because the user would have seen the meeting appear under "Meetings in Progress", but it would be better if a message like the one above appeared in the error alert on the page as well.

![image](https://user-images.githubusercontent.com/27447/205161890-037679f3-f0a6-46e6-a65a-2dc3cb5c5656.png)

ssciolla:
@zqian, best way to test this is with two people, so let me know and I can assist. Basically you want to reproduce these situations:
1) One host is trying to re-assign a meeting to a different host right after the meeting gets started by another host.
2) Attendee tries to change the backend type right after a host starts the meeting. This case may be hard or impossible to produce via the UI, since I believe it blocks you from changing the backend type of your meeting once a host is assigned. It is possible to do this via the API though, which we can test with Swagger locally.

zqian:
Tested on QA was success for the first situation. See the screenshot below: 
![image](https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/359944/7b724880-8d25-43ab-ab81-b732c2e429d6)

PR #407, opened 2023-04-22
added try catch and test case to fix issue 357

Garima-Shah:
My partner @challasr and I have took a look at Issue 357. We saw the previous student had tried to add a try and catch block in a different function then wanted so we edited the try catch block to be on lines 176-180


ssciolla:
Thanks for your work on this. We ended up fixing this in a separate PR here, in case you are interested: #357. What you did looks like it's on the right track though. The test needs a bit more I think in terms of assertions and tidying up.

Issue #330, opened 2021-05-05
Menu icon for small screens


teshsopher:
Related: #327 

On small screens, the menu collapses into a "hamburger" or "stacked" menu. While the menu is open/expanded, this icon change to an "X" or other icon to indicate that the menu can be closed.

This is mostly a "nice-to-have" change that could help with some minor confusion for users, but it is not a strictly necessary one.

zshahidumich:
Hey there! Nice to meet you all! My name is Zaid and I'm part of EECS 481 Software Engineering. I hope you're all doing well. I just wanted to drop a quick message here to make sure I've covered all my bases. I've already sent an email and joined the Google group, but I was wondering if there's a chance I could be assigned to this task. I am also dropping a message on issue #207. Thanks in advance!

HemanginiShah:
I checked this in current **QA** and  **prod** environment.  When I clicked on hamburger icon.  Not able to recreate the 
**Before clicking on hamburger icon:**

![ROHQ hamburger icon](https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/29413657/8277608f-6025-46e0-9d75-30cc2d8a19f7)

**After clicking on Hamburger icon:**
![ROHQ after clicking on hamburger icon](https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/29413657/9bd2d0bb-f15b-4582-a6fe-e04a572d48d3)

HemanginiShah:
Sam Thanks for clarification.  Yes, menu hamburger icon works as intended to be used.  Passes usability test as well. (right click inspect when Manage queues are listed, click on the icon from the right window and then tab through each functions on the screen.
Test- Pass

PR #408, opened 2023-04-22
Issue #330 Fixed

zshahidumich:
I added a javascript script at the end of the base.html file. This event script gets triggered when the header button is pushed causing the icon to switch to an 'X'.


jonespm:
This will be resolved by #375 so closing this commit. Thanks for the contribution.

This looks like it would have fixed it similarly but there was some unnecessary files as part of this PR committed so I just went with updating the other one.

PR #412, opened 2023-05-31
Update Google Analytics library and usage (#392)

### LINKED ISSUE #392: Migrate to Google Analytics 4
We'll need to move off Universal Analytics to Google Analytics 4 this year. We can use this library as a replacement for `react-ga` which doesn't yet support GA 4: https://www.npmjs.com/package/react-ga4

ssciolla:
This PR aims to resolve #392 and #414.

Testing notes:

Make sure "pageview" events are sent when the URL, even within the SPA.

To trigger custom events:

Host
- [x] Create a queue
-  [x] Add a host to a queue
- [x] Add a meeting on management page
- [x] Remove a meeting on management page
- [x] Start a meeting on management page
- [x] Close and/or open a queue from management page
- [x] Change an assignee on the management page
- [x] Remove a host from the settings page
- [x] Update the queue name, description, in-person meeting location, or meeting types from settings page
- [x] Delete a queue from settings page

Attendee
- [x] Join a queue
- [x] Leave a queue
- [x] Join a different queue when already in line in another queue

### DIFF
--- a/src/assets/src/hooks/useGoogleAnalytics.ts
+++ b/src/assets/src/hooks/useGoogleAnalytics.ts
@@ -1,7 +1,9 @@
 import { useState, useEffect } from 'react';
 import GoogleAnalytics from 'react-ga4';
+import { useLocation } from 'react-router-dom';
 
 export const useGoogleAnalytics = (googleAnalyticsId?: string, debug?: boolean) => {
+    let location = useLocation();
     const [initialized, setInitialized] = useState(false);
     const [previousPage, setPreviousPage] = useState(null as string | null);
 
@@ -11,10 +13,10 @@ export const useGoogleAnalytics = (googleAnalyticsId?: string, debug?: boolean) 
     }
 
     useEffect(() => {
-        const page = window.location.pathname + window.location.search + window.location.hash
+        const page = location.pathname + location.search + location.hash;
         if (googleAnalyticsId && page !== previousPage) {
             setPreviousPage(page);
             GoogleAnalytics.send({ hitType: "pageview", page });
         }
-    });
+    }, [location]);
 }
--- a/src/assets/src/hooks/useGoogleAnalytics.ts
+++ b/src/assets/src/hooks/useGoogleAnalytics.ts
@@ -1,5 +1,5 @@
-import { useState, useEffect } from 'react'
-import GoogleAnalytics from 'react-ga'
+import { useState, useEffect } from 'react';
+import GoogleAnalytics from 'react-ga4';
 
 export const useGoogleAnalytics = (googleAnalyticsId?: string, debug?: boolean) => {
     const [initialized, setInitialized] = useState(false);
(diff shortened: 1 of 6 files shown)


pushyamig:
can you provide me the GA-id configuration field used in the `.env` since this is not written up in READme. 
Also can you provide me the id as well inorder for testing? does it matter? since I can use GA browser plug-in to test without GA-ID

pushyamig:
The `.env` uses GA_TRACKING_ID always, this should be documented in the Readme.

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,20 +1,22 @@
...
 
 export const useGoogleAnalytics = (googleAnalyticsId?: string, debug?: boolean) => {
+    let location = useLocation();
     const [initialized, setInitialized] = useState(false);
     const [previousPage, setPreviousPage] = useState(null as string | null);
 
     if (googleAnalyticsId && !initialized) {
         setInitialized(true);
-        GoogleAnalytics.initialize(googleAnalyticsId, { debug });
+        GoogleAnalytics.initialize(googleAnalyticsId, { testMode: debug });
```
pushyamig, line 12: the `debug` is currently tied up so Django debug setting, which means when Django.debug = true then GA tracking won't happen, This is doesn't effect Prod. But when testing Non-prod the debug option will be turned on in that case this could cause for confusion if particularly testing for GA events. This should be documented along with GA configuration options

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,20 +1,22 @@
...
     useEffect(() => {
-        const page = window.location.pathname + window.location.search + window.location.hash
+        const page = location.pathname + location.search + location.hash;
         if (googleAnalyticsId && page !== previousPage) {
             setPreviousPage(page);
-            GoogleAnalytics.pageview(page);
+            GoogleAnalytics.send({ hitType: "pageview", page });
         }
-    });
+    }, [location]);
```
pushyamig, line 21: Nice fix for #414

pushyamig (approved):
This PR is functional, I was able verify that events are logged using real time GA dashboard. I did not exhaust the testing. But based on code changes these seems to work

src/assets/src/hooks/useGoogleAnalytics.ts
ssciolla: It actually looks like we use `DEBUG` with `False` in all our deployments, so it's probably simplest to test in one of our non-prod environments. I noted this in the addition to the `README` I made: https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/412/commits/fbbc22f94db3d9302367a91f11be15afed1bfa5e

It's not a bad idea to have some other kind of flag like `GA_DEBUG` so we can more effectively do local testing, but I think this isn't something we're going to have worry about often.

ssciolla:
Thanks for the review! I'm going to merge, and I'll work on preparing this for QA next week.

HemanginiShah:
6/14 - I have followed all required event steps in QA but not seeing data updates for the Eventname or Users by User Property.

PR #415, opened 2023-05-31
Remove `console.warn` from `validateString` (#413)

### LINKED ISSUE #413: Remove warning from `validateString`
https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/9a183fb140406f9c2c0dddd82b67272922126814/src/assets/src/validation.ts#L89-L91

I thought this was a good addition in #405, but I forgot that this would trigger every time if we don't pass `showRemaining` as `true`, this would always be triggered. This happens with uniqname entry (we don't show remaining there, but we do have a `maxLimit`). We could add a separate check to see if both `showRemaining` is `true` and `maxLimit` is `undefined`, but we could also just remove this message since we control usage of this function and we don't use it this way.

ssciolla:
This PR aims to resolve #413.

### DIFF
--- a/src/assets/src/validation.ts
+++ b/src/assets/src/validation.ts
@@ -86,8 +86,6 @@ export function validateString (value: string, schema: StringSchema<string>, sho
         const maxLimit = getMaxLimit(schema.describe());
         if (showRemaining && maxLimit) {
             messages.push(createRemainingCharsMessage({value: transformedValue, max: maxLimit}));
-        } else {
-            console.warn('validateString called with showReamining but schema does not include a maxiumum.');
         }
     } catch (error) {
         if (!ValidationError.isError(error)) {


src/assets/src/validation.ts
```diff
@@ -86,8 +86,6 @@ export function validateString (value: string, schema: StringSchema<string>, sho
         const maxLimit = getMaxLimit(schema.describe());
         if (showRemaining && maxLimit) {
             messages.push(createRemainingCharsMessage({value: transformedValue, max: maxLimit}));
-        } else {
-            console.warn('validateString called with showReamining but schema does not include a maxiumum.');
```
pushyamig, line 90: What is the user see from the home screen/UI when this scenario occurs?

ssciolla: Nothing, this just shows up a bunch in the console log.

ssciolla: If `showRemaining` is `false` (like with the uniqname input on the queue management screen), then there just isn't a remaining message, like there is for the larger fields.

We could add a separate check for the scenario where `showRemaining` is `true` and `maxLimit` is `undefined`, but it doesn't seem super important. See the issue for more details.

pushyamig: I did test his use case last time, as you said user might not care unless the UI doesn't notify him in case of error. Which seems to be not the case here.

pushyamig (approved):
LGTM!

PR #421, opened 2023-06-13
Fixes #420 - Remove args and command from service so it will use the Dockerfile values

### LINKED ISSUE #420: Update default command in base to match what's in the Dockerfile
This application uses Gunicorn rather than runserver now, so the web-deployment should be updated to reflect this. 

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/57ca6144d9cfcd6d0941090b026f4dc398d3b014/service/base/web-deployment.yaml#L27-L28

jonespm:


jonespm:
@ssciolla Okay I modified this to remove the base and overlays.

ssciolla:
Thanks, will review in a minute.

ssciolla (approved):
LGTM!

Issue #422, opened 2023-06-14
Handle string in URL parameters that pass through React Router


ssciolla:
This is related to but a bit beyond #236. 

Currently, if the user enters a string in the URL parameter for queue ID, the tool will throw an error in the backend, a Queue Connection will appear in the UI, and `NaN` will appear in the breadcrumbs. This likely also affects `manage` URLs, so the solution should be reusable if possible.

After checking with the service owner, here are the suggested/desired changes:

- Change the breadcrumb text to display the incorrect string or integer value entered, not the `NaN`
- Change the error message to `Oops! It looks like you've entered an invalid queue ID <copy entered str value from URL>in the URL. Please make sure to enter the correct numeric queue ID. You can find the accurate queue ID in your 'Manage Queue' list for reference.`
- Sample changes apply to the `manage` URLs

ssciolla:
We should also look at whether this can be enforced more cleanly in the backend.

Nyu10:
I would like to tackle this issue

jaydonkrooss:
This fix is properly handled in QA environment. moving to completed.

PR #423, opened 2023-06-14
Remove alert and unused utility function (#236)

### LINKED ISSUE #236: Remove notification for uniqname queue redirect
image
We added the above yellow box as part of the queue redirect for the April 8 release that changed from /queue/uniqname to /queue/id. It has been long enough that I do not think we need this notification anymore. Let's keep the functionality that redirects to a search, but remove the notification box. That way I can delete the documentation article that it links to. https://documentation.its.umich.edu/office-hours-links

ssciolla:
This PR aims to resolve #236. Note that it does not address the newly created issue #422.

Background

My understanding is that it used to be possible to access a queue associated with a uniqname via `/queue/{uniqname}/`. When this was removed, they supported old links by redirecting users to the search and conditionally showing an alert explaining. This is no longer the case and it's been long enough that it can be safely removed.

How/what to test

- [x] Removing `redirectToSearch` doesn't affect anything.
- [x] Navigated to `/search/?redirected=true&term=something` does not show an alert.

### DIFF
--- a/src/assets/src/utils.ts
+++ b/src/assets/src/utils.ts
@@ -11,15 +11,6 @@ export const redirectToLogin = (loginUrl: string) => {
     location.href = loginUrl + '?next=' + location.pathname;
 }
 
-export const redirectToSearch = (term: string) => {
-    ReactGA.event({
-        category: "Navigation",
-        action: "Redirected to Search",
-        nonInteraction: true,
-    });
-    location.href = `/search/${term}/?redirected=true`;
-}
-
 export const redirectToBackendAuth = (backend: string) => {
     ReactGA.event({
         category: "Auth",
--- a/src/assets/src/components/search.tsx
+++ b/src/assets/src/components/search.tsx
@@ -44,18 +44,11 @@ export function SearchPage(props: PageProps) {
                 </p>
             )
             : <QueueTable queues={searchResults} />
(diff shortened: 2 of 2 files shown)


pushyamig:
I will review this tomorrow

src/assets/src/utils.ts
```diff
@@ -11,15 +11,6 @@ export const redirectToLogin = (loginUrl: string) => {
...
 }
 
-export const redirectToSearch = (term: string) => {
-    ReactGA.event({
-        category: "Navigation",
-        action: "Redirected to Search",
-        nonInteraction: true,
-    });
-    location.href = `/search/${term}/?redirected=true`;
-}
```
pushyamig, line 21: Is this code just removing Google Analytics event?

pushyamig: shouldn't this event be tracked? Since a user did a search and got redirect. or GA event is not tracked at all with this code?

pushyamig (approved):

HemanginiShah:
During QA test:
https://qa.officehours.it.umich.edu/search/?redirected=true&term=hemashah  (This link displayed following screenshot)
image

When trying this link: https://qa.officehours.it.umich.edu/queue/hemashah/  (saw the alert on the screen as attached here)
image

PR #424, opened 2023-06-14
Return HTTP status code and message when trying to change a started meeting (#357)

### LINKED ISSUE #357: "Exception: Can't change assignee once meeting is started!" message should be shown to host

ssciolla:
This PR aims to resolve #357. It also handles a possible similar case where a user tries to change the `backend_type` after the meeting is started. There are tests for both.

Credit to @anishsundaram and Divyanga Bajoria for getting this work started. This PR will supersede #376, and if this is merged, that will be closed.

### DIFF
--- a/src/officehours_api/views.py
+++ b/src/officehours_api/views.py
@@ -186,6 +186,12 @@ class MeetingDetail(DecoupledContextMixin, LoggingMixin, generics.RetrieveUpdate
     serializer_class = MeetingSerializer
     permission_classes = (IsAuthenticated, IsHostOrAttendee,)
 
+    def update(self, request, *args, **kwargs):
+        try:
+            super().update(request, *args, **kwargs)
+        except Exception as e:
+            return Response({'Meeting Detail': e.args[0]}, status=status.HTTP_400_BAD_REQUEST)
+
 
 class MeetingStart(DecoupledContextMixin, LoggingMixin, APIView):
     permission_classes = (IsAuthenticated, IsAssignee,)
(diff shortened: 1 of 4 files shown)


src/officehours_api/views.py
```diff
@@ -186,6 +186,12 @@ class MeetingDetail(DecoupledContextMixin, LoggingMixin, generics.RetrieveUpdate
     serializer_class = MeetingSerializer
     permission_classes = (IsAuthenticated, IsHostOrAttendee,)
 
+    def update(self, request, *args, **kwargs):
+        try:
+            super().update(request, *args, **kwargs)
+        except MeetingStartedException as e:
+            return Response({'Meeting Detail': e.message}, status=status.HTTP_400_BAD_REQUEST)
```
jonespm, line 193: This seems fine. Maybe a nitpick but I wonder if 409 would be the most appropriate response for this error instead of a generic 400?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

jonespm (approved):

src/officehours_api/views.py
ssciolla: Hm, maybe, likely we'd have to change this in other places as well. I'm going to keep it as is for now.

PR #426, opened 2023-06-14
Fix Manage Hosts spacing, remove unused `pull-left` class (#425)

### LINKED ISSUE #425: Remove icon not correctly placed in list of queue's hosts
I noticed a visual bug that fell through the cracks during #405. For any host list items past the first one in the Manage Hosts tab in queue settings, the host's name and the remove icon will be jumbled together, instead of having the remove icon floating to the right as in production. This is because in Bootstrap 5, utilities changed to using `start` and `end` over `left` and `right.

Screenshot 2023-06-14 at 3 38 48 PM

I should also do a quick search for `-left` and `-right` to make sure I didn't miss any other instances of this.

ssciolla:
This PR aims to resolve #425.

I reviewed all the instances of `class` in the templates and `className` in the React application. I only found one other instances with `pull-left` on the navigation bar image, which I removed. This seems to have been legacy addition from before my time even, since `pull-` is from from Bootstrap 3 (see https://getbootstrap.com/docs/4.6/migration/#utilities). Removing the class resulted in no changes.

**Note:** For local testing, you'll need to create a user via the Django admin. I usually just create one called `testuser`. Then I go to a queue, and under Settings > Manage Hosts, add the extra user.

### DIFF
--- a/src/officehours_ui/templates/base.html
+++ b/src/officehours_ui/templates/base.html
@@ -30,7 +30,7 @@
 			<div class="container">
 				<div class="navbar-brand">
           <a href="{% url 'home' %}">
-            <img src="{% static 'images/its-logo.png' %}" class="container banner-img img-responsive pull-left" alt="Remote Office Hours Queue"/>
+            <img src="{% static 'images/its-logo.png' %}" class="container banner-img img-responsive" alt="Remote Office Hours Queue"/>
           </a>
 				</div>
 				<button class="navbar-toggler ml-auto collapsed" type="button" data-bs-toggle="collapse" data-bs-target=".navbars" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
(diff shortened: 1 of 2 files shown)


src/officehours_ui/templates/base.html
```diff
@@ -30,7 +30,7 @@
 			<div class="container">
 				<div class="navbar-brand">
           <a href="{% url 'home' %}">
-            <img src="{% static 'images/its-logo.png' %}" class="container banner-img img-responsive pull-left" alt="Remote Office Hours Queue"/>
+            <img src="{% static 'images/its-logo.png' %}" class="container banner-img img-responsive" alt="Remote Office Hours Queue"/>
           </a>
```
pushyamig, line 34: Yeah.. this have no effect on this change when removed

pushyamig (approved):
This change look good and nice catch and good fix

ssciolla:
Thanks for the review, @pushyamig!

PR #430, opened 2023-06-15
Add --save "" to redis-server command (#429)

### LINKED ISSUE #429: Suppress `redis` log messages about failed writes to disk
After upgrading to `redis:7`, we are seeing some additional logging messages in the `redis` service container on OpenShift like this:

```
Background saving started by pid 477
Failed opening the temp RDB file temp-477.rdb (in server root dir /data) for saving: Permission denied
```
This is almost certainly because we have not given the user permission to write to `/data` or mounted a volume for the writing to occur. The deployment YAML was already specifying an option to disable "stop-writes-on-bgsave-error", which seems to indicate that writes (to disk) were not working before either.

After discussing with @jonespm, it seems like it would be safe to disable writing entirely using the `save` setting with an empty string. Or we could add a volume, but I think the main purpose of the background saving would be to recover important data. My understanding of channels and its usage of redis is that any connection data would just be reestablished if it was lost in a disaster.

Resources
- https://redis.io/docs/management/persistence/
- Default `redis-conf`: https://raw.githubusercontent.com/redis/redis/7.0/redis.conf

ssciolla:
This PR aims to resolve #429.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,7 @@ services:
 
   redis:
     image: redis:7
+    command: redis-server --stop-writes-on-bgsave-error no --save ""
 
   webpack_watcher:
     build:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,12 @@ services:
   redis:
     image: redis:7
     command: redis-server --stop-writes-on-bgsave-error no --save ""
+    # This is to simulate an Openshift container with no write permission except to /tmp
+    read_only: true
+    tmpfs:
+        - /tmp
+    volumes:
+        - officehours-redis-ro:/data:ro
 
   webpack_watcher:
     build:
@@ -45,3 +51,4 @@ services:
 
 volumes:
   officehours-postgres-data:
+  officehours-redis-ro:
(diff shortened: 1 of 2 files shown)


jonespm (approved):
Okay this seems good. It's hard to test locally since the default is "3600 1" but according to the documentation this stops saves entirely. I don't think the other option would do anything anymore but still seems less risk to leave it in. 

```
# Snapshotting can be completely disabled with a single empty string argument
# as in following example:
#
# save ""
```

https://github.com/redis/redis/blob/5da9eecdb8b53aa214e9e9ac1b977252360ac5ee/redis.conf#L420

I  tested locally that the error comes up with this change the the docker-compose.yml. It might be worth adding this read only file system to the local config to more closely match prod.

I'm good with rolling this out and testing but looks like it works good. 

```
diff --git a/docker-compose.yml b/docker-compose.yml
index 1a083f4..0ec0d7c 100755
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,9 @@ services:

   redis:
     image: redis:7
+    command: redis-server --stop-writes-on-bgsave-error no --save "30 1"
+    volumes:
+      - officehours-redis-ro:/data:ro

   webpack_watcher:
     build:
@@ -44,3 +47,4 @@ services:

 volumes:
   officehours-postgres-data:
+  officehours-redis-ro:
```

ssciolla:
Thanks for the review, @jonespm, and contributing the extra `docker-compose.yml` configuration to make the `redis` container read-only to align it with our deployments.

PR #433, opened 2023-06-16
Return response in overridden update method (#432)

### LINKED ISSUE #432: Meeting update throwing Internal Server Error
https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/e0c8276f2f323bbc16888af094100f77f4bda7c1/src/officehours_api/views.py#L189-L193

It looks like I neglected to return the response when calling update in the `try` clause. This is causing an internal server error to be thrown whenever a meeting is patched (update) by the attendee or host.

ssciolla:
This PR aims to resolve #432.

This is something I overlooked in #424. I've added a success test case that will fail without my fix.


jonespm (approved):
Missed this return too, thanks!

PR #435, opened 2023-06-19
Fix four UI issues (#434)

### LINKED ISSUE #434: Fix UI issues
While preparing for our 1.8.0 release, I identified four user interface issues that pose either usability or accessibility concerns. I'll detail each of them below. The first three are regressions that occurred as part of the changes in #405; the last was an existing issue.

1) Edit button in agenda does not have visible focus.

https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/35741256/aed68ec8-7831-4352-9a92-c8d3773893c5

2) Required symbol is not next to Select Meeting Type in the dialog from the attendee side.

The dropdown also goes all the way across when it did not before.

Screenshot 2023-06-19 at 11 25 07 AM

3) Assignee dropdown in the Meetings in Queue table does not have a dropdown arrow.

Screenshot 2023-06-19 at 12 38 28 PM

4) Cancel button in Add Queue workflow has two "stops" in the keyboard Tab order.

https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/35741256/ae6d4497-e750-4c9b-862f-3b578dfbcc95

ssciolla:
This PR aims to resolve #434. See issue for details on the regressions or problems this resolves.

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -37,7 +37,7 @@ const JoinQueue: React.FC<JoinQueueProps> = (props) => {
         <>
         <Row>
             <Col lg>
-                <p className="mb-0">Select Meeting Type<span className="required">*</span></p>
+                <p className="mb-0">Select Meeting Type <span className="required">*</span></p>
             </Col>
         </Row>
         <Row>
@@ -412,14 +412,21 @@ const ChangeMeetingTypeDialog = (props: ChangeMeetingTypeDialogProps) => {
                 <Modal.Title>Change Meeting Type</Modal.Title>
             </Modal.Header>
             <Modal.Body>
-                <div className="row col-lg">
-                    <p>Select Meeting Type</p>
-                    <p className="required">*</p>
-                </div>
-                <BackendSelector backends={props.backends}
-                    allowedBackends={new Set(props.queue.allowed_backends)}
-                    onChange={props.onChangeBackend}
-                    selectedBackend={props.selectedBackend}/>
+                <Row>
+                    <Col lg>
+                        <p>Select Meeting Type <span className="required">*</span></p>
+                    </Col>
+                </Row>
+                <Row>
+                    <Col xs='auto'>
+                        <BackendSelector
+                            backends={props.backends}
+                            allowedBackends={new Set(props.queue.allowed_backends)}
+                            onChange={props.onChangeBackend}
+                            selectedBackend={props.selectedBackend}
+                        />
+                    </Col>
+                </Row>
             </Modal.Body>
             <Modal.Footer>
                 <Button variant="secondary" onClick={props.onClose}>Cancel</Button>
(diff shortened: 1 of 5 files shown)


pushyamig:
I am review now

src/assets/src/components/queue.tsx
```diff
@@ -37,7 +37,7 @@ const JoinQueue: React.FC<JoinQueueProps> = (props) => {
         <>
         <Row>
             <Col lg>
-                <p className="mb-0">Select Meeting Type<span className="required">*</span></p>
+                <p className="mb-0">Select Meeting Type <span className="required">*</span></p>
```
jonespm, line 40: I know this is just adding a space here but it looks like using the HTML form validation is recommended over an * both for usage and accessibility. 

https://www.accessibility-developer-guide.com/examples/forms/required/#using-html-5-client-side-validations
https://react-bootstrap.netlify.app/docs/forms/validation

Was this not changed just because that's a separate (more effort) issue? We should probably file it as a new issue to fix it in the future.

src/assets/src/components/queue.tsx
```diff
@@ -412,14 +412,21 @@ const ChangeMeetingTypeDialog = (props: ChangeMeetingTypeDialogProps) => {
...
             <Modal.Body>
-                <div className="row col-lg">
-                    <p>Select Meeting Type</p>
-                    <p className="required">*</p>
-                </div>
-                <BackendSelector backends={props.backends}
-                    allowedBackends={new Set(props.queue.allowed_backends)}
-                    onChange={props.onChangeBackend}
-                    selectedBackend={props.selectedBackend}/>
+                <Row>
```
jonespm, line 415: Do you think the col-lg on line 372 might also be an issue and need to be updated?

src/assets/src/components/queue.tsx
ssciolla: I think there are some interesting ideas in this article. I'm not exactly sure how it interacts with dropdowns, or this particular case. Maybe the required indicator isn't the best here anyway. I'm going to open this as a separate issue to reconsider, since it would affect other areas of the application.

ssciolla: See #437.

src/assets/src/components/queue.tsx
ssciolla: It's not an issue per se, but it's an instance of the direct Bootstrap class usage I've tried to move away from. Actually, this column turned out to be not needed, since it's not part of a larger grid structure (other `Row`s or `Col`s). Since this was the last instance of direct Bootstrap class usage in the module, I went ahead and removed it and used an `Alert` component there. It does have one side effect of setting `role="alert"`, which makes it pretty noisy for screenreaders, but we're using `Alert` for other usages on the page. I could see setting `role={undefined}` for the SMS and management alerts, but they are conditional as well. Going to leave it like this for now.

PR #443, opened 2023-07-03
Update `kustomize` process, migrate to ROSA (#441)

### LINKED ISSUE #441: Update `service` artifacts to support latest version of `kustomize`, ROSA
We will need to update the YAML files and other artifacts under `service` to use the latest version of `kustomize` and to be compatible with ROSA. The ingress configuration is an area that has been flagged and will need to be investigated.

ssciolla:
This PR aims to resolve #441. Changes to kustomize files were mostly arrived at by 1) running the CLI tool referenced in errors and 2) reviewing errors and then looking at the documentation for that resource type. The changes also make use of the new cert-manager functionality available in the new cluster for auto-updating and application of certificates.

Resources
- https://kubectl.docs.kubernetes.io/guides/

### DIFF
--- a/service/base/kustomization.yaml
+++ b/service/base/kustomization.yaml
@@ -41,11 +41,6 @@ secretGenerator:
   name: secrets
   type: Opaque
 - files:
-  - secret/tls.key
-  - secret/tls.crt
-  name: web-tls
-  type: kubernetes.io/tls
-- files:
   - secret/WebHookSecretKey
   name: github
   type: Opaque
--- a/service/base/kustomization.yaml
+++ b/service/base/kustomization.yaml
@@ -1,7 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-commonLabels:
-  org: umich
 resources:
 - redis-deployment.yaml
 - redis-service.yaml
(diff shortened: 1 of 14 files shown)


service/base/kustomization.yaml
```diff
@@ -38,16 +38,12 @@ secretGenerator:
...
+- files:
   - secret/WebHookSecretKey
-crds:  # Use CRDs to support OpenShift resource kinds
+  name: github
+  type: Opaque
+crds:
 - openshift-buildconfig.json
-- openshift-deploymentconfig.json
\ No newline at end of file
+- openshift-deploymentconfig.json
```
ssciolla: Still not a 100 percent certain how the crds work or if they're still needed...

jonespm:
Great this looks good. Do we still need the tls.yaml file(s) anymore for qa and prod? It doesn't look like they used and the route overrides are just happening in hostname.yaml. Seems like it's good for dev though.

jonespm (approved):

jonespm:
Okay, I see, thanks! I don't see anything else, can give it a run on QA.

lsloan:
QA OK.
:shipit:

I've been able to run `kustomize` locally to produce config files.  The `kustomize` process was also used by @jonespm to deploy the updated app to the test environment.

I noticed some small things that needed to be changed and information added in `service/README.md`.  It doesn't need to be completed for this release (1.9.0).  I opened a pull request (#459) about that, but I should probably also create an issue for it, too.

PR #446, opened 2023-07-12
Fix "Return to Previous Queue" button on another queue (#445)

### LINKED ISSUE #445: "Return to Previous Queue" button does not work from another queue
**Expected Behavior**

When currently in line in one queue, if you visit another queue, you should see an alert saying you are already in line and be able to click a "Return to Previous Queue" button to return to the other queue.

**Actual behavior**

Clicking the "Return to Previous Queue" button does not make the page re-load, though it does change the URL. See the video below.

Note that this does not affect the "Return to Previous Queue" button that appears conditionally on the home page.

https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/35741256/8f884461-4ebe-4698-ba5c-3606ee0cafd3

**Commentary**

This behavior seems to have been a side effect of updating React Router Dom and changing from a props approach to a [`useParams`](https://reactrouter.com/en/main/hooks/use-params) approach. I'm not entirely sure why the behavior changed, but I tested with 1.7.0 and the behavior is not present. The solution seems to be modify `useWebSocket` so that the internal `useEffect` retriggers if the URL (using the queue ID) changes. I think it wasn't re-rendering because the `useEffect` has no dependencies and the page component itself isn't being re-loaded (just the param changes). Could be off on that.

ssciolla:
This PR aims to resolve #445.

Resources
- https://react.dev/reference/react/useEffect
- https://reactrouter.com/en/main/hooks/use-params

### DIFF
--- a/src/assets/src/hooks/useWebSocket.ts
+++ b/src/assets/src/hooks/useWebSocket.ts
@@ -67,6 +67,6 @@ export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onD
             ws.onmessage = null;
             ws.close();
         }
-    }, []);
+    }, [url]);
     return error;
 }


zqian (approved):
The fix works as expected.

src/assets/src/hooks/useWebSocket.ts
```diff
@@ -67,6 +67,6 @@ export const useWebSocket = <T>(url: string, onUpdate: (content: T) => void, onD
             ws.onmessage = null;
             ws.close();
         }
-    }, []);
+    }, [url]);
     return error;
```
pushyamig, line 71: I see this is working. 

I see this function is called from 2 places, `useQueueWebSocket` and `useUserWebSocket`, this bug might have originated when updating Queue and not user?

Would it be possible to explain what is these 2 function purpose is?

ssciolla: I didn't update either of the hooks you mentioned in my frontend update (https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/405/files). These hooks are thin wrappers around the hook I'm modifying here, `useWebSocket`, which is like the websocket equivalent of `usePromise`, it manages the state around the lifecycle of a websocket connection. I believe all the wrappers do is form a URL using an ID and pass the needed callback functions.

I think I'm on the right track for explaining the cause of this problem in the issue under Commentary. I don't quite understand why migrating `react-router-dom` caused this, but perhaps it's because of the change from using `props` to a hook (`useParams`). I haven't yet tracked down a GitHub issue or anything like that to explain this.

I think adding `url` here just makes sense. It's possible it could have side effects on other parts of the application, but in general, if the ID (either for a user, or a queue) changes, it makes sense to re-initiate the Websocket connection. By adding `url`, we're subscribing to changes to that URL (and by extension the queue ID or user ID), whereas before we were not).

I looked at all the `useUserWebSocket` usages, and I don't expect the user ID passed to them to change (since it would require a logout, we're only really using this websocket to get the current user's data). And looking at all the `useQueueWebSocket` usages (which on the Settings page, the Manage Queue page, and the Queue visitor page), the only time I think the ID would change without a full re-render is in the case related to this bug here. It's possible if we added more functionality that you'd see this problem with navigating between `/manage/{queueID}` pages or `/manage/{queueUD}/settings` pages, but the fix I'm providing here would resolve those cases too.

Hope that helps.

pushyamig (approved):
Thanks for the Fix, appreciate your sense of responsibility and dedication toward Job you do!

hemashahumich:
Question: Is the red button "Return to Previous Queue" suppose to take you back to the previous queue user has joined?

hemashahumich:
Test Successful in QA.

PR #451, opened 2023-12-01
Handles string in URL parameters

### LINKED ISSUE #422: Handle string in URL parameters that pass through React Router

lhuang408:
### PR Summary

This PR handles invalid queue IDs that are strings. Prior to this fix, if the user enters a string in the URL parameter for queue ID, it will throw an error in the backend, a Queue Connection will appear in the UI, and NaN will appear in the breadcrumbs. With this fix, the error message returned is the same as an invalid queue ID that is a number.

Fixes #422 

### Notes
I set the queue ID that is passed into useQueueWebSocket equal to a large number to obtain the proper error message. This is assuming that the large number is always an invalid queue ID. As such, if future queues are set to that number, the solution will not work as intended.

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -444,7 +444,7 @@ export function QueuePage(props: PageProps) {
     const { queue_id } = useParams();
     if (queue_id === undefined) throw new Error("queue_id is undefined!");
     if (!props.user) throw new Error("user is undefined!");
-    const queueIdParsed = parseInt(queue_id);
+    const queueIdParsed = isNaN(parseInt(queue_id)) ? 10000000000 : parseInt(queue_id);
 
     //Setup basic state
     const [selectedBackend, setSelectedBackend] = useState(undefined as string | undefined);
@@ -577,7 +577,7 @@ export function QueuePage(props: PageProps) {
             <Dialog {...dialogState} />
             <LoginDialog visible={loginDialogVisible} loginUrl={props.loginUrl}/>
             {meetingTypeDialog}
-            <Breadcrumbs currentPageTitle={queue?.name ?? queueIdParsed.toString()}/>
+            <Breadcrumbs currentPageTitle={queue?.name ?? queue_id}/>
             {loadingDisplay}
             {errorDisplay}
             {queueDisplay}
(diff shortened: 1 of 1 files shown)


zqian (commented):
@lhuang408 : The description of #422 suggests to adopt the similar approach to handle invalid queue id, as how it currently handles string values for queue id: 

`If the user enters a string in the URL parameter for queue ID, it will throw an error in the backend, a Queue Connection will appear in the UI, and NaN will appear in the breadcrumbs. ` 

You can test the behavior on master branch with the URL:`http://localhost:8003/queue/a_string_value/`. I attached the screenshot here:

image

Note the `NaN` display in the breadcrumb line. 

However, `NaN` is no longer displayed with your PR on https://localhost:8003/queue/a_string_value/

lhuang408:
Pushed a fix for the breadcrumb line. This is what is now displayed when the a user puts a string in the URL.
image

zqian:
@lhuang408 I have cleaned the description of #422 , to make it clear that the issue is about error-handling of `/queue/<string>` and `/manage/<string>`

However, I think the desired behavior is already in the master branch, namely "If the user enters a string in the URL parameter for queue ID, it will throw an error in the backend, a Queue Connection will appear in the UI, and NaN will appear in the breadcrumbs." Could you please verify? We may just close #422 as non-issue.

lhuang408:
@zqian Based on the original description, it seemed like they wanted the error-handling of /queue/<string> to match the /queue/<invalid-id>. The statement you commented was in reference to the original behavior: "If the user enters a string in the URL parameter for queue ID, it will throw an error in the backend, a Queue Connection will appear in the UI, and NaN will appear in the breadcrumbs."

Instead of returning the error "An unexpected error occurred. Trying to reconnect...", the error "The resource you're looking for does not exist. Maybe it was deleted?" should be returned instead. This is to indicate that the string is an invalid queue ID (equivalent to a 404 error), rather throwing an error in the backend that is unexpected.

I believe that my PR addresses the issue that was posted in #422 prior to when you edited the description.

jonespm:
This was reworked and going to be implemented in #504

Issue #458, opened 2024-02-14
Newlines & whitespace in REDIS_HOST cause connection errors


jonespm:
It's relatively easy to create a REDIS_HOST file that has a newline in it. It would probably be nice if the code stripped this value out, just calling .strip() on the value. 

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/582957e2a9530be8ec902ef2dac518e3bcee0e29/src/officehours/settings.py#L316C20-L316C20

Leonard6261:
Hi, @jonespm. I am a student at EECS 481, and I would like to claim the issue! Also, I would write a unit test for this to test changes we made.

jonespm:
Okay I assigned you to this. I don't know if this class has any specific tests for the settings yet but sure you could add one!

zqian:
The REDIS_HOST settings on test, qa and prod servers contain no whitespace characters.

PR #473, opened 2024-04-03
add tests for settings, fix possible newline/space in REDIS_HOST

Leonard6261:
This PR introduces a suite of tests for the Django settings of the remote-office-hours-queue project, as well as solving the issue mentioned in #458 . The `test_settings.py` is aimed to catch potential configuration errors early in the development process and maintain a higher standard of reliability. Right now the current app code passes all the setting tests.

Reviewer Notes:
- For some setting variables that take value by using `os.getenv()` without setting a default value, sometimes there may not be a value for them and therefore these variables are `None` type. I am not sure if it is right to raise an error when calling something like `self.assertEqual(settings.EMAIL_HOST, settings.EMAIL_HOST.strip())`  (Here `EMAIL_HOST` is a `None` type as we tested and same thing happens for some other variables), so right now we are using `if ... not None` before testing them. Feel free to leave comment if it is not appropriate.  

- Any feedback on the testing approach and any potential edge cases that might not be covered is highly appreciated!

### DIFF
--- a/src/officehours/test_settings.py
+++ b/src/officehours/test_settings.py
@@ -1,0 +1,151 @@
+import os
+from django.test import TestCase
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_email
+
+class ChannelLayersConfigTest(TestCase):
+    def test_base_dir(self):
+        # BASE_DIR should be set to the project root, and it should be an existing directory
+        base_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+        self.assertTrue(os.path.isdir(base_dir), f"The BASE_DIR path '{base_dir}' does not exist.")
+
+        #check that certain key directories/files exist relative to BASE_DIR
+        manage_py = os.path.join(base_dir, 'manage.py')
+        self.assertTrue(os.path.isfile(manage_py), f"The 'manage.py' file does not exist at '{manage_py}'.")
+
+        #check that other expected directories exist
+        for subdir in ['officehours', 'officehours_api']:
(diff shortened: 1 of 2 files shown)


src/officehours/test_settings.py
```diff
@@ -0,0 +1,151 @@
...
+        self.assertTrue(set(settings.ADMINS).issubset(set(settings.MANAGERS)))
+
+    def test_google_analytics_id(self):
+        if settings.GA_TRACKING_ID is not None:
+            self.assertEqual(settings.GA_TRACKING_ID, settings.GA_TRACKING_ID.strip())
+
+    def test_channel_layers(self):
+        redis_host = settings.CHANNEL_LAYERS['default']['CONFIG']['hosts'][0][0]
+        redis_port = settings.CHANNEL_LAYERS['default']['CONFIG']['hosts'][0][1]
+        self.assertFalse('\n' in redis_host, f"The REDIS_HOST value '{redis_host}' should not contain newlines")
```
zqian, line 103: Should we use self.assertEqual(redis_host, redis_host.strip()) instead? This is to be consistent with the other test cases.

Leonard6261: Changed. Thanks for noticing me to keep it consistent!

jonespm (approved):
I think this one looks good. We'll probably want to create a new issue to create some mock environment probably with [monkeypatch.setenv](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) or similar to provide potential test cases that are going to get passed this and test all the cases. But I can put that in as a new issue.

PR #489, opened 2024-05-01
487: twilio settings check for local build

jaydonkrooss:
Skip over the initialization of Twilio client when the settings are not set locally, and add some helpful logs.

Test Plan:
1. See if local build without Twilio settings runs properly
2. Try running the app with test Twilio settings 
3. Verify that OTP texting service works with Twilio settings (this is what loads the notifications.py import)

### DIFF
--- a/src/officehours_api/notifications.py
+++ b/src/officehours_api/notifications.py
@@ -21,7 +21,7 @@ def initialize_twilio():
         twilio_client = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
         logger.info("Twilio client initialized.")
     else:
-        logger.info("Twilio client setup skipped. Twilio settings values are not set (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN, TWILIO_MESSAGING_SERVICE_SID).")
+        logger.warning("Twilio client setup skipped. Twilio settings values are not set (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN, TWILIO_MESSAGING_SERVICE_SID).")
         twilio_client = None
     return twilio_client
 
--- a/src/officehours_api/notifications.py
+++ b/src/officehours_api/notifications.py
@@ -6,24 +6,26 @@ from django.conf import settings
 from django.dispatch import receiver
 from django.db.models.signals import pre_delete, post_save
 from django.contrib.sites.models import Site
+from django.http import HttpResponseServerError
 from django.urls import reverse
 
+from officehours_api.exceptions import TwilioClientNotInitializedException
 from twilio.rest import Client as TwilioClient
 
 from officehours_api.models import Queue, Meeting, MeetingStatus
 
-def initialize_twilio():
-    logger = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
+def initialize_twilio():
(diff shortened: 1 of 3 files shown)


src/officehours_api/notifications.py
```diff
@@ -12,10 +12,18 @@
...
-twilio = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
-
+def initialize_twilio():
+    logger = logging.getLogger(__name__)
+
+    if (settings.TWILIO_ACCOUNT_SID and settings.TWILIO_AUTH_TOKEN and settings.TWILIO_MESSAGING_SERVICE_SID):
+        twilio_client = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
+        logger.info("Twilio client initialized.")
+    else:
+        logger.info("Twilio client setup skipped. Twilio settings values are not set (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN, TWILIO_MESSAGING_SERVICE_SID).")
```
zqian, line 22: Should the logging level be "warn" instead of "info" here?

src/officehours_api/notifications.py
```diff
@@ -12,10 +12,18 @@
 
 from officehours_api.models import Queue, Meeting, MeetingStatus
 
-logger = logging.getLogger(__name__)
-
-twilio = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
-
+def initialize_twilio():
+    logger = logging.getLogger(__name__)
```
zqian, line 16: line 16 shall be put outside of the initialize_twilio() function. 
logger does not need to be a return object

src/officehours_api/notifications.py
```diff
@@ -12,10 +12,18 @@
...
+
+    if (settings.TWILIO_ACCOUNT_SID and settings.TWILIO_AUTH_TOKEN and settings.TWILIO_MESSAGING_SERVICE_SID):
+        twilio_client = TwilioClient(settings.TWILIO_ACCOUNT_SID, settings.TWILIO_AUTH_TOKEN)
+        logger.info("Twilio client initialized.")
+    else:
+        logger.info("Twilio client setup skipped. Twilio settings values are not set (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN, TWILIO_MESSAGING_SERVICE_SID).")
+        twilio_client = None
+    return logger, twilio_client
+
+logger, twilio = initialize_twilio()
```
zqian, line 26: Is there any further error handling needed when the `twilio` value is None?

src/officehours_api/notifications.py
jaydonkrooss: Pulled it out of the function

src/officehours_api/notifications.py
jaydonkrooss: Not sure if we're using the logger warning elsewhere in the app, but I don't think it'd impact user experience, so went ahead with using a warning.

src/officehours_api/notifications.py
jaydonkrooss: I’ve added a new error type to address this case when calling on “send otp” while twilio settings are not set. Wasn’t super sure if this was the best way to raise an exception but it works for now.

zqian (approved):

jxiao21:
LGTM moving into done

Issue #197, opened 2020-10-29
How would a meeting end up with no attendees?


ssciolla:
While testing on `dev`, I managed to discover a bug (that I don't know how to reproduce) whereby a meeting can end up with no attendees, which throws a JS exception because the UI (in `queueManager.tsx`) assumes meetings have at least one attendee. During the testing, I was deleting meetings, and I was changing a test user's first and last name and email address (actually repeating my own, which caused a separate login problem, but which won't occur in production since email addresses aren't managed manually). This will require more investigation to figure out exactly what happened.

The problematic queue can be found at [https://dev.officehours.it.umich.edu/manage/41/](https://dev.officehours.it.umich.edu/manage/41/)

jlost:
One way it could happen:
* Create meeting for an attendee
* Delete that attendee
* Deletion cascades through the meeting, creating a meeting with no attendees

ssciolla:
This is some internal information on this bug (explicit access is required): https://docs.google.com/document/d/14octPEnQ8MMMmwgu4OgDszPVHv4isom4vSciHNn4hng/edit?usp=sharing

ssciolla:
We also just had an incident today where three meetings with attendees ended up in the queue, but they didn't cause a JavaScript error. It looked like this:
![Screenshot 2023-06-30 115459](https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/35741256/39cbf1d9-a0b1-4acf-a4b0-c77745db29a6)

I couldn't determine from logs that caused this. There were no tracebacks about this, and no evidence of those meetings being created in response to API calls. Puzzling...

jonespm:
We had what I believe is issue again today. This caused the manage page for a specific queue not to work.

When it got to this line in the StartedMeetingEditor there it returned undefined. I had to manually clear the meetings with the UI. 

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/12e2185faf35307564bea910ff1baad7bef1ff76/src/assets/src/components/meetingTables.tsx#L141

The browser error was
`TypeError: Cannot read properties of undefined (reading 'first_name')
`

And within the javascript object it has 

`attendees: []`

jaydonkrooss:
@HemanginiShah  @jxiao21  I've created an invalid meeting to view in ROHQ QA by tweaking the database, just to verify that this is how we'd like to handle this issue: https://qa.officehours.it.umich.edu/manage/483/

[...]

PR #498, opened 2024-05-28
#197 Handle meetings with no attendees, add delete action

jaydonkrooss:
#197 Addresses the issue we've been seeing with meetings without attendees. My solution is to just handle the issue client-side, allowing users to view all affected meetings and manually remove them. 

Test plan to replicate local development environment:
1. On a manage meeting page, add a user to the meeting queue (for in-person meeting)
2. Click "ready for attendee" to admit the user to the "Meetings in Progress" section
3. In local DB, delete the attendee from the officehours_api_attendee table (where meeting_id is from the id from officehours_api_meeting table): `delete from officehours_api_attendee where meeting_id={meeting_id}`
4. On browser, open the console and refresh the page. You should see a meeting with the message "Invalid meeting ID:" and no attendee. It should have a trash icon to delete the meeting
5. Delete the meeting by pressing the delete button and confirming. Double check the database to see that the meeting info is fully cleared

To test the meeting without attendees in just the Meeting Queue, repeat the above but skip over step 2

### DIFF
--- a/src/assets/src/components/meetingTables.tsx
+++ b/src/assets/src/components/meetingTables.tsx
@@ -80,13 +80,20 @@ interface UnstartedMeetingEditorProps extends MeetingEditorProps, AssigneeSelect
 }
 
 function UnstartedMeetingEditor (props: UnstartedMeetingEditorProps) {
-    const attendee = props.meeting.attendees[0];
+    const attendee = props.meeting.attendees[0] ?? null;
     if (!attendee) return (
         <>
         <td>No attendee</td>
         <td></td>
         <td>Invalid meeting ID: {props.meeting.id}</td>
-        <td></td>
+        <td>
+            <RemoveButton
+                onRemove={() => props.onRemoveMeeting(props.meeting)}
+                size="sm"
+                screenReaderLabel={`Remove Meeting ${props.meeting.id} with no attendee`}
+                disabled={props.disabled}
+            />
+        </td>
         </>
     );
     const attendeeString = `${attendee.first_name} ${attendee.last_name}`;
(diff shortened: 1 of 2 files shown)


jxiao21:
Works as you described. Just making sure I understand, the goal is that if there is a meeting with no attendees, that it will say somewhere that the meeting is invalid and that it needs to be deleted? I think it may help to make it a little more clear to the user with something a little more eye-catching that the meeting should be deleted if that's the case.

jaydonkrooss:
> Works as you described. Just making sure I understand, the goal is that if there is a meeting with no attendees, that it will say somewhere that the meeting is invalid and that it needs to be deleted? 

Yes, the invalid data should be presented as opposed to crashing the page, which we saw before.

> I think it may help to make it a little more clear to the user with something a little more eye-catching that the meeting should be deleted if that's the case.

Good suggestion, I'll look into some error message or at least changing the text inside the table to be more clear.

jaydonkrooss:
@jxiao21 I just added a suggestion to delete the meeting in the table entries. This is a rare scenario to begin with so I think that will suffice to indicate to the user that the invalid meeting should be removed.

src/assets/src/components/meetingTables.tsx
```diff
@@ -138,7 +145,22 @@ function UnstartedMeetingEditor (props: UnstartedMeetingEditorProps) {
 }
 
 function StartedMeetingEditor (props: MeetingEditorProps) {
-    const attendee = props.meeting.attendees[0];
+    const attendee = props.meeting.attendees[0] ?? null;
+    if (!attendee) return (
```
zqian, line 149: line 148-163 is identical to line 84-98. Could reformat to use one single source.

zqian (approved):
Following the test plan and tested the feature works as expected. Left reformatting comments.

src/assets/src/components/meetingTables.tsx
```diff
@@ -80,13 +80,20 @@ interface UnstartedMeetingEditorProps extends MeetingEditorProps, AssigneeSelect
...
 function UnstartedMeetingEditor (props: UnstartedMeetingEditorProps) {
-    const attendee = props.meeting.attendees[0];
+    const attendee = props.meeting.attendees[0] ?? null;
     if (!attendee) return (
         <>
         <td>No attendee</td>
         <td></td>
-        <td>Invalid meeting ID: {props.meeting.id}</td>
-        <td></td>
+        <td>Invalid meeting ID: {props.meeting.id}. Please remove this meeting</td>
```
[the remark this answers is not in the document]
jaydonkrooss: added a period

src/assets/src/components/meetingTables.tsx
jaydonkrooss: I went ahead and made the "Invalid Meeting" component generic

Issue #213, opened 2020-11-04
Getting logged out in another window should trigger redirect to login


mfldavidson:
If I am logged in to ROHQ in one window, and in another window I log out, nothing happens on the original page to tell me that I have been logged out. However, when I click on anything, I get errors that might not make sense to an end user. For example, if I get logged out while sitting on the /manage page, then click on a queue, it gives me `Access Denied: You are not a host of this queue. If you believe you are seeing this message in error, contact the queue host(s).` BUT it still shows me as "logged in" if you look in the upper right--i.e. it shows my uniqname and not a Login button.
![image](https://user-images.githubusercontent.com/31591389/98126452-9daf6a00-1e83-11eb-8161-ebcbd77f126f.png)

The only way to get clarity on what is really going on and get the opportunity to login again is if I manually refresh the page.

Particularly bad if someone thinks they are in line and just waiting for the page to update to say it's their turn, or they are a host watching the queue waiting for someone to join, and they don't realize they've been logged out and thus the page isn't updating.

Ideal would be that getting logged out in another window causes the original window to redirect to Login, maybe with a red error message saying "You have been logged out. Please login again." or something.

tayheath:
Hello! I'm currently a student at the University of Michigan in the EECS 481 Software Engineering class and I'd love to try and tackle this issue. I've just recently joined the EECS-481-MyLA google group if you would like to contact me at all. I plan on making a pull request according to the contribution guidelines in a week or so when I complete my implementation of the bug fix!

hemashahumich:
Steps to Reproduce the Issue:
I am hoping I have created the steps as per issue described by Maggie.

1. Open two ROHQ windows and log into both simultaneously.
2. In the first ROHQ window:
   - Log in as a visitor.
   - Open a queue and join it.
3. In the second ROHQ window:
   - Log out.
4.  Return to the first ROHQ window and attempt to create a meeting:
    - Click the "Create Meeting" button.
9. Observed Behavior:
   - I am unable to start the meeting when you click the "Create Meeting" button.
   - At the top of the screen, a message saying "Start Meeting" appears, but the meeting does not actually start.
**Screenshot:**
![image](https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/155757888/2bcf1f55-303c-41b7-9a1f-eaeefe9359cb)

Seeking advice if how should I be testing it?

jonespm:
[...]

PR #499, opened 2024-06-03
PR for #213: Automatic redirect to login

jxiao21:
Made this PR so that we can continue on code changes from #401

### DIFF
--- a/src/officehours_ui/templates/base.html
+++ b/src/officehours_ui/templates/base.html
@@ -3,112 +3,133 @@
 {% get_flatpages as flatpages %}
 <!doctype html>
 <html lang="en">
-  <head>
-  {% block head %}
-	<link rel="shortcut icon" type="image/png" href="{% static 'images/favicon.png' %}"/>
-	<meta charset="utf-8">
-	<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-  <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-rbsA2VBKQhggwzxH7pPCaAqO46MgnOM80zW1RWuH61DGLwZJEdK2Kadq2F9CUG65" crossorigin="anonymous">
-  <link href="{% static 'fontawesomefree/css/all.min.css' %}" rel="stylesheet" type="text/css">
-  <link rel="stylesheet" type="text/css" href="{% static 'css/main.css' %}">
-	<link rel="stylesheet" mo type="text/css" href="//fonts.googleapis.com/css?family=Roboto:400"/>
-	<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:400i"/>
-	<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:700"/>
-	<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:700i"/>
-  {% block styles %}{% endblock %}
-  <title>Remote Office Hours Queue - {% block title %}{% endblock %}</title>
-  {{ spa_globals|json_script:"spa_globals" }}
-  {% endblock %}
-  </head>
+ <head>
+ {% block head %}
+ <link rel="shortcut icon" type="image/png" href="{% static 'images/favicon.png' %}"/>
+ <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


jaydonkrooss:
I'll work on reviewing. as Matt suggested, the formatting on base.html is a little off from the original version, so maybe try to restore the spacing that was on the <head> tag

src/officehours_ui/templates/base.html
```diff
@@ -3,112 +3,140 @@
...
+ <link rel="stylesheet" type="text/css" href="{% static 'css/main.css' %}">
+ <link rel="stylesheet" mo type="text/css" href="//fonts.googleapis.com/css?family=Roboto:400"/>
+ <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:400i"/>
+ <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:700"/>
+ <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:700i"/>
+ {% block styles %}{% endblock %}
+ <title>Remote Office Hours Queue - {% block title %}{% endblock %}</title>
+ {{ spa_globals|json_script:"spa_globals" }}
+ {% endblock %}
+ </head>
```
jaydonkrooss, line 22: these are the formatting changes you made, maybe try copying and pasting from the old version and see if the diff changes

jxiao21: Ok I think it should be good now let me know if I missed something!

jxiao21: I don't know if it's exactly how it was before but at least it hopefully looks more intuitive than whatever it was when I originally pushed it

jonespm:
Note: When this is merged please include

Co-authored-by: tayheath (Taylor Heath) <tayheath@umich.edu>

src/officehours_ui/templates/base.html
```diff
@@ -21,94 +21,121 @@
...
-  <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.min.js" integrity="sha384-cuYeSxntonz0PPNlHhBs68uyIAVpIIOZZ5JqeqvYYIcEL727kskC66kF92t6Xl2V" crossorigin="anonymous"></script>
-  {% block scripts %}{% endblock %}
 
-  {% get_flatpages '/dataalert/' as flatpages %}
-  {% if flatpages.first and flatpages.first.content %}
-    <footer class="sticky-footer">
-      <div class="alert alert-info alert-footer" role="alert">
-        <div class="alert-footer-text">
-          {{flatpages.first.content|safe}}
+    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
```
jaydonkrooss, line 102: @jonespm has maybe mentioned this before, but I'm not seeing anywhere in the code that actually makes use of the JQuery import. If everything behaves as intended without this, I suggest removing it.

src/officehours_ui/templates/base.html
```diff
@@ -21,94 +21,121 @@
...
+      if (logoutButton !== null) {
+        logoutButton.addEventListener('click', () => {
+          window.localStorage.setItem("loggedIn", "false");
+        });
+      }
+      window.addEventListener('storage', (event) => {
+        //  if (document.cookie.indexOf('JSESSIONID') === -1) {
+        //  if (document.cookie.indexOf('sessionid') === -1) {
+        //    window.location.href = "/accounts/login/";
+        //  }
```
jaydonkrooss, line 117: Remove unused code in comments

jaydonkrooss:
Logic is looking like it's working, I tested logging out on different browsers and the redirect behaved as expected

src/officehours_ui/templates/base.html
jxiao21: Appears to work without JQuery so I removed it

src/officehours_ui/templates/base.html
jxiao21: Done and pushed

jaydonkrooss (approved):

jxiao21:
Test plan:
1. Open ROHQ on two different tabs (ex. 2 different Chrome tabs; I don't know if it's possible to fix this issue on, for example, one Chrome tab and one Safari tab).
2. Make sure you're logged in on both tabs.
3. Log out on one tab, then check the other tab. It should be logged out now as well.

Issue #97, opened 2020-06-24
Self-service reporting by queue


mfldavidson:
In order to meet the potential use case of lobby management, as well as to respond to general inevitable requests for reporting capabilities, Remote Office Hours Queue should make it possible for users to pull reports about usage of their queue. Some elements that may be important to report on include:

- Most popular times of the day, days of the week, and weeks of the year
- Number of attendees seen by each host
- Most popular words used in the agenda field
- Number of unique attendees in a period of time

Reporting could be implemented by sending Caliper events to the Unizin Data Platform (UDP) and providing reports that utilize the Google Cloud Platform connection to the UDP.

mfldavidson:
Requested by 2 users to date.
- TDX 79708
- TDX 103155

HemanginiShah:
[[TDX 5344805 - ](https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet.aspx?TicketID=5344805)
advisors like to keep a record of who they talk to for follow ups. Currently they are manually recording the information and thought it would be easier if they could get a report from the queue itself

hemashahumich:
Additional request for the historing data reporting feature: TDX:
https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet.aspx?TicketID=5640567
https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet.aspx?TicketID=5640293

Can we give this a high priority to include in next release?
Thanks.

hemashahumich:
Another ticket came in with log request: https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet.aspx?TicketID=5733961

hemashahumich:
"1. Queue name is not appearing in downall all history button download csv file.  
2. Deleted queue are not getting listed. 
3. Can we get that to downloaded
Currently, the reporting button ""Download"" and the ""Download All Queue History"" are tracked based on the signed-in user.  Question: Should the admin or service manager have access to view the Download button or download all queue history button for a single queue and for all queues?"

jonespm:
1. What do you mean by queue name, like in the report? Yes the query only has the queue id. I think including the name might take more effort since I'm not sure if it's in the log table and would need a join against the queue table. I feel adding this for this release could delay this a week. We could create an issue for next release. 
[...]

Issue #470, opened 2024-03-22
Reporting download link options


hemashahumich:
We need to determine the most effective way to provide queue-level information to individual queue owners. One option we discussed is to include a reporting link either under the "Manage Queue" page or at the top pulldown option next to the login name top right corner. Additionally, we could consider placing a link in a column preceding or following the status column for each queue. For instance, we could incorporate a download button on the individual queue details page, accessible to anyone with queue management privileges. This would allow for easy access and utilization of the reporting feature.

I will let you decide on the easiest way to implement this feature.

OscarSong2003:
Hey, I'm a current student in EECS 481 who is looking to contribute. Wondering if I can potentially contribute to this task? Thanks!

zqian:
@OscarSong2003 Thanks for your interest. However, this issue may need more efforts and may not be an ideal candidate for your EECS 481 assignment. 

Instead, please choose from those open issues with the label [Good First Issue](https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22)

zqian:
will access whether this is still needed once #97 is implemented

hemashahumich:
Sounds good to me.

hemashahumich:
Thanks so much Jaydon.  I have added the test cases to the Release
Plan/test file:
https://docs.google.com/spreadsheets/d/1JZUKW2WCPPPJccGXcCM1cNm8Ksv5DaNTnUzDpEbxCK0/edit?gid=1624795005#gid=1624795005


<https://docs.google.com/spreadsheets/d/1JZUKW2WCPPPJccGXcCM1cNm8Ksv5DaNTnUzDpEbxCK0/edit?gid=1624795005#gid=1624795005>


*With Warm Regards,**  <https://www.name-coach.com/hemangini-shah>*

*Hemangini Shah <https://www.name-coach.com/hemangini-shah> *She/Her/Hers*, PMP
, *BI Analyst Senior, ITS

Manages Services: Student Success Viewer
<https://its.umich.edu/academics-research/teaching-learning/student-success-viewer>
, Remote Office Hour Queue
<https://its.umich.edu/communication/videoconferencing/remote-office-hours-queue>
, Canvas/NameCoach
<https://its.umich.edu/academics-research/teaching-learning/namecoach>,
Canvas/Perusall
[...]

PR #501, opened 2024-06-03
Fixes issue #97 - Initial work for reporting feature

jonespm:
I've opened this for review as some of what this needs to do is completed. 

Adding view to retrieve log data for reporting
- [x] Add view for retrieving reporting data
- [x] Add Django Model to represent reporting data view 
- [x] Add some way of retrieving the CSV data
- [x] Change CSV data export to retrieve hosts from the queue hosts table rather than searching since it won't include meetings where someone is a host but wasn't an assignee
- [x] Write unit test
- [x] Drop unnecessary backend metadata and view columns
- [x] Flatten attendees and assignee columns
- [x] Rename assignee to Host
- [x] Review other columns requested on the meeting notes
- [x] Look at rewriting this to use the view more directly with Django with something like this
```
    with connection.cursor() as cursor:
        cursor.execute("SELECT * FROM my_table")
        rows = cursor.fetchall()
    return rows
```
- [x] Update failing tests or fix code 
- [x] Verify queue_id security check is working
- [x] Verify and/or write test to verify deleted queues are showing in the report

Front end will be implemented as part of #470


jonespm:
All of the changes have been made here.

The tests are currently failing and I have to look at this
Also I don't know if the security check is working

So will add those as tasks.

src/officehours_api/test_views.py
```diff
@@ -92,6 +97,29 @@ def test_cannot_change_backend_type_when_meeting_has_started(self):
...
+        self.client.login(username='hosttwo', password='rohqtest')
+        # Start the meeting through the api to generate logs
+        response = self.client.post(f'/api/meetings/{self.meeting.id}/start')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.client.login(username='hostone', password='rohqtest')
+        # Start the meeting through the api to generate logs
+        response = self.client.get(f'/api/export_meeting_start_logs/')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        # This may change if the output of the export changes
+        self.assertEqual(len(response.content), 343)
```
zqian, line 110: Tested with command `docker-compose run web python manage.py test officehours_api.test_views.MeetingTestCase`, and this test failed

======================================================================
FAIL: test_export_meeting_start_logs (officehours_api.test_views.MeetingTestCase)
----------------------------------------------------------------------
  File "/usr/src/app/officehours_api/test_views.py", line 110, in test_export_meeting_start_logs
    self.assertEqual(len(response.content), 343)
AssertionError: 289 != 343
 test case failed

src/officehours_api/test_views.py
```diff
@@ -92,6 +97,29 @@ def test_cannot_change_backend_type_when_meeting_has_started(self):
...
+
+    def test_export_meeting_start_logs_by_queue(self):
+        self.client.login(username='hosttwo', password='rohqtest')
+        # Start the meeting through the api to generate logs
+        response = self.client.post(f'/api/meetings/{self.meeting.id}/start')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        response = self.client.get(f'/api/export_meeting_start_logs/{self.queue.id}/')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        # This may change if the output of the export changes
+        self.assertEqual(len(response.content), 345)
```
zqian, line 120: Tested with command `docker-compose run web python manage.py test officehours_api.test_views.MeetingTestCase`, and this test failed

======================================================================
FAIL: test_export_meeting_start_logs_by_queue (officehours_api.test_views.MeetingTestCase)
----------------------------------------------------------------------
  File "/usr/src/app/officehours_api/test_views.py", line 120, in test_export_meeting_start_logs_by_queue
    self.assertEqual(len(response.content), 345)
AssertionError: 291 != 345

src/officehours_api/test_views.py
jonespm: Yes, one of the unchecked TODO's I have is "Update failing tests or fix code". These are new tests I added and rather than verify against a string I figured checking the length was good. 

I didn't completely re-verify that this was working after making the SQL updates and expected it to fail. I just pushed the work I had so I could go over it with Justin and Jaydon. The new size looks like it could be correct but I need to verify the security checks were working.

src/officehours_api/views.py
```diff
@@ -290,3 +295,48 @@ def get_queryset(self):
...
+        else:
+            filename = f"meeting_start_logs_{username}.csv"
+        
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = f'attachment; filename="{filename}"'
+
+        writer = csv.writer(response)
+        with connection.cursor() as cursor:
+            # Add a query to just get the logs for the queues_user_is_in
+            cursor.execute("SELECT * FROM meeting_start_logs where queue_id in (%s)" % queues_user_is_in)
```
jaydonkrooss, line 329: While locally testing, this works for individual queues, but for exporting all queues runs into problems with more than one. This is the error that appears on this cursor line:

> TypeError at /api/export_meeting_start_logs/
not all arguments converted during string formatting

I looked up the this error message & found a solution to rewrite the tuple: https://stackoverflow.com/a/1455623
You may need two different versions of this line based on the length of `queues_user_is_in`, so that the multiple queues case looks like this:
`cursor.execute("SELECT * FROM meeting_start_logs where queue_id in %s" % (queues_user_is_in,)) # for multiple only`

jonespm:
Thanks, I don't think I have a unit test for exporting multiple yet. I should probably also add that in there. I don't think I'd tested yet with more than 1 and wasn't sure how this code would hold up. I can try that.

src/officehours_api/test_views.py
zqian: I wonder whether the assert can be based on the count of rows in the csv file, not on the length.

jonespm: Yeah that was a good idea, I was thinking the same. The data will vary in length and content because of some of the fields. I changed it to this.

src/officehours_api/views.py
jonespm: I changed this to create placeholders and use the correct way of passing values into the execute. This looks like it's working now for single and multiple queues. I also found that the call "to all_objects" instead of just "objects" returns deleted objects as well.

src/officehours_api/views.py
```diff
@@ -290,3 +295,50 @@ def get_queryset(self):
...
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = f'attachment; filename="{filename}"'
+
+        writer = csv.writer(response)
+        with connection.cursor() as cursor:
+            # Generate a string of placeholders, one for each item in the list
+            queue_id_placeholders = ', '.join(['%s'] * len(queues_user_is_in))
+
+            # Add a query to just get the logs for the queues_user_is_in
+            cursor.execute(f"SELECT * FROM meeting_start_logs where queue_id in ({queue_id_placeholders})", queues_user_is_in)
```
jaydonkrooss, line 331: image

if the user has no queues to export, we get a 500 error (where we probably want [200/204?](https://stackoverflow.com/questions/11746894/what-is-the-proper-rest-response-code-for-a-valid-request-but-empty-data)). alternatively an empty csv would work as well

jaydonkrooss (commented):
Tests run and the 0 queues case is handled well. Go ahead with merging, I can try to add the 204 message to the frontend

jaydonkrooss (approved):

jonespm:
Okay if we have to adjust something else here because of difficulty with the frontend that seems easier just to do in a new PR at this point. I think it could also wait till next release. The chances of someone trying to export without any queues at all seems low.

PR #504, opened 2024-06-07
Issue 422: handle urls

### LINKED ISSUE #422: Handle string in URL parameters that pass through React Router

jxiao21:
fixes #422 

To test locally, start local environment and then go to `http://localhost:8003/queue/<any-string>/` and `http://localhost:8003/manage/<any-string>/` to check that the breadcrumb matches the invalid one that was entered and error message.

Note: once Django and websocket libraries are updated, we can try adding the` <any-string>` into the error message, but for now this part of the fix is probably more trouble than it's worth.

### DIFF
--- a/src/assets/src/components/queue.tsx
+++ b/src/assets/src/components/queue.tsx
@@ -444,7 +444,6 @@ export function QueuePage(props: PageProps) {
     const { queue_id } = useParams();
     if (queue_id === undefined) throw new Error("queue_id is undefined!");
     if (!props.user) throw new Error("user is undefined!");
-    const queueIdParsed = parseInt(queue_id);
 
     //Setup basic state
     const [selectedBackend, setSelectedBackend] = useState(undefined as string | undefined);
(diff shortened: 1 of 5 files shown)


jaydonkrooss:
I will look into reviewing the logic for this tomorrow.

jaydonkrooss:
Looks good and both /manage/ and /queue/ endpoints test correctly. Nice work getting the backend to handle these types of invalid requests better.

Last thing I'd suggest is to remove any unused, commented-out code.

jxiao21:
Commented out code should be all removed now

jaydonkrooss (approved):

PR #505, opened 2024-06-07
Fixes #461 - Message in log about migrations 1.8.0 and 1.9.0

### LINKED ISSUE #461: Message in log about migrations 1.8.0 and 1.9.0 release
In the logs from at least 1.8.0 there's a message in the logs that some migrations haven't been created 

```
  Your models in app(s): 'officehours_api' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.			
```

Running this creates 5 migrations. These don't appear to have an impact on the app's running.

This is what is generated:
```
Migrations for 'officehours_api':
  officehours_api/migrations/0024_auto_20240219_1419.py
    - Alter field deleted on attendee
    - Alter field backend_type on meeting
    - Alter field deleted on meeting
    - Alter field allowed_backends on queue
    - Alter field deleted on queue
```

It looks like most of these involved adding `db_index=True` when safe_delete was upgraded in 1.8.0 or changing the "choices" when all backends aren't enabled. 

It looks like according to the documentation. you need to run makemigrations with` python manage.py makemigrations --settings=officehours.makemigrations_settings` to avoid that change and there hasn't been consistency with running that. However this still generates the Bluejeans. I think we should remove the Bluejeans #462 and re-run this for 1.10.0. Running without the indexes seems fine for now.

jonespm:
Merge after #501


zqian (approved):
I have tested the added migration file in this PR: 

- copied `0025_meeting_start_logs_view.py` from #501 into `src/officehours_api/migrations/`
- ran the PR
- confirmed there is no more missing migration alert as reported in #461

PR #509, opened 2024-06-17
Fixes #359 Fixing quotes in access log format

### LINKED ISSUE #359: Update log format to include X-Forwarded-For to capture users IP in logs
We had an email discussion in January 2022 to determine the source of traffic and changing the log to include the origin IP. It was suggested that the [guincorn configuration](https://github.com/chriskretler/openshift-python-sample/blob/main/src/guniconf.py#L23) could pick this up from:

`access_log_format = '%({X-Forwarded-For}i)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s %(	`

@jlost commented that "As for ROHQ, you could probably inline that line on the CMD if you think it's worth it to update."

jonespm:


jonespm:
This change was reverted and till will probably need more work.

PR #511, opened 2024-06-19
Automatic logout - second PR try, redirects by triggering logout button click event

### LINKED ISSUE #213: Getting logged out in another window should trigger redirect to login

jxiao21:
Fixes #213 
Co-authored-by: tayheath (Taylor Heath) [tayheath@umich.edu](mailto:tayheath@umich.edu)

### DIFF
--- a/src/officehours_ui/context_processors.py
+++ b/src/officehours_ui/context_processors.py
@@ -14,6 +14,9 @@ def feedback(request):
 def login_url(request):
     return {'LOGIN_URL': getattr(settings, 'LOGIN_URL', None)}
 
+def logout_redirect_url(request):
+    return {'LOGOUT_REDIRECT_URL': getattr(settings, 'LOGOUT_REDIRECT_URL', None)}
+
 
 def debug(request):
     return {'DEBUG': settings.DEBUG}
--- a/src/officehours_ui/context_processors.py
+++ b/src/officehours_ui/context_processors.py
@@ -14,9 +14,6 @@ def feedback(request):
 def login_url(request):
     return {'LOGIN_URL': getattr(settings, 'LOGIN_URL', None)}
 
-def logout_redirect_url(request):
-    return {'LOGOUT_REDIRECT_URL': getattr(settings, 'LOGOUT_REDIRECT_URL', None)}
-
 
 def debug(request):
     return {'DEBUG': settings.DEBUG}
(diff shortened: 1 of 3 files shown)


jaydonkrooss (approved):

src/officehours_ui/context_processors.py
```diff
@@ -14,6 +14,9 @@ def feedback(request):
 def login_url(request):
     return {'LOGIN_URL': getattr(settings, 'LOGIN_URL', None)}
 
+def logout_redirect_url(request):
```
jonespm, line 17: Discussed that this isn't needed with the change in how it clicks.

src/officehours_ui/context_processors.py
```diff
@@ -42,5 +45,6 @@ def spa_globals(request):
             'backends': backend_dicts,
             'default_backend': settings.DEFAULT_BACKEND,
             'otp_request_buffer': settings.OTP_REQUEST_BUFFER,
+            'logout_redirect_url': settings.LOGOUT_REDIRECT_URL,
```
jonespm, line 48: This isn't needed either

src/officehours_ui/templates/base.html
```diff
@@ -110,7 +113,8 @@
       }
       window.addEventListener('storage', (event) => {
         if (event.key === 'loggedIn' && event.newValue === 'false' && event.oldValue === 'true') {
-          window.location.href = '/accounts/login';
+          // window.location.href = "{{LOGOUT_REDIRECT_URL}}";
```
jonespm, line 116: Remove comment

jonespm (approved):

Issue #493, opened 2024-05-21
complete last step of Google Analytics 4 migration 


zqian:
We have integrated the GA4 property. The only remaining step is the verify user consent settings:

```
Verify consent settings
Set up your property to record consent from end users in the European Economic Area (EEA) to allow their personal data to be shared with Google for advertising purposes. To ensure there is no impact to your ads personalization and measurement, verify your consent settings for the Google tag and other data sources. [Learn more about sending consent signals to Google Analytics](https://support.google.com/analytics/answer/14275483?hl=en_US&utm_id=ad)
```

jonespm:
This will be implemented similarly to https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1580

And likely use these patterns [vpcomm.umich.edu/resources/cookie-disclosure](https://vpcomm.umich.edu/resources/cookie-disclosure/)

jaydonkrooss:
Test plan for Google Analytics OneTrust consent integration. You will need: 
- Access to the Google Analytics realtime dashboards: [Dev](https://analytics.google.com/analytics/web/#/p312490036/realtime/overview) or [QA](https://analytics.google.com/analytics/web/#/p384149532/realtime/overview)
- A free-to-use VPN that allows you to spoof your location (eg. [TunnelBear ](https://www.tunnelbear.com/) or [Windscribe](https://windscribe.com/))

SETUP: Once you have a VPN installed and configured, you need to make sure that Google Analytics activity is not blocked, so make sure to turn OFF any automatic analytics & tracking blocking on both your VPN and browser (such as if you're using Brave). For example, turn off "Ad + trackers" blocking in the preferences of Windscribe like so:
![Screenshot 2024-06-28 at 3 28 54 PM](https://github.com/tl-its-umich-edu/remote-office-hours-queue/assets/147423601/ec8b2ee1-edf1-4a0c-adf6-321e7b79bb78)

Scenario 1: for a user in a non-EU country, Office Hours automatically enables GA tracking and does NOT show the consent banner. 
- Have your VPN disabled or enabled for a country not in the EU
- start an incognito browser and navigate to Officehours dev or QA. As you remain on the home page, you should not see a popup window at the bottom of the page
- Log in to officehours normally and navigate to any link on the page (such as "Manage Queues" or "Preferences")
- Check the google analytics realtime dashboard. You should see recent activity for your location

[...]

PR #518, opened 2024-06-25
493: update Google Analytics settings, use Umich OneTrust banner

jaydonkrooss:
Attempt to adapt the oneTrust cookie disclosure as specified by umich resources: https://vpcomm.umich.edu/resources/cookie-disclosure/ 

Uses react effects to keep track of the user's selections and update google analytics tags accordingly.

see VPN test plan here: https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/493#issuecomment-2197611029


pushyamig:
I will review it in the morning.

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,22 +1,72 @@
...
+                ad_user_data: "granted",
+                ad_personalization: "granted",
+                personalization_storage: "granted"
+              });
+            } else {
+              // Remove Google Analytics cookies if tracking is declined
+              document.cookie.split(';').forEach(cookie => {
+                const [name] = cookie.split('=');
+                if (name.trim().match(/^_ga(_.+)?$/)) {
+                  document.cookie = `${name}=;path=/;domain=.${window.location.host.replace(/^(.*\.)?(.+\..+)$/, '$2')};expires=Thu, 01 Jan 1970 00:00:01 GMT`;
```
jonespm, line 61: Is this something we have to do or something that's done automatically? Did you find this as a suggestion somewhere?

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,22 +1,72 @@
...
+            ad_personalization: "denied",
+            wait_for_update: 500
+        });
         setInitialized(true);
         GoogleAnalytics.initialize(googleAnalyticsId, { testMode: debug });
     }
 
+    const [loaded, oneTrustActiveGroups, setOptanonWrapper] = useOneTrust();
+    if (loaded) {
+        console.log("loaded oneTrustActiveGroups" + oneTrustActiveGroups);
```
jonespm, line 29: I'd probably lower this to` console.debug` if you want to have some type of indication in the frontend still.

src/assets/src/hooks/useOneTrust.ts
```diff
@@ -0,0 +1,50 @@
...
+
+declare global {
+    interface Window {
+      OnetrustActiveGroups?: string;
+      OptanonWrapper?: () => void;
+    }
+  }
+
+export const useOneTrust = (): [boolean, string, (callback: () => void) => void] => {
+    const src = 'https://cdn.cookielaw.org/consent/03e0096b-3569-4b70-8a31-918e55aa20da/otSDKStub.js'
```
jonespm, line 11: Ideally we could pass these 2 values in via config rather than hardcode them, maybe with `spa_globals`? I'm not sure how difficulty that would be with this app though. I'm also not sure if these would ever change, but would require a code change if they do.

zqian: The value "03e0096b-3569-4b70-8a31-918e55aa20da" is defined in line 12.

pushyamig: If the `src` and `dataDomainScript`  variables  values is made available from `spa_globals`  (which i am assuming is Django FE variable shared during the launch of the tool) then the variable values `src` and `dataDomainScript` could be passed as React Props to the component.

As Zhen, shared the alpha numeric values seems to be same at this point `src` and `dataDomainScript` and will be in future then we could reuse with some logic.

src/assets/src/hooks/useOneTrust.ts
```diff
@@ -0,0 +1,50 @@
...
+                document.head.removeChild(script);
+            }
+        };
+    }, []);
+
+    useEffect(() => {
+        if (window.OnetrustActiveGroups) {
+            setOneTrustActiveGroups(window.OnetrustActiveGroups);
+        }
+    }, [window.OnetrustActiveGroups]);
```
pushyamig, line 43: I don't see anything wrong in having multiple useEffects. As long as it work without error its fine.

https://stackoverflow.com/a/54004148/724238

 Can you also share what these useEffects are doing? I would also leave some comment in code

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,22 +1,72 @@
...
 
     if (googleAnalyticsId && !initialized) {
+        GoogleAnalytics.gtag("consent", "default", {
+            ad_storage: "denied",
+            analytics_storage: "denied",
+            functionality_storage: "denied",
+            personalization_storage: "denied",
+            ad_user_data: "denied",
+            ad_personalization: "denied",
+            wait_for_update: 500
```
pushyamig, line 21: Optional: Make the `denied` as const and add as object value  here. So if need to change only one place to change

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,22 +1,72 @@
...
+    useEffect(() => {
+        const updateGtagConsent = () => {
+
+            if (oneTrustActiveGroups.includes("C0002")) {
+              GoogleAnalytics.gtag("consent", "update", { analytics_storage: "granted" });
+            }
+            if (oneTrustActiveGroups.includes("C0003")) {
+              GoogleAnalytics.gtag("consent", "update", { functional_storage: "granted" });
+            }
+            if (oneTrustActiveGroups.includes("C0004")) {
```
pushyamig, line 49: If you add some links or info defining what these code are  `C0002` `C0003' `C0004` would be useful. looks like magic alpha-numeric series.

pushyamig:
Is there any resource/site you were looking at when implementing this code? I don't have much background with onetrust and adding the cookie banner. If you provide a source to reference that would to helpful to review.

src/assets/src/hooks/useGoogleAnalytics.ts
pushyamig: I think you can add this link https://vpcomm.umich.edu/resources/cookie-disclosure/ to the file for more details. 
Could benefit from one-line commets

Strictly Necessary Cookies: C0003 (Functional)

src/assets/src/hooks/useGoogleAnalytics.ts
pushyamig: This could be generated dynamically

```
function getSpecificExpirationDate() {
    // Create a new Date object set to January 1, 1970, 00:00:01 GMT
    const date = new Date(Date.UTC(1970, 0, 1, 0, 0, 1));
    // Convert the date to a string in the required format
    return date.toUTCString();
}

document.cookie = `${name}=;path=/;domain=.${window.location.host.replace(/^(.*\.)?(.+\..+)$/, '$2')};expires=expires=${getSpecificExpirationDate}
```

Looks like this is set to expire in past. what is the purpose of setting a date in past? or it is working as cookie to never expire

pushyamig: `/^(.*\.)?(.+\..+)$/, '$2'`

to would be helpful to add comment saying what string it is parsing? RegEx is be confusing

zqian: The code block is provided in https://vpcomm.umich.edu/resources/cookie-disclosure/#3rd-party-google-analytics

jaydonkrooss: This is from step 4 of the "GOOGLE ANALYTICS / TAG MANAGER" section on the [UMich cookie disclosure](https://vpcomm.umich.edu/resources/cookie-disclosure/) integration. Seems like denying tracking just deletes the google analytics cookie manually. I'm not a fan of this manual extraction, there might be a way to do it with ga4-react library but I haven't looked into it

pushyamig:
I will try to test it on Dev instance as @jaydonkrooss provided a link.

jaydonkrooss:
> Is there any resource/site you were looking at when implementing this code? I don't have much background with onetrust and adding the cookie banner. If you provide a source to reference that would to helpful to review.

Here's my high level understanding of what _should_ happen based on the [Umich cookie disclosure instructions ](https://vpcomm.umich.edu/resources/cookie-disclosure/) under "Google Analytics / Tag manager" section. I've also seen similar instructions on google analytics and onetrust documentation:

1. Google Analytics is initialized with default ("denied") consent signals. React-ga4 library already took care of this for me, except for the initial consent config which I added.
2. The Umich-configured OneTrust script runs the banner code. Appears for users in EU, not anywhere else
3. Frontend should define an "**OptanonWrapper**" function and push it to the window. OneTrust code uses the fn for callback whenever consent signals (stored as a string "**OnetrustActiveGroups**") are changed. I assume "Accept All" flow is immediately run for Non-EU users, while EU users stick with denied unless Accept is selected
4. Inside OptanonWrapper are a number of "**gtag**" updates to Google Analytics so that consent settings are updated. In the denied flow, the GA cookie is completely removed from the window. Also we send a custom Google Analytics event called "**um_consent_updated**" (not sure what it does)

I'll try to add clearer comments and better variable storage. Mainly the challenge has been trying to find out how to combine the oneTrust expected behavior with Google Analytics react library, instead of using purely script tags.

src/assets/src/hooks/useOneTrust.ts
jaydonkrooss: I'll try to leave more clear comments in the code in my next commit. The effects are trying to mimic the Javascript flow listed in the umich cookie instructions (see [my comment](https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/518#issuecomment-2192068882) ) while keeping the OnetrustActiveGroups state updated, but I hope I'm not causing into too many effect runs that aren't necessary compared to the purely JS script-tag method laid out in the instructions

src/assets/src/hooks/useGoogleAnalytics.ts
jaydonkrooss: I've added the use of Cookies.remove() from js-cookie to clear the GA cookies, looks like it works well.

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,22 +1,72 @@
...
+import { useOneTrust } from './useOneTrust';
+
 
 export const useGoogleAnalytics = (googleAnalyticsId?: string, debug?: boolean) => {
     let location = useLocation();
+
     const [initialized, setInitialized] = useState(false);
     const [previousPage, setPreviousPage] = useState(null as string | null);
 
     if (googleAnalyticsId && !initialized) {
```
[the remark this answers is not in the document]
jaydonkrooss: I've added some comments referencing the instructions and clearing up what it does.

src/assets/src/hooks/useGoogleAnalytics.ts
jaydonkrooss: added an enum and some clarifying comments

src/assets/src/hooks/useGoogleAnalytics.ts
jaydonkrooss: added an enum for this purpose

pushyamig:
I feel the code is cleaned up a lot and useEffect in the oneTrust component is simplified.  I did not test his but @zqian seems to be testing all the useCase. I feel if the code is working as expected then @zqian will be the approver.

pushyamig:
@jaydonkrooss Thanks for adding all the comment I feel the code is now easier to understand.

src/assets/src/hooks/useGoogleAnalytics.ts
jaydonkrooss: removed the console logs for now

src/assets/src/hooks/useOneTrust.ts
jaydonkrooss: I went ahead and simplified useOneTrust to just export the JS code we needed to run, without useEffects which was complicating it.

src/assets/src/hooks/useGoogleAnalytics.ts
```diff
@@ -1,22 +1,72 @@
...
+            if (oneTrustActiveGroups.includes("C0003")) {
+              GoogleAnalytics.gtag("consent", "update", { functional_storage: "granted" });
+            }
+            if (oneTrustActiveGroups.includes("C0004")) {
+              GoogleAnalytics.gtag("consent", "update", {
+                ad_storage: "granted",
+                ad_user_data: "granted",
+                ad_personalization: "granted",
+                personalization_storage: "granted"
+              });
```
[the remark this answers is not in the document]
jaydonkrooss: now assigned to granted enum

PR #527, opened 2024-07-02
🚧 add secret keys (iss. #523)

### LINKED ISSUE #523: add new secret `ONE_TRUST_SCRIPT_DOMAIN`
Make a new environment variable available to the application, `ONE_TRUST_SCRIPT_DOMAIN`.

It requires two small changes:

1. In `service/base/kustomization.yaml`, add the line…
   ```
   secret/ONE_TRUST_SCRIPT_DOMAIN
   ```
   _This change needs to be committed to this GitHub repo._

2. Then create the file service/base/secret/ONE_TRUST_SCRIPT_DOMAIN which contains a single line, the value for the key…
   ```
   03e0096b-3569-4b70-8a31-918e55aa20da
   ```
   The file was put in that location based on the premise that the same value would be used in all environments, prod, test, and dev.  If instead the environments will have unique values for this key, it can be added to the directories for each one.

   _This change needs to be committed to the secret archives in Dropbox._

When running the Kustomize process as shown in README.md, it produces a long output containing the base64-encoded secrets and other resources.  The value was found and decoded to verify it was correct.

Using the full command to apply the output to ROSA, the ROSA web UI was used to find the secret, see its value, and confirmed it is correct.

As noted above, the two changes described above need to be committed in two places.  The first step must be committed to the GitHub repo.  The second step must be saved in an updated secrets file in Dropbox.

lsloan:
Resolves #523.

Adds a key for a new secret, `ONE_TRUST_SCRIPT_DOMAIN`.  This new key causes Kustomize to expect a file with that name to be present and contain the key's value.  That file should be included in `base.zip`, which is generated from the contents of the Dropbox `base` folder.

Also update `service/README.md` with details about where the key values are kept and how to update them.


jonespm:
The change this referenced was reverted back in https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/518/commits/a78f282d56c8d468bbf2692f4f583f5032e5c9cd to get the release out. 

It should be added back in as a separate PR one one that's put in along with this one for a future release.

jonespm:
This was mostly (or entirely?) resolved in #594. 

I've added the file to Dropbox. 

Can you review if there's anything else here that needs to be included and either update or close this?

jonespm:
ONE_TRUST_SCRIPT_DOMAIN is no longer being used

PR #528, opened 2024-07-03
Fixes #526 - Rename dataalert

### LINKED ISSUE #526: Rename dataalert flat-page
In #469 the flat page `/dataalert/` was added but that naming doesn't really reflect what this does anymore. This should be renamed to be something like `/page-footer/` or '/privacy-and-copyright/'.

Also it would be nice if we had a migration to fix this renaming so we didn't have to manually change it during the release. You should be able to do that with the [FlatPage Python API](https://docs.djangoproject.com/en/5.0/ref/contrib/flatpages/#via-the-python-api)

In the migration is should create a new placeholder for the new name if `/dataalert/` doesn't exist, or rename it if `/dataalaert/` already does exist.

jxiao21:
Fixes #526 

Still WIP, need to clarify which migration operations are necessary


jonespm (approved):
I think this looks good, I'm reviewing why the migrations came up for you but I think it's related to the configuration you have and the fact we haven't removed Bluejeans yet and they shouldn't be included.

jonespm:
I think I might have screwed some of this up myself in https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/1.10.0/src/officehours_api/migrations/0026_auto_20240604_1620.py and committed it :/

PR #530, opened 2024-07-11
Fixes #529 - Adding a model to prevent Django from managing the view

### LINKED ISSUE #529: Add proxy model to prevent migration of view
Django is trying to migrate a view that was manually created, we should add a proxy model so it won't try to manage this. The alternative looks like might be a SAVEPOINT but this seems the best.

jonespm:


jonespm:
@jxiao21 This should fix one of the migrations you're seeing. I think the other one is going to require removing bluejeans but I'll have to look more into it.

zqian (commented):
Should there be a migration file for the newly added database table?

jonespm:
Hi @zqian. The idea of this was to prevent Django from trying to creating migration files for this.

There is already a migration to create this view, it was done manually in https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/src/officehours_api/migrations/0025_meeting_start_logs_view.py rather than through Django so this change will have Django ignore this and not try to drop the table.

zqian (approved):

PR #544, opened 2024-10-07
#462 removed BlueJeans from codebase

### LINKED ISSUE #462: Remove Bluejeans as a backend choice
Bluejeans is currently not used and tested as a backend so it's unknown if it works. Leaving it in there makes it generate code in the migrations for it to be a choice. I don't think we have any reason to leave this there and should either be deleted or disabled.

zqian:
Fixes #462


jonespm (approved):
This looks good, I don't see any references. `0026_auto_20240604_1620`.py removes it as part of the conversion so migrations are stable.

PR #546, opened 2024-10-09
issue_533 note text updated

### LINKED ISSUE #533: Add additional text on queue page about when notifications are sent to hosts
Notifications are only sent to hosts when there is nobody in the queue. This is mentioned in the help but we should add some text to the queue page about this. 

There's currently a note like

> Note: The person you want to add needs to have logged on to Remote Office Hours Queue at least once in order to be added.

We could add

> If the queue is not empty, hosts will not be notified. Please manually remove attendees from queue after meeting with them or they are no longer available.

zqian:
Fixes #533

### DIFF
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -28,8 +28,7 @@ export const UserDisplay = (props: UserDisplayProps) =>
 
 export const userLoggedOnWarning = (
     <Alert variant='primary'>
-        <strong>Note:</strong> The person you want to add needs to have logged on to Remote Office Hours Queue
-        at least once in order to be added.
+        <strong>Note:</strong> If the queue is not empty, hosts will not be notified. Please manually remove attendees from queue after meeting with them or they are no longer available.
     </Alert>
 );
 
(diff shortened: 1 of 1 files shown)


src/assets/src/components/common.tsx
```diff
@@ -28,8 +28,7 @@ export const UserDisplay = (props: UserDisplayProps) =>
 
 export const userLoggedOnWarning = (
     <Alert variant='primary'>
-        <strong>Note:</strong> The person you want to add needs to have logged on to Remote Office Hours Queue
-        at least once in order to be added.
+        <strong>Note:</strong> The person you want to add needs to have logged on to Remote Office Hours Queue at least once in order to be added. If the queue is not empty, hosts will not be notified. Please manually remove attendees from queue after meeting with them or they are no longer available.
```
jonespm, line 31: Reading over this seems fine. I added @HemanginiShah incase she wants to review this at all.

I added a few additional words and rewrote the second line to avoid the double negative.

```suggestion
        <strong>Note:</strong>The person you want to add needs to have logged on to Remote Office Hours Queue at least once in order to be added. Hosts will only receive notifications if the queue is empty. Please manually remove attendees from the queue after meeting with them or if they are no longer available.
```

PR #552, opened 2024-10-17
Fixes #454 Create packages and builds on Github

### LINKED ISSUE #454: Create packages and builds on Github
Now that we can use Github for storing packages, we should use the same pattern as created for MyLA to store packages in this repository. This would help for the releases and also reduce the chance of any build problems on Openshift. 

This workflow can probably be modified to support this repository. https://github.com/tl-its-umich-edu/my-learning-analytics/tree/master/.github/workflows

jonespm:
This was modeled after MyLA. There aren't branches here like there is in MyLA, just master. 

I tested the build process for master and that creates that package fine. I've also tested the tags.

# Testing

Once this is merged it will be straightforward to test.

If you'd like to checkout this branch and test it on your repository before merging, you'd have to update the value of `tl-its-umich-edu` to your user name and merge or cherry-pick this change to your repositories `master` for it to trigger a build. It will not do anything in this branch. 

Afterward you'd need to do a hard reset to your fork's master branch to clean all of this up.

Assuming your remote is `origin ` and tl-its-umich-edu is `upstream` for instance for me I have this

```
$ git remote -v
origin	https://github.com/jonespm/remote-office-hours-queue (fetch)
origin	https://github.com/jonespm/remote-office-hours-queue (push)
upstream	https://github.com/tl-its-umich-edu/remote-office-hours-queue (fetch)
upstream	https://github.com/tl-its-umich-edu/remote-office-hours-queue (push)
```

Ensure that your master from upstream is updated. 

Then you'd run
`git reset --hard upstream/master`

Then push it

`git push origin master --force-with-lease`

You can't force push onto the wrong repository, it will give you an error.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,62 @@
+name: ROHQ Build/Release
+
+on:
+  push:
+    branches: 
+      - master
+ 
+# Used to release a new version
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version'
+        required: true
+        default: 'MAJOR.MINOR.MICRO'
+    description: 'Manually trigger the workflow to create a new version tag and draft a release. This must be run from a tag that is pre-created.'
+          
+
+jobs:
+  build:
+    # to test a feature, change the repo name to your github id. I'm not sure of a better way to check this.
+    if: github.repository_owner == 'tl-its-umich-edu' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+  
+      - name: Checkout code 
+        uses: actions/checkout@v4
+    
+      - name: Extract branch name
+        id: extract_branch
+        run: echo "BRANCH_NAME=$(basename ${{ github.ref }})" >> $GITHUB_ENV
+      
+      - name: build Docker image
+        run: |
+          docker build ./src --tag ghcr.io/${{ github.repository }}:${BRANCH_NAME}
+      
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,62 @@
+name: ROHQ Build/Release
+
+on:
+  push:
```
zqian, line 4: The [build file inside MyLA repository,](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/.github/workflows/main.yml) include tag push. Do we need to include the tags here for ROHQ?

zqian (approved):
I have cherry-picked the change to my fork, and the Github action was invoked upon checkin.

```
git cherry-pick fffaeefc20707e8bb826bcd1baafad9957b00f3f
git push
```

 I left a minor comment.

.github/workflows/main.yml
jonespm: I removed them because I don't _think_ the `on: push: tags:` is actually necessary for having this work. We would manually trigger a release for that tag and don't need it to automatically trigger. It created a tagged package as expected for my. I don't think it's needed for MyLA either and just duplicates the manual release. We should never be pushing changes to a tag. 

If it _was_ needed it would require a little research as the tag format currently for ROHQ is different and uses semver rather than calver.

zqian: Make sense. No need to include the tags trigger.

Issue #512, opened 2024-06-20
Broken Zoom meetings in QA: "Access token is expired"


jaydonkrooss:
Example Queue:
[broken queue in QA](https://qa.officehours.it.umich.edu/manage/486) (you'll need to be added as admin to try this example)

I'm not sure if zoom doesn't work across the board, but the way I've replicated this issue:

1. Be the queue manager of a Zoom-enabled queue in [QA](https://qa.officehours.it.umich.edu/) env.
2. Add a user as an attendee for a Zoom meeting
3. Select "Create Meeting" --> causes an error. In the UI the error banner says "Start Meeting:" 

These two errors pop up in the logs:
> `ERROR:root:Unsuccessful request to https://api.zoom.us/v2/users/me/meetings: [401] {'code': 124, 'message': 'Access token is expired.'}`
> `ERROR 2024-06-20 15:28:26,604 Internal Server Error: /api/meetings/4749/start`
 
> `pyzoom.err.NotAuthorized: {'code': 124, 'message': 'Access token is expired.'}`
>` ERROR 2024-06-20 15:28:27,511 Internal Server Error: /api/meetings/4749/start`

After these errors, the logs show a bunch of Header parsing issues as a result of the invalid access token. Not sure if our Zoom credentials need to be re-evaluated, but once we confirm the cause, we may also want to handle this API failure and give the UI a proper error to display to the user.

jonespm:
Probably related to #466, #481. Not sure yet where that problem might be.

jonespm:
We talked on the meeting and likely we'll revert #481 for 1.10 and fix it for 1.11.

jonespm:
This should be fixed by #513. We'll want to check this as a case when we re-merge #466.

zqian:
Verified on ROHQ QA that queue manager can create Zoom meeting successfully without the "Access token is expired" error.

jonespm:
I re-opened this. 

To test this will likely require a dev.

You should be able to do it in the /admin interface or on the database. On Admin, go to a user you want to test and change the value of `backend_metadata`. It will have a "access_token_expires" in the JSON. Just need to change this to be earlier. By default this is set for 1 hour so subtracting 3600 or more seconds will trigger the refresh. 

`"access_token_expires": 1729628939.4155385`

zqian:
I have followed the above instructions: 
- in qa ROHQ /admin page. invalidate the access_token_expires value to past date, 
[...]

PR #553, opened 2024-10-22
Adding support for Zoom pyzoom package 

### LINKED ISSUE #466: Switch Zoom code over to use the package
The Zoom code we have here is all custom and may be difficult to maintain in the future. There is a [well maintained package in PyPi](https://pypi.org/project/pyzoom/) that we may want to consider using if we want to make any changes to this.

jonespm:
Fixes #466 and #512


zqian (approved):
the code change looks consistent with prior zoom settings, and added an error handling for Zoom API token refresh error. We may go ahead merge the change and test it on ROHQ test server.

PR #554, opened 2024-10-22
Fixes #488 - Adding env sample files

### LINKED ISSUE #488: Implement environment file configs (and update documentation)
In its current state, the app is in need of a way to extract environment variables set by developers when needed for integrations like Zoom, Twilio, and BlueJeans. Currently, running the app locally can only be done without these settings. 

Make sure to reference a .env file in the Docker build (like how we have it on other projects such as MyLA). The Readme instructions ought to be updated to account for the user to create their own .env file as well.

Dev Test plan: try to add credentials for an integration using .env file (Twilio would probably be the easiest to check). And verify the integration is functional.

jonespm:
Leaving as draft while a few more values are researched and tested.


.env.sample
```diff
@@ -0,0 +1,22 @@
+#DEBUG=False
```
zqian, line 1: There are a couple of environment attributes defined in the docker-compose.yaml file, inside both web and database services. Can those be moved into the .env too?

lsloan: Good question.  The master branch currently has…

```
      - ALLOWED_HOSTS=*
      - DEBUG=True
      - DATABASE_URL=postgresql://admin:admin_pw@database/admin
      - FEEDBACK_EMAIL=office-hours-devs@umich.edu
```

As I was working with the app in local dev., I had to set some values there for running the Twilio tests.  I was thinking at the time that they should go in a config file as Matt does here.  I think the values listed above could also go there.

docker-compose.yaml
```diff
@@ -10,6 +10,8 @@ services:
       - DEBUG=True
       - DATABASE_URL=postgresql://admin:admin_pw@database/admin
       - FEEDBACK_EMAIL=office-hours-devs@umich.edu
```
lsloan, line 12: As suggested in earlier comments, let's move these to `.env`, too.

jonespm: Sure we can move these, but I think having these might *require* that a .env file exists unless these are in the code. That's probably fine as a practice. I'd also update the README, I was still testing this to make sure there was nothing else and having it as draft.

This would only affect development but is really useful to have for development.

.env.sample
jonespm: Okay I added these.

docker-compose.yaml
jonespm: I added these and found a way to do fallback

jonespm:
I found out that docker-compose supports variable fallback/interpolation. So I made that change to the docker-compose to put the defaults but they can be overridden by .env. I tested that this works. 

https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/

.env.sample
```diff
@@ -0,0 +1,39 @@
...
```
zqian, line 1: Since the .env.sample file is added in this PR, maybe put an instruction line inside the README file accordingly? Something like "Copy the .env.sample file as .env and set values for those environment variables".

zqian (approved):

PR #556, opened 2024-10-22
Fixes #555 use get_random_secret_key

### LINKED ISSUE #555: Use get_random_secret_key rather than a hardcoded secret
The code for development is using a hardcoded secret key. This should be using `get_random_secret_key` from

`from django.core.management.utils import get_random_secret_key
`

jonespm:
Noticed this when I was looking at #488. Prod uses hardcoded secrets and doesn't rely on this but this should be updated.


zqian (approved):
Nice catch!

lsloan (approved):
That's a simple solution.

PR #562, opened 2024-10-24
Fixes #460 - Change qa hostnames to test

### LINKED ISSUE #460: Change qa hostnames to test
In #456 many changes were made to switch qa->test however the hostnames still have qa in there. We should update that in the next release and completely align this to test. 

Will require requesting 2 new domain names and adding these to the service templates.

jonespm:
These are the updated names that need to be run on test


zqian (approved):
The change looks straightforward to me. Will test once it is deployed to test server

Issue #532, opened 2024-08-12
Add a new column to display how long user was in queue for


jonespm:
It was suggested it would be nice to see how long the attendees have been waiting in the queue. The "Join Info" button has a date when they joined but it doesn't show on the main page and needs to be calculated I'm thinking something like this as a mockup but could be changed.

![image](https://github.com/user-attachments/assets/91e3e62d-834d-4465-bdd1-45e3c630b65a)

```
# Calculate the difference
time_difference = time2 - time1

# Extract days, hours, and minutes from the difference
days = time_difference.days
hours, remainder = divmod(time_difference.seconds, 3600)
minutes, _ = divmod(remainder, 60)

# Pluralization logic
def pluralize(value, unit):
    return f"{value} {unit}{'s' if value != 1 else ''}"

# Display the difference
days_str = pluralize(days, "day")
hours_str = pluralize(hours, "hour")
minutes_str = pluralize(minutes, "minute")

print(f"{days_str}<br>{hours_str}<br>{minutes_str}")
```

HemanginiShah:
Lets go with this format:  “23h 47m 2s”

PR #563, opened 2024-10-25
#532 added display of Time in Queue

zqian:


src/assets/src/components/meetingTables.tsx
```diff
@@ -98,6 +98,15 @@ const InvalidMeeting = (props: MeetingEditorProps) => {
 
 }
 
+// format the time in seconds to hours, minutes, and seconds
+function formatTimeInSeconds(timeInSeconds: number): string {
+    const hours = Math.floor(timeInSeconds / 3600);
+    const minutes = Math.floor((timeInSeconds % 3600) / 60);
+    const seconds = timeInSeconds % 60;
+
+    return `${hours} hours ${minutes} minutes ${seconds} seconds`;
```
jonespm, line 107: Probably _minor_ but having the (s) in the plural in a parenthesis would probably be more appropriate.

`    return `${hours} hour(s) ${minutes} minute(s) ${seconds} second(s)`;
`
It could probably handle the special case in logic but I don't know if that's necessary. 

This was GPT generated

```
    function formatUnit(value: number, unit: string): string {
        return `${value} ${unit}${value !== 1 ? 's' : ''}`;
    }

    return `${formatUnit(hours, 'hour')} ` +
           `${formatUnit(minutes, 'minute')} ` +
           `${formatUnit(seconds, 'second')}`;
```

jonespm:
This seems like a good idea having it in the frontend.

Is it able to update the time dynamically like as new component with a useEffect hook that will update every second? I think that would be more useful since this is an SPA and they might not navigate away.

src/assets/src/components/meetingTables.tsx
lsloan: Or use an ISO 8601 kind of format…

```
return `${hours}h${minutes}m${seconds}s`
```

Or simple timestamp style.  I think people are used to seeing this format in video timestamps…

```
return `${hours}:${minutes}:${seconds} (h:m:s)`
```

Could add a "legend" just in case.

lsloan (commented):
I think @jonespm suggestion is helpful.  You have a few choices of format to avoid the plural time units.

src/assets/src/components/meetingTables.tsx
zqian: I have implemented the string change with the unit pluralized if the value is greater than 1

zqian: and also added the refresh in every 5 seconds

src/assets/src/components/meetingTables.tsx
```diff
@@ -140,11 +155,22 @@ function UnstartedMeetingEditor (props: UnstartedMeetingEditorProps) {
                 </>
             );
 
+        // calculate the time different from now to the time user joined the queue
+        const [secondsInQueue, setSecondsInQueue] = useState(Math.floor((new Date().getTime() - new Date(props.meeting.created_at).getTime()) / 1000));
+
+        useEffect(() => {
+            const intervalId = setInterval(() => {
+                setSecondsInQueue(Math.floor((new Date().getTime() - new Date(props.meeting.created_at).getTime()) / 1000));
+            }, 5000);
```
jonespm, line 164: I think it looks fine going to 1000. There's not going to be any load on the server. I'm fine either way, it looks good!

src/assets/src/components/meetingTables.tsx
jonespm: Not to keep nitpicking on this but I kind of like Lance's suggestion. Maybe compact it down to just '#h #m #s' and then we don't have to worry about the plural? Maybe Hema has a suggestion?

I think it's fine but having '0 hour' I think i incorrect and that needs to be plural or dropped like

`if (value == 0) return '';`

HemanginiShah:
Let's go with this format: “23h 47m 2s”

jonespm (approved):
I think this looks good!
image

In a future release we may also want to add this timer to the visitors view somewhere around here to make it feel more dynamic but not for this release.
image

Issue #568, opened 2024-10-28
cannot connect to localhost:8003 link 


zqian:
The problem was introduced in the recent [docker-compose.yaml change](https://github.com/tl-its-umich-edu/remote-office-hours-queue/commit/b84dad20e13245795624dc7d5378b72e483969e1), where the option `--wait-for-client` should not be included in the command line

PR #569, opened 2024-10-28
#568 remove the redundant option

zqian:


jonespm (approved):

jonespm:
This starts up as expected marking as done. Having this parameter in there made is so you had to connect a debugging client before the server would start, but now it starts and you can connect anytime.

PR #572, opened 2024-10-31
Fixes #460 - Change qa hostnames to test

### LINKED ISSUE #460: Change qa hostnames to test

jonespm:
Fixes #460 Fix reverted in 1.11.0 because we needed more work.

### DIFF
--- a/service/overlays/test/hostname.yaml
+++ b/service/overlays/test/hostname.yaml
@@ -1,6 +1,6 @@
 - op: add
   path: /spec/rules/0/host
-  value: qa.officehours.it.umich.edu
+  value: test.officehours.it.umich.edu
 - op: add
   path: /spec/rules/1/host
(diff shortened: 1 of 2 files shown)


zqian:
How can we test the change before actually deploying to the ROHQ test project on OpenShift?

jonespm:
There is no way to test, other than to maybe create a duplicate project and apply it there. It's easier to rollback when if needed.

It was rolled back after it didn't work before because of a SAML issue which I didn't resolve and should look into before merging this.

zqian (approved):
Prove the PR so that it can be tested on the OpenShift server directly.

PR #576, opened 2024-11-01
use correct Zoom settings (iss. #575)

### LINKED ISSUE #575: Zoom meetings created with wrong settings
Zoom meetings are being created without permission for users to join via phone audio only (dial-in).  The problem comes from [lines 145–154 of /src/officehours_api/backends/zoom.py](https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/50bd2ac/src/officehours_api/backends/zoom.py#L145-L154)…

```python
        # Specify the meeting settings
        # We are initializing the meeting settings with the default settings
        # and then modifying the settings we want to change
        meeting_settings = ZoomMeetingSettings.default_settings()
        meeting_settings.waiting_room = True
        meeting_settings.join_before_host = False
        meeting_settings.meeting_authentication = False
        meeting_settings.use_pmi = False
        meeting_settings.contact_name = user.first_name + ' ' + user.last_name
        meeting_settings.contact_email = user.email
```

The `pyzoom` module's default setting for `audio` is `voip`.  The API documentation explains the values of this setting as…

```
audio
enum
Determine how participants can join the audio portion of the meeting.
both - Both Telephony and VoIP.
telephony - Telephony only.
voip - VoIP only.
thirdParty - Third party audio conference.
Default: both
Allowed: both┃telephony┃voip┃thirdParty
```

We need to use the `both` value for the setting or possibly no value at all and let Zoom use the default, which also is `both`.

We also need to check the other setting values obtained through `default_settings()`.

lsloan:
Fixes #575.

Zoom API reference docs:
* https://developers.zoom.us/docs/api/
* https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/meetingCreate

### DIFF
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -142,7 +142,7 @@ class Backend(BackendBase):
         """Creates a Zoom meeting for the given user."""
         client = cls._get_client(user)
         meeting_settings = ZoomMeetingSettings(
-            approval_type=0,
+            approval_type=2,
             audio='both',
             auto_recording='none',
             cn_meeting=False,
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -146,11 +146,12 @@ class Backend(BackendBase):
         # We are initializing the meeting settings with the default settings
         # and then modifying the settings we want to change
         meeting_settings = ZoomMeetingSettings.default_settings()
+        meeting_settings.audio='both'
         meeting_settings.waiting_room = True
         meeting_settings.join_before_host = False
         meeting_settings.meeting_authentication = False
         meeting_settings.use_pmi = False
-        meeting_settings.contact_name = user.first_name + ' ' + user.last_name
+        meeting_settings.contact_name = user.get_full_name()
         meeting_settings.contact_email = user.email
         # invoke the create_meeting method of the ZoomClient instance
         try:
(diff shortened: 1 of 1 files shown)


lsloan:
Although I marked this PR as "ready for review", we still should check other values in `meeting_settings` to make sure there aren't any other settings that will cause us trouble.

zqian (approved):
tested the Zoom dial in and it works now!

lsloan:
I changed the code to use explicit settings, not defaults from `pyzoom`.  The settings seem to be correct for the UMich environment.

src/officehours_api/backends/zoom.py
```diff
@@ -141,17 +141,25 @@ def _get_client(cls, user: User) -> ZoomClient:
...
-        # and then modifying the settings we want to change
-        meeting_settings = ZoomMeetingSettings.default_settings()
-        meeting_settings.waiting_room = True
-        meeting_settings.join_before_host = False
-        meeting_settings.meeting_authentication = False
-        meeting_settings.use_pmi = False
-        meeting_settings.contact_name = user.first_name + ' ' + user.last_name
-        meeting_settings.contact_email = user.email
+        meeting_settings = ZoomMeetingSettings(
+            approval_type=0,
```
jonespm, line 145: Okay, I compared this to the previous settings and [zoom defaults](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/meetingCreate)

This should be changed and gets it in sync. I think everything else looks fine.

```suggestion
            approval_type=2,
```

jonespm (dismissed):

src/officehours_api/backends/zoom.py
lsloan: Thanks!  The documentation seems to indicate that would resolve the registration issue, too.

Issue #524, opened 2024-07-02
Investigate and implement additional fixes for prod service deployment


jonespm:
Things we need to investigate/fix for next release:

- [X] Need to turn off auto-deploy (d10049e)
- [X] Prod was still deploying from `officehours` project instead of `officehours-prod` (14a12dc)
- ~~[ ] We should probably deploy from a tag built in the test env rather than rebuild. Is there any easy way to keep this in sync?~~

TODO: Need to think about test/release plan for this change

lsloan:
@jonespm, can you elaborate on the 2nd and 3rd bullet items?

jonespm:
So 
2) All of our other projects deploy from an imagestream in the test project. For this project you have to build and deploy from the prod project. 

I think that's because this is defined here in the overlay
https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/177caed90e68cfaf1f150d43d303e3e9f5529201/service/overlays/prod/deployment.yaml#L3

I haven't ran this kustomize on every deploy though, if nothing hasn't changed. However I'd also prefer it deployed from a tag rather than latest so maybe this would need to be updated and ran every-time unless there's some other way to do it. 

So maybe when the next tag is getting ready this might have to be changed to something like

`value: image-registry.openshift-image-registry.svc:5000/officehours-test/officehours:1.11.0`

Ideally this would reference an image built on Github via #454 also like our other apps rather than just built on Openshift. 

3) I think that's referring to this like that still says `officehours`, and should be `officehours-prod`. 
https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/177caed90e68cfaf1f150d43d303e3e9f5529201/service/overlays/prod/deployment.yaml#L12

This is what it shows on the test overlay https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/service/overlays/test/deployment.yaml

So I think I might have had to manually change this but if this kustomize file is ever run on prod again it will mess it up.

jonespm:
For the idea of replacing the tag, GPT suggested from this prompt

> In kustomize if I have this can I pull the value latest from a secret file instead of hardcoding it?
> 
> - op: replace
>   path: /spec/template/spec/containers/0/image
>   value: image-registry.openshift-image-registry.svc:5000/officehours-test/officehours:latest
> 


adding this to kustomization.yaml

```
vars:
  - name: IMAGE_TAG
    objref:
      kind: Secret
[...]

PR #579, opened 2024-11-07
Need to also build from officehours-prod in #524

jonespm:


zqian (approved):
The problem was discovered during the ROHQ release. Hotfix applied.

Issue #582, opened 2024-11-07
Broken INTERNAL link triggered whenever someone is removed from a meeting


jonespm:
Whenever a user is removed from a meeting by the host by clicking the trash can an email is triggered to us. This is because code in the api doesn't have a trailing slash for the `removeMeeting `and it's getting a 301 redirect. I'm not sure why a 301 redirect is causing us to get emails though. 

So there's 2 problems here:

1) Fix the `removeMeeting `fetch to have a trailing slash. This should avoid the 301 entirely

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/777a0addf41806380d24694ea2d059dafb58ffec/src/assets/src/services/api.ts#L164

2) Figure out why 301's (URL's without a trailing slash) are internally being treated as 404s.

https://github.com/zqian/remote-office-hours-queue/blob/064ae0cd2ce33f42c9d51efa8c961c9f2628cd09/src/officehours_api/middleware.py

I found this [old issue](https://code.djangoproject.com/ticket/25971) that was fixed 9 years ago in Django that kind of relates to this, but I didn't see it in in Django 3 before we upgraded to Django 4. 

To testing: Email's aren't sent for this locally, however there will be a log message from [middleware.py](https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/src/officehours_api/middleware.py) stating "`Super class BrokenLinkEmailsMiddleware handling 404 broken link`". This URL pattern is allowed explicitly by this and could possibly also be denied, however when it checks 
`isinstance(response, HttpResponseNotFound):` then response matches for this 301. 

On the test server, emails are sent out.

lsloan:
One possibility for resolving the problem with trailing slash would be to change the pattern for the API endpoint(s).

[src/officehours_api/routing.py](https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/777a0ad/src/officehours_api/routing.py#L5-L8), lines 5–8:
```
websocket_urlpatterns = [
    re_path(r'ws/queues/(?P<queue_id>\w+)/$', consumers.QueueConsumer.as_asgi()),
    re_path(r'ws/users/(?P<user_id>\w+)/$', consumers.UserConsumer.as_asgi()),
]
```

It requires a slash, but if that could be made optional, then maybe the redirect wouldn't be required.  If the style of regular expression used here is similar in syntax to common REs, then maybe replace `/$` with `/+$`, indicating one optional slash.

[...]

PR #585, opened 2024-11-12
Remove trailing slash from API endpoints (iss. #582)

lsloan:
After this change, requesting endpoints with a trailing slash results in a HTTP 404 (not found) error.  Following this change, it's important to check other parts of the application to ensure they do not use a trailing slash when calling API endpoints.


jonespm:
Closing as #595 was merged

PR #587, opened 2024-12-06
Fixes #559: No newline in EMAIL_SUBJECT_PREFIX.

### LINKED ISSUE #559: BadHeaderError if EMAIL_SUBJECT_PREFIX has newlines
If EMAIL_SUBJECT_PREFIX contains newlines in the config there's an error like this one which makes debugging impossible.

`django.core.mail.message.BadHeaderError: Header values can't contain newlines (got '[TEST_ROHQ] \nERROR (EXTERNAL IP): Internal Server Error: `

While we can change this config, the code should ensure that this is value has newlines removed.

jadfalaoui:
Added an extra assert statement to ensure EMAIL_SUBJECT_PREFIX doesn't contain a newline character. I made sure to stay consistent with the existing tests for redis_host as mentioned in #473.

### DIFF
--- a/src/officehours/settings.py
+++ b/src/officehours/settings.py
@@ -296,8 +296,8 @@ LOGGING_METHODS = csv_to_list(
 
 
 # Email
-EMAIL_HOST = os.getenv('EMAIL_HOST')
-EMAIL_SUBJECT_PREFIX = os.getenv('EMAIL_SUBJECT_PREFIX', '[ROHQ] ')
+EMAIL_HOST = os.getenv('EMAIL_HOST').strip()
+EMAIL_SUBJECT_PREFIX = os.getenv('EMAIL_SUBJECT_PREFIX', '[ROHQ] ').strip()
 
 ADMINS = [('Admins', os.getenv('ADMIN_EMAIL'))]
 MANAGERS = ADMINS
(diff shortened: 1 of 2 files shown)


jonespm:
Thanks @jadfalaoui. This looks like it just fixes the test for the issue but the issue isn't fixed. If the EMAIL_SUBJECT_PREFIX contains new lines this test will fail. 

So this should also be fixed in https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/src/officehours/settings.py to prevent this from ever being an issue regardless if it has newlines or not.

jonespm (changes requested):
Requested changes as this just fixes the test, not the code.

jadfalaoui:
Hi @jonespm, thank you for bringing it to my attention. I added it to the settings.py file to prevent this from being an issue.

jonespm (approved):
Okay this seems good, thanks for the contribution!

jadfalaoui:
Thank you for the quick feedback!

zqian:
@jadfalaoui I got the following error with `docker compose up`, after this PR merge:

"remote-office-hours-queue-web-1              |   File "/usr/src/app/officehours/settings.py", line 299, in <module>
remote-office-hours-queue-web-1              |     EMAIL_HOST = os.getenv('EMAIL_HOST').strip()
remote-office-hours-queue-web-1              | AttributeError: 'NoneType' object has no attribute 'strip'"

Neither `EMAIL_HOST` or `EMAIL_SUBJECT_PREFIX` is defined as environment variable inside the docker-compose.yaml file. 

Could you please fix the above error?

hanzheg:
Got the same error with zqian

Issue #521, opened 2024-06-27
Provide a way for admins to export everything


jonespm:
It was requested that admins should be able to get a full export of the view. This may need to be something built in as a custom admin interface. I'm not sure the effort around this or if it would be worth putting in that time to complete this task as currently devs can export CSV's right from the view on the database.

Related to this was a way for department chairs and managers to export all of the reports for employees they manage. I'm not sure how that would be implemented and/or if it would be similar to this. 

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/97#issuecomment-2195296577_

Some additional implementation details, these may be subtasks.

1. Refactor the database logic currently in `ExportMeetingStartLogs.get` either into the models.py `MeetingStartLogsView` proxy or into a separate utils file.
2. Rewrite this to be less memory intensive, avoid a `fetchall` and instead have a loop that does a `fetchmany`
3. Have a method parameter for whether or not to do the filter by user. (To make it work for admins) Or maybe have it pass in the courses to filter by or a None to do no filtering. 
4. Call this method from where you are defining the admin action with no restrictions as well as in the view with restrictions.

jonespm:
Sure this might be a little harder one but you're welcome to try it. I think my idea was that we'd add a new custom action to the Django admin for queues. Right now there's 3 options, Delete, Hard Delete and Soft Delete. You'd be able to check the ones you also want to export and it would run it through the export code and provide an export. 

![Image](https://github.com/user-attachments/assets/3df3c8e0-2c24-4658-bd08-998f057312e0)

jonespm:
I'm thinking maybe something like this for the refactored model method

Then you'd pass like 
```
    response = HttpResponse(content_type='text/csv')
    response['Content-Disposition'] = f'attachment; filename="{filename}"'
    writer = csv.writer(response)
    MeetingStartLogs.export_to_csv(queues_user_is_in or None, writer)
```


```
class MeetingStartLogs(models.Model):
    @classmethod
    def export_to_csv(cls, queue_ids, writer):
        with connection.cursor() as cursor:
            if queue_ids:
                # Filter by specified queue IDs
                queue_id_placeholders = ', '.join(['%s'] * len(queue_ids))
                query = f"SELECT * FROM meeting_start_logs WHERE queue_id IN ({queue_id_placeholders})"
                params = queue_ids
            else:
                # No filtering by queue_id
[...]

PR #588, opened 2024-12-06
#521 Add Export to CSV for Queue Details and Enhance Admin Configuration

hanzheg:
We add a feature to export selected Queue details to a CSV file, including associated Meeting and Attendee data, improving administrative functionality.

### DIFF
--- a/src/officehours_api/admin.py
+++ b/src/officehours_api/admin.py
@@ -50,7 +50,7 @@ class QueueAdmin(SafeDeleteAdmin):
     list_display = ('id', highlight_deleted, 'created_at', 'status') + SafeDeleteAdmin.list_display
     list_filter = ('hosts', 'status',) + SafeDeleteAdmin.list_filter
     search_fields = ['id', 'name']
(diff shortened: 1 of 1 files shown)


jonespm:
Thanks for the contribution but I don't think this is really what we were looking for here.

There's an existing method that was added to the views.py for [ExportMeetingStartLogs](https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/79a4309079bd186aa86be5cad4221402ec238b10/src/officehours_api/views.py#L299). 

To fix this one for admin I'd probably do a few things that I should have mentioned on the issue.
1. Probably refactor the database logic from that method either into the models.py `MeetingStartLogsView `proxy or into a separate utils file.
2. Rewrite this to be less memory intensive, avoid a `fetchall `and instead have a loop that does a `fetchmany`
3. Have a parameter for whether or not to do the filter by user. (To make it work for admins)
4. Call this method from where you are defining the admin action. 

I can add some more detail to the ticket, some of these might be subtasks.

PR #594, opened 2025-01-10
591: use Google Analytics module for OneTrust privacy banner

jaydonkrooss:


jonespm:
Thanks for contributing this!

I was noticing that `oneTrustScriptDomain` is hardcoded and currently it is a fixed value for Ann Arbor. We should _probably_ make this configurable like `ga_tracking_id` is. I'm fine if you want to just open that as a new issue rather than resolving it. 

I was cleaning up the env in #554 so that may need to be committed first so it's documented there.

jaydonkrooss:
> Thanks for contributing this!
> 
> I was noticing that `oneTrustScriptDomain` is hardcoded and currently it is a fixed value for Ann Arbor. We should _probably_ make this configurable like `ga_tracking_id` is. I'm fine if you want to just open that as a new issue rather than resolving it.
> 
> I was cleaning up the env in #554 so that may need to be committed first so it's documented there.

Would adding it as a secret in the new .env require a kustomize process to be run? I remember discussing this with @lsloan a while back and he created issue #523 to address it.

If .env is a good workaround without too much trouble I'm fine with waiting for that PR to merge first before adding it

src/assets/src/containers/app.tsx
```diff
@@ -11,14 +11,19 @@ import { QueuePage } from "../components/queue";
...
 
 interface AppProps {
     globals: Globals;
 }
 
 export function App(props: AppProps) {
-    useGoogleAnalytics(props.globals.ga_tracking_id, props.globals.debug);
+    useGoogleAnalytics({
+        googleAnalyticsId: props.globals.ga_tracking_id, 
+        debug: props.globals.debug,
```
jonespm, line 24: What does this package do when debug is turned on? I couldn't  find that in the README for that project but might be useful to document there. Does it log anything to the console or do anything?

jonespm:
Getting this error this this trying it out, were you seeing this one? I'm not sure what it means yet as <App> is in a <Router> component. 

```
router.js:241 Uncaught Error: useLocation() may be used only in the context of a <Router> component.
    at App (app.tsx:22:23)
```

jonespm:
So it *seems* like this might be an issue more with the upgrade of react and 2 different versions of react-router-dom being loaded? It looks like both 6.25.1 and 6.28.1 are added? 

I added this to package.json and that seems to get it loaded. Do you know if this is the best way? It looks like onetrust *might* be the only one that has an old version to updating that might also fix this, but possible save to have this in here? 

```
  "overrides": {
    "react-router-dom": "^6.28.1"
  },
```

src/assets/src/containers/app.tsx
jaydonkrooss: From what I've seen, the debug prop was inherited from the first implementation of google analytics, apparently it just enables `testMode` for the googleAnalytics instance, which is [useful only for testing event calls](https://github.com/react-ga/react-ga?tab=readme-ov-file#test-mode). I think this prop should be renamed to testMode and we utilize `debug` for logging, like you said.

I created an issue to address this in the project https://github.com/tl-its-umich-edu/react-ga-onetrust-consent/issues/7

jaydonkrooss:
> So it _seems_ like this might be an issue more with the upgrade of react and 2 different versions of react-router-dom being loaded? It looks like both 6.25.1 and 6.28.1 are added?
> 
> I added this to package.json and that seems to get it loaded. Do you know if this is the best way? It looks like [onetrust](https://github.com/tl-its-umich-edu/react-ga-onetrust-consent/blob/26a8fa825f298cb522d9d1ab867adefd511eebb8/package.json#L39) _might_ be the only one that has an old version to updating that might also fix this, but possible save to have this in here?
> 
> ```
>   "overrides": {
>     "react-router-dom": "^6.28.1"
>   },
> ```

I think updating it in the module may make the most sense in this case, no need to put in an override as it might lead to dependency issues in the future. Open to other suggestions but I'll just go ahead with that patch as version 1.1.2 on react-ga-onetrust-consent

zqian (approved):
Successfully tested with the test plan and TunnelBear VPN settings.

PR #595, opened 2025-01-10
Add a trailing slash to the delete request as needed now in Django

### LINKED ISSUE #582: Broken INTERNAL link triggered whenever someone is removed from a meeting

jonespm:
Fixes #582

I feel like this fix would have less chance of regression than changing the backend to be completely without trailing slashes in #585. 

I don't think that would work unless the front-end was also changed to match? This seems to be a recent change in Django 4.2 that caused this.


jonespm:
I also added a trailing slash to the /start/ request which was the only one I saw that didn't have one.

zqian (approved):
Tested with the following steps:

1. repeat the problem with source code without this PR: add a user to a meeting, then try to remove this user from the newly added meeting. Verified the delete generated the error log message: "Super class BrokenLinkEmailsMiddleware handling 404 broken link".

2. do the same steps with the PR in placce, and found no error log files, as show in above.

PR #597, opened 2025-01-14
Draft - WIP Changing form validation, evaluating use of *

### LINKED ISSUE #437: Evaluate use of `*` as required indicator
I know this is just adding a space here but it looks like using the HTML form validation is recommended over an * both for usage and accessibility. 

https://www.accessibility-developer-guide.com/examples/forms/required/#using-html-5-client-side-validations
https://react-bootstrap.netlify.app/docs/forms/validation

Was this not changed just because that's a separate (more effort) issue? We should probably file it as a new issue to fix it in the future.

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/435#discussion_r1234323441_

jonespm:
Fixes #437

### DIFF
--- a/src/assets/src/components/PhoneVerification.tsx
+++ b/src/assets/src/components/PhoneVerification.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import { useState } from "react";
 import { FormStatus } from "./preferences";
 import { validatePhoneNumber } from "../validation";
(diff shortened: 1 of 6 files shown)


jonespm:
@jaydonkrooss Not sure if you have any ideas on the frontend patterns here or not. It might just be some legacy that was used years ago. I think there's more to cleanup here.

Issue #593, opened 2025-01-06
Update Django and related dependencies


jonespm:
We should upgrade any dependencies manually and also add in a dependabot.yaml file similar to one used in the Instructor Tools Project. This might need the upgrade for Python to also be completed in #592 

https://github.com/tl-its-umich-edu/instructor-tools/blob/main/.github/dependabot.yml

jonespm:
Django is the latest version, need to add a dependabot.yaml file still.

PR #598, opened 2025-01-14
Upgrade Python 3.10 to 3.13

### LINKED ISSUE #592: Upgrade Python and Dependencies up to 3.13
Currently this project is Python 3.10. We should upgrade this and all the dependencies up to 3.13.

jonespm:
Fixes #592

Ran the tests, read the upgrade notes and did some basic testing, looks like there wasn't any breaking issues here.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -12,7 +12,7 @@ RUN npm run prod
 
 #####################################################################
 
-FROM python:3.10-slim-bullseye
+FROM python:3.13-slim-bullseye
 
 ENV PYTHONUNBUFFERED=1
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1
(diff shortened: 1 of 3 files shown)


zqian (approved):
The upgrade works for me. Good to take the pip install libraries out from Dockerfile.

jonespm:
Thanks! Yeah, I'm not sure why they were in the Dockerfile in the first place. With them here it will allow Dependabot to be include them in vulnerabilities scans and updates too.

jonespm:
I plan to also fix #593 tomorrow so dependabot works correctly with the grouped updates and holding back the Django version.

zqian:
Correct Python version is shown in server log file: "/usr/local/lib/python3.13/site-packages/django/http/response.py"

PR #601, opened 2025-01-15
Provide a way to get Github info into build image

### LINKED ISSUE #600: Provide a way for Github information to be added to Docker image
> Sorry, I didn't see this comment back in December. So it seems like there's to solutions to this.
> 
> 1. Change the context to be one directory up while building so that the `.git` directory can be accessed during the build.
> 2. Pass in an argument for this value when building. Locally I think this would require either defining this version in the .env file but a pre-build script in Github actions could automate this. 
> 
> I think I'll look into doing 1 since this is how most of our other apps are then if that doesn't work try 2. The value will be present in the environment though as `GIT_COMMIT` which could be pushed to the footer.  

 _Originally posted by @jonespm in [#567](https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/567#issuecomment-2593233180)_

jonespm:
Fixes #600

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
       
       - name: build Docker image
         run: |
-          docker build ./src --tag ghcr.io/${{ github.repository }}:${BRANCH_NAME}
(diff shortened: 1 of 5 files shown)


jonespm:
I'll add a commit to update the github action here before merging this as that context changes.

PR #602, opened 2025-01-16
issue_567 add github version info to the footer

zqian:
I made an update to the /etc/git.version content format, so that it contains only name=value pairs.

### DIFF
--- a/src/docker-entrypoint.sh
+++ b/src/docker-entrypoint.sh
@@ -2,6 +2,11 @@
 
 set -e
 
+# Load in git version information
+source /etc/git.version
+
+echo "Loaded Git information: ${GIT_REPO} ${GIT_BRANCH} ${GIT_COMMIT}"
+
 python manage.py collectstatic --noinput
 python manage.py migrate
 
--- a/src/officehours/settings.py
+++ b/src/officehours/settings.py
@@ -142,6 +142,9 @@ MIDDLEWARE = [
 
 ROOT_URLCONF = 'officehours.urls'
 
+# Git info settings
+SHA_ABBREV_LENGTH = 7
+
 TEMPLATES = [
     {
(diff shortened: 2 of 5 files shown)


jonespm:
It was in that format there so it could be loaded in the environment and pulled with `os.getenv`. This might make it simpler and more efficient than reading and parsing the file in Python every time. 

https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1463/files#diff-65a5db1ae8f8e01ebdbbfafcf4e24618ad7cffe036bba92dce1390383e5c73e5

Another idea would be to just create it as a regular Python file and directly import it to skip all the parsing?

In git_version_info.sh

```
cat <<EOF > /usr/src/app/officehours/git_version.py
# Auto-generated file containing Git version information
version = {
    "repo": "$GIT_REPO",
    "commit": "$GIT_COMMIT",
    "branch": "$GIT_BRANCH"
}
EOF
```

Then in settings.py

```
from git_version import version

# Example usage:
GIT_REPO = version.get("repo")
GIT_COMMIT = version.get("commit")
GIT_BRANCH = version.get("branch")
```

zqian:
"It was in that format there so it could be loaded in the environment and pulled with os.getenv" Yes, MyLA calls start.sh to set the env values. What is the corresponding script used in ROHQ?

jonespm:
ROHQ loads https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/src/docker-entrypoint.sh as the starting script.

zqian:
Great! I have updated the code to read from environment variables instead.

jonespm:
Thanks this looks better, I'll run this branch this afternoon and approve, am working on a different issue currently.

jonespm (approved):
I think this looks good. Tested with admin and regular user and with the footer that's present. It shows them both and it doesn't show this information for a regular user.

PR #604, opened 2025-01-21
Adding dependabot.yml file for auto updates

### LINKED ISSUE #593: Update Django and related dependencies

jonespm:
Fixes #593

This file mostly exists so that grouped updates can work as expected and django is held back to 4.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,29 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


jonespm:
It's close but won't ever be completely identical because of the different project structure. 

Github does support having a template repository and creating new projects from that, but I don't know what other files or folder structure might be common. Agreed though that for updates to work correctly we'll need this file in every repository to hold back from major automatic updates.

PR #605, opened 2025-01-21
Update log format to include X-Forwarded-For

### LINKED ISSUE #359: Update log format to include X-Forwarded-For to capture users IP in logs

jonespm:
Fixes #359

Draft for the moment while I test a few more things locally

### DIFF
--- a/src/officehours/logging_config.py
+++ b/src/officehours/logging_config.py
@@ -1,0 +1,22 @@
+import logging
+
+class CustomLogFormatter(logging.Formatter):
+    def format(self, record):
+        record.http_status = getattr(record, 'http_status', '-')
+        record.x_forwarded_for = getattr(record, 'x_forwarded_for', '-')
+        record.request_line = getattr(record, 'request_line', '-')
+        record.response_time = getattr(record, 'response_time', '-')
+        return super().format(record)
+
+    log_format = '%(asctime)s [HTTP:%(http_status)s] [%(x_forwarded_for)s] "%(request_line)s" %(response_time)s'
+
+def setup_logger():
+    log_format = '%(asctime)s [HTTP:%(http_status)s] [%(x_forwarded_for)s] "%(request_line)s" %(response_time)s'
+    logger = logging.getLogger('custom_logger')
+    logger.setLevel(logging.DEBUG)
+
+    console_handler = logging.StreamHandler()
+    console_handler.setFormatter(CustomLogFormatter(log_format))
+    logger.addHandler(console_handler)
+
+    return logger
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -31,7 +31,9 @@ COPY src/requirements.txt /tmp
 RUN pip install -r /tmp/requirements.txt --no-cache-dir
 
 ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
-CMD ["sh", "-c", "gunicorn officehours.asgi:application --bind=0.0.0.0:8001 -k uvicorn.workers.UvicornWorker"]
+
+CMD ["sh", "-c", "gunicorn officehours.asgi:application --bind=0.0.0.0:8001 -k uvicorn.workers.UvicornWorker --access-logfile - --access-logformat='%(t)s [HTTP:%(s)s] [%({x-forwarded-for}i)s] \"%(r)s\" %(L)s \"%(r)s\" %(s)s %(b)s %({Referer}i)s %({User-Agent}i)s'"]
+
 
 COPY --from=node-webpack /usr/src/app/package-lock.json package-lock.json
 COPY --from=node-webpack /usr/src/app/webpack-stats.json webpack-stats.json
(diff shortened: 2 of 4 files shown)


jonespm:
I think this is ready for review. Running this locally should give you logs for both access and server that are in a similar format.

jonespm:
Deploying to prod will need to get the new secret FORWARDED_ALLOW_IPS as only some IP's are allowed to be forwarded.

zqian:
Is there any changes needed in the .env file? I got " Error: Got unexpected extra arguments (# Trust all IPs in development for testing)" and the app is not up running

jonespm:
No, it doesn't need any locally as * is in the docker-compose, it just looks like it's pulling the comments into the array, I'll move that 😁

docker-compose.yaml
```diff
@@ -20,7 +20,11 @@ services:
       - ./src:/usr/src/app:z
     command: >
       python -m debugpy --listen 0.0.0.0:5678
-      -m uvicorn officehours.asgi:application --host=0.0.0.0 --port=8001 --reload
+      -m uvicorn officehours.asgi:application 
+      --host=0.0.0.0 
+      --port=8001 
+      --reload 
+      --forwarded-allow-ips='*'  # Trust all IPs in development for testing
```
jonespm, line 27:
```suggestion
      --forwarded-allow-ips='*'

File: service/base/web-deployment.yaml [hunk 2]
```diff
@@ -23,7 +23,16 @@ spec:
...
-        image: docker-registry.default.svc:5000/officehours-dev/officehours:latest
+        image: image-registry.openshift-image-registry.svc:5000/officehours-dev/officehours:latest
+        command: ["python"]
+        args: [
+          "-m",
+          "uvicorn",
+          "officehours.asgi:application",
+          "--host=0.0.0.0",
+          "--port=8001",
+          "--forwarded-allow-ips=${FORWARDED_ALLOW_IPS}"  # Will be set via environment variable
```
jonespm, line 34:
```suggestion
          "--forwarded-allow-ips=${FORWARDED_ALLOW_IPS}" 
```

src/officehours/logging_config.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def format(self, record):
+        if len(getattr(record, 'args', [])) == 5:
+            # HTTP request from uvicorn
+            client_addr, method, path, http_version, status = record.args
+            record.client = client_addr
+            record.request = f"{method} {path} HTTP/{http_version}"
+            record.status = status
+        else:
+            # Regular Python log
```
zqian, line 21: I wonder whether the regular Python log can be of any usage. Currently I only see patterns of:

web-1              | 2025-01-23 14:03:29,760 [DEBUG] [-] "% received keepalive pong" 760ms
web-1              | 2025-01-23 14:03:49,764 [DEBUG] [-] "% sending keepalive ping" 764ms
web-1              | 2025-01-23 14:03:49,764 [DEBUG] [-] "> %s" 764ms
web-1              | 2025-01-23 14:03:49,766 [DEBUG] [-] "< %s" 766ms
web-1              | 2025-01-23 14:02:29,752 [INFO] [-] "connection open" 752ms
web-1              | 2025-01-23 14:02:29,752 [DEBUG] [-] "= connection is OPEN" 752ms

zqian (approved):
I left a minor comment

src/officehours/logging_config.py
jonespm: Without this patch current format looks like

```
[2025-01-23 19:37:09 -0500] [12] [INFO] ('1.2.3.4', 53268) - "WebSocket /ws/users/59784/" [accepted]
[2025-01-23 19:37:09 -0500] [12] [INFO] connection open
[2025-01-23 19:37:18 -0500] [12] [INFO] connection closed
```

These noisy DEBUG messages here shouldn't be displayed if "DEBUG=False". You could test this. 

So new consistent format in this patch for Python and access logs is 
`2025-01-23 14:02:29,752 [INFO] [-] "connection open" 752ms`

The [-] would have an IP address if it was from the web with the number being like 200 or 404. 

Which is similar but not identicalto the old format
`[2025-01-23 19:37:09 -0500] [12] [INFO] connection open`

Happy to make additional adjustments if we want. 

Some changes - 

- The time looks like it was changed to remove the TZ information
- The time it took to execute the request is attached now
- The message is quoted
- There is no process ID

lsloan (approved):
I like these changes.  I'd like to suggest that before merging, make sure **_you're_** happy with the log format, too.  If possible, make the information in the line as compact as possible if you want to leave more line space available for the actual log message.  In the timestamps, removing TZ info helps.  Maybe removing the milliseconds would help, too, if the server doesn't seem busy enough to warrant it.

src/officehours/logging_config.py
zqian: No need to worry about the debug messages!

PR #608, opened 2025-01-23
use test imagestream for deployment (iss. #561)

### LINKED ISSUE #561: prod. deploy from GitHub package
As discussed in #524, this issue is broken out…

> We should probably deploy from a tag built in the test env rather than rebuild. Is there any easy way to keep this in sync

As @jonespm suggests in the first comment below, rather than deploy from an imagestream in the test version of this project in OpenShift, the prod. project can be deployed from the package in this GitHub repo.  The packaging was enabled as a result of #454.

lsloan:
Resolves #561.

Update production overlay to refer to test imagestream tag, defined in the test environment.

### DIFF
--- a/service/overlays/prod/deployment.yaml
+++ b/service/overlays/prod/deployment.yaml
@@ -1,6 +1,6 @@
 - op: replace
   path: /spec/template/spec/containers/0/image
-  value: image-registry.openshift-image-registry.svc:5000/officehours-prod/officehours:latest
+  value: image-registry.openshift-image-registry.svc:5000/officehours-test/officehours:latest
 - op: replace
   path: /spec/selector/variant
   value: prod


lsloan:
@jonespm: I changed the reference to the imagestream in the prod overlay as seen here.  When I run kustomize locally and look at the resulting YAML, it appears correct in the `DeploymentConfig-officehours-web-prod.yaml` generated.

However, later in that same DC file, there is still a trigger for `officehours-prod`.  I think it would be a mistake to change that to `officehours-test`, because it would probably cause a production redeployment each time the test imagestream is updated.  So, I _**think**_ that trigger should probably be removed.  Does that seem like the right way to go?

jonespm:
1. Yes, we don't want a trigger on prod
2. I was hoping we'd have a ImageStream in here that would reference ghcr.io/tl-its-umich-edu/remote-office-hours-queue:master rather than the build. If weren't not going to be able to do that this release, [because of this change](https://github.com/tl-its-umich-edu/remote-office-hours-queue/commit/fdc5e60db6480de8da0e555645a5f59016ad045d) we need to update the build.yml to change the context and the Dockerfile to be something like

```
source:
  contextDir: .
  dockerfile: src/Dockerfile
```

3. What is the service/out/run.sh file doing? If this is something useful could you document it in the README in the service directory?
4. (Eventually?) I'd like to modify the Github action so that during the build and creating the image it would also modify the version in the overlay. This might be a separate task that can be in the next release and we've been doing these steps manually for each release. (Tagging the image in Openshift, and switching the prod DeploymentConfig over to using this tagged image)

lsloan:
1. I've removed the trigger just now.
2. OK.  I remember the earlier comments about using an imagestream that references ghcr.io.  I'll work on that.
3. Yes, `out/run.sh` is a convenient script to run the steps in `service/README.md`.  I'll update that document to explain the script.
4. The other day we talked about this issue, we decided that dynamically updating the version was probably a little too complex to accomplish in this release.

lsloan:
@jonespm, this is ready for review.  I've updated Kustomize to make an imagestream referring to ghcr.io and the DC uses that IS.

jonespm (approved):
This all looks good to me, I haven't tried to run it yet. Have you ran/applied this on the dev instance yet? I can run it tomorrow. 

We're still waiting on the change for the domain name to update test.

lsloan:
Thanks for the approval, @jonespm.  I'll merge the PR.

No, I've not run it on the dev server yet.  I wasn't sure how I could restore our working config there if the new config were incorrect.

jonespm:
Ideally, we should just be able to check out the previous tag of the code and deploy with that to rollback. But I haven't tried that in awhile.

PR #609, opened 2025-01-23
Move the get_backend_types logic out of model to avoid migrations.

### LINKED ISSUE #574: Avoid triggering migrations for get_backend_types
In the model there's this code for `choices` this is both in the Queue and Meetings. This results in a new migration whenever `get_backend_types()` changes. This is dynamic and cannot be predicted     

```
   backend_type = models.CharField(
        max_length=20,
        choices=get_backend_types(),
        null=False,
        default=get_default_backend,
    )
```

and

```
    allowed_backends = ArrayField(
        models.CharField(max_length=20, choices=get_backend_types(), blank=False),
        default=get_default_allowed_backends,
    )
```

This shouldn't be in the model because it makes it think a migration is missing. Instead we probably should move the choices completely out of the model and into the API.

Test Plan:
---
- Start without Zoom configuration, note that you can only create In Person meetings on a Queue. 
- Check the logs (`docker compose logs web`) , Verify no migration is needing to be run
---
- Add configuration to the .env for Zoom, note that you can create both Zoom and In Person on a Queue.
- Check the logs (`docker compose logs web`) , Verify no migration is needing to be run

Edit the queue and ensure you can select different backend types. 
Verify you can create a meeting with both typs.

jonespm:
Fixes #574


jonespm:
Leaving this as draft for the moment as I'm still testing this. I think this feels like a better pattern, moving this out of models and into the serlializer. This avoids migrations.

Test plan on the ticket.

src/officehours_api/models.py
```diff
@@ -168,6 +174,10 @@ def change_backend_type(self, new_backend_name: Optional[IMPLEMENTED_BACKEND_NAM
             # Dependent on queue.allowed_backends remaining not nullable.
             self.backend_type = self.queue.allowed_backends[0]
 
+    def clean(self):
```
jonespm, line 177: Remove these clean methods here

src/officehours_api/serializers.py
```diff
@@ -228,6 +228,13 @@ def validate_host_ids(self, host_ids):
         else:
             return host_ids
 
+    def validate_allowed_backends(self, value):
```
zqian, line 231: I cannot find the usage of this new validate_allowed_backends() or the validate_backend_type() function in the code base.

zqian: @jonespm Is the new validate_allowed_backends() used anywhere in the code?

jonespm: This is a [special DRF pattern](https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation
) for `validate_<field_name>` so they're used from the library. 

I can see if I can provide a unit test or test case for it to validate that it's working and useful. I did check in a debugger that it was being called but all of the values were always valid.

zqian (approved):
Test successfully with the test plan in #574.

PR #614, opened 2025-01-28
Update dependabot.yml (restrict react updates)

jaydonkrooss:
After testing on my branch, this should work to prevent react updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,6 +31,9 @@ updates:
       - dependency-name: "react"
         versions:
           - ">=19.0.0"
+      - dependency-name: "react-dom"
+        versions:
+          - ">=19.0.0"
       - dependency-name: "@types/react"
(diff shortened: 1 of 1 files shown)


jonespm:
Do you think we should also manually hold back `webpack `and `typescript`? It looks like it's bumping typescript from 4 to 5. 

I saw this article for this syntax

https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/
https://stackoverflow.com/questions/60649404/how-can-i-change-my-dependabot-config-to-exclude-major-versions

It looks like you might be able to also use `version-update:semver-major`. Had you tried this option? 

> Ignore updates to one or more semantic versioning levels. Supported values: version-update:semver-minor, version-update:semver-patch, and version-update:semver-major.

jonespm (approved):

PR #617, opened 2025-01-30
Fixes #561 - Updates to service files

### LINKED ISSUE #561: prod. deploy from GitHub package

jonespm:
1. The name on the package in https://github.com/tl-its-umich-edu/remote-office-hours-queue/pkgs/container/remote-office-hours-queue was master, not latest so updated that
2. Needed to change image names to the full value new Imagestream of mage-registry.openshift-image-registry.svc:5000/officehours-dev/officehours-rohq-latest-at-ghcr-test

### DIFF
--- a/service/base/imagestream.yaml
+++ b/service/base/imagestream.yaml
@@ -7,6 +7,6 @@ spec:
     - name: latest
       from:
         kind: DockerImage
-        name: ghcr.io/tl-its-umich-edu/remote-office-hours-queue:latest
+        name: ghcr.io/tl-its-umich-edu/remote-office-hours-queue:master
       importPolicy:
         scheduled: true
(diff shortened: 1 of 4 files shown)


lsloan (approved):
Thank you for making those corrections.  I misunderstood the references in some of the config files.  What you have here looks better, but I'll need to study it more to understand how it works.

PR #620, opened 2025-02-03
implement admin meeting data export (iss. #521)

### LINKED ISSUE #521: Provide a way for admins to export everything

lsloan:
Resolves #521.

This feature allows the Django admin UI to download data about meetings as CSV files, to support troubleshooting.  The meeting data to be downloaded is based on queues.  Queues may be selected in one of two ways:

1. In the Queues table (`/admin/officehours_api/queue/`), tick the checkboxes of one or more queues in the list, choose "Export meeting data for selection" from the "Action" menu, then activate the "Go" button.  All meeting data about those queues is downloaded as a CSV.
2. In the Users table (`/admin/auth/user/`), tick the checkboxes of one or more users in the list, choose "Export meeting data for selection" from the "Action" menu, then activate the "Go" button.  Queues are found in the DB by searching for those which have the selected user(s) as hosts.  If queues are found, then all meeting data about those queues, including data about other hosts, is downloaded as a CSV.  If the selected user(s) are not hosts in any queues, an error message is shown.

The CSV files generated are named `meeting_data.csv`.  It was not feasible to name the files as specifically as they are when a non-admin user requests downloads from the regular UI.  Because the admin may choose data by Queue or User and it may be one or multiple of those, it could be excessively complicated to make specific filenames for such a simple feature.


.gitignore
```diff
@@ -70,3 +70,4 @@ db.sqlite3
 secret
 node_modules
 webpack-stats.json
+service/out/*
```
lsloan, line 73: This line was not committed in my previous PR, so I'm adding it here.  It's unrelated to this issue.

src/officehours_api/admin.py
```diff
@@ -1,17 +1,78 @@
...
 from officehours_api.models import Queue, Meeting, Attendee, Profile
+from officehours_api.views import ExportMeetingStartLogs
+
+logger = logging.getLogger(__name__)
+
+admin.site.site_title = admin.site.site_header = 'ROHQ Admin'
+admin.site.index_title = 'Home'
+
+
+class ExporterAdminMixin:
```
lsloan, line 18: It felt to me like this class is the Python "mixin" style.  Maybe I'm using the term incorrectly.  If this doesn't seem like a "mixin", tell me and I will rename the class.

lsloan:
@jonespm & @jaydonkrooss:  This is ready for review.  If you have any questions about the review or with testing, ask me and I'll be happy to help.

jonespm:
It seems like many changes here are just formatting changes. Was this autoformatted by some code formatter you have in your IDE? 

I was thinking of doing a one pass re-format possibly using black and adding a [.git-blame-ignore-revs.](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) but it makes it harder to figure out when formatting is combined with functionality and can't add this rev to ignore. 

I can't exactly tell what formatter this might be, because it's changing single quotes to double quotes but leaving some single quotes in there. It also looks like short line length. 

It just makes it hard to review the actual work when also having to think about the formatting changes.

jaydonkrooss:
I agree that the extra formatting can be removed, or at least may work better in a different PR. The changes themselves appear to work, locally tested the admin page for both users and queues selection.

lsloan:
Some of the formatting here was done by the reformat feature in my IDE.  I use it mostly because it cleans up `import` statements and fixes spacing in the code I contributed.  However, there is the downside that it could also affect indentation in other parts of the code.  

**_If the whitespace changes in `src/officehours_api/views.py` are too distracting, I can undo them._**

The quoting changes were done by me, for two reasons…

1. In sections of code where I was working, I changed double-quoted strings to single, because that's the way I would write them.  PEP 8 doesn't make recommendations for quotes.  I use singles because I don't need to press the Shift key while typing them and also, it's the kind of quote that Python's `repr()` uses…
    ```sh
    $ python -c 'print(repr("fubar"))'
    'fubar'
    ```
    
2. There were docstrings written with `'''`, which is illegal according to PEP 257 (referred by PEP 8), so I changed them.

Some of the quote and whitespace changes didn't **_NEED_** to be done, but I've received critiques in past code reviews about not fixing some of these things, even though I didn't write them.  So I fix these things if I think they seem to be in pretty close proximity to the part I'm coding.

jonespm:
I personally don't care about formatting but I've always just left what's there untouched and configured my IDE to not adjust reformatting code that doesn't have functional changes. Much of this PR, especially in views.py, seems like it's reformatting without function changes.

I'd prefer to automatically enforce a format for the entire project via Github Actions with something like `black` to avoid spending time and mental energy on this. If you change something, then another developer has different settings (like line length) it will just keep changing back and forth. Most likely `black` would apply and agree with the changes you made and I'd agree that this project could be completely reformatted. 

If we bulk reformat in one commit, having that in the `.git-blame-ignore-revs` would be great to keep the history easier to follow.  Looking at a diff of `black --diff .` it _is_ very similar to this. It seems the most contentious issue for black is they have an 88 character default line length, which I'm fine and the output of this seems very readable. I know none of our projects use this but the some I've worked on that do seem fine. 

Black also changes the `'''`->`"""` but there's `'''` used in other files and would be _best_ if they were all changed together on a single commit. Black seems to change almost all single quotes over to double.

lsloan:
@jonespm and @jaydonkrooss:  I've undone most of the reformatting in `src/officehours_api/views.py` by getting a patch from my commit and applying it in reverse.  I kept some changes, like the `import` cleanup and docstring quotes.

I agree that automated formatting in the repo would be helpful.  It would be less to think about.

Please review the PR again.  If there are any other problems, formatting or otherwise, tell me and I'll fix it.

Thanks!

src/officehours_api/views.py
```diff
@@ -2,36 +2,41 @@
...
 from officehours_api.models import Attendee, Meeting, Queue
-from officehours_api.serializers import (
-    ShallowUserSerializer, MyUserSerializer, ShallowQueueSerializer, QueueAttendeeSerializer,
-    QueueHostSerializer, MeetingSerializer, AttendeeSerializer, PhoneOTPSerializer
-)
-from officehours_api.permissions import (
-    IsAssignee, IsHostOrReadOnly, IsHostOrAttendee, is_host
-)
+from officehours_api.notifications import send_one_time_password
+from officehours_api.permissions import (IsAssignee, IsHostOrReadOnly,
```
jonespm, line 27: Formatting nitpick, but these imports should probably be on separate lines like the import below is. For instance

```
from officehours_api.permissions import (IsAssignee, 
                                         IsHostOrReadOnly,
                                         IsHostOrAttendee, 
                                         is_host)
```

jonespm:
Thanks this is improved. I'll try to run it before the end of day. 

I'd be fine creating a Issue to use `black` if that is what we agree on and running through a formatter on this project after this is merged which should help out some of this. We can discuss this at the meeting on Monday.

src/officehours_api/views.py
lsloan: Yeah, perhaps.

The `import` formatting was done entirely by the IDE.  I didn't change anything there after running the reformatter.  My guess is that its goal isn't to put them on individual lines, but to put them in lines where the left side is aligned with the open parenthesis and the right side is no longer than the line length limit (80 characters).

Note that in the import you referenced below, line 34 has `MeetingSerializer, AttendeeSerializer,`, 2 symbols on a line, not only 1.  That's because they're short enough to fit together within the line length limit.

jaydonkrooss (approved):
Looks good to me. Local testing works the same

lsloan:
Thanks for reviewing, @jaydonkrooss! 🙇

src/officehours_api/views.py
```diff
@@ -324,23 +329,22 @@ def get(self, request, queue_id=None, format=None):
...
+                'SELECT * FROM meeting_start_logs '
+                f'where queue_id in ({queue_ids})')
             rows = cursor.fetchall()
 
-            # Get column names from the cursor description
-            column_names = [desc[0] for desc in cursor.description]
-
-            # Write the headers
-            writer.writerow(column_names)
+            column_names = map(lambda λ: λ[0], cursor.description)
```
jonespm, line 348: I was reviewing through this and don't know why this change was made unless it was just trying to be cute or done by AI. What was there before was fine and this just makes it less maintainable by introducing a unicode character into a function as a variable.

PR #631, opened 2025-03-03
630 upgrade cookie consent react module 1.2.0

jaydonkrooss:
upgrade will require to run kustomize to remove the old onetrust domain script variable


zqian (approved):
The change looks good to me. Banner showed up when l changed the location to EU using TunnelBear. Vote for merge and test the GA on ROHQ test server

PR #633, opened 2025-03-06
add queue name and status to meeting report (iss. #520)

### LINKED ISSUE #520: Include queue name & Status columns in report download
It was requested that the queue name and Status columns be included in the download report. This may require a join against the queue table as I'm not sure if it's in the log table.

We also wanted to include the deleted date in this report so this would probably be resolved at the same time as #516

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/97#issuecomment-2195296577_

Here is the list of items requested by Anne Rickert, along with a few additional ones from me.

- Add Queue name to views as well as reporting download feature and Admin feature for report download. NOTE: First she asked for Meeting description but then she came back in another email and said meeting **name.**
- Add Queue Status column to views as well as reporting download feature and Admin feature for report download
- ROHQ "meeting_start_logs' view seems to be producing multiple identical records in some cases.
- Another point she made: - The "meeting_start_logs" view also seems to have several cases of the exact same host/attendee/meeting_created_at, but different meeting_url values. That seems odd since the meeting_created_at times are very specific.

lsloan:
Resolves #520.

### DIFF
--- a/src/officehours_api/views.py
+++ b/src/officehours_api/views.py
@@ -2,6 +2,7 @@ import csv
 import logging
 from datetime import datetime, timezone, timedelta
 from random import randint
+from typing import List
 
 from asgiref.sync import async_to_sync
 from django.conf import settings
@@ -335,14 +336,20 @@ class ExportMeetingStartLogs(APIView):
         return response
 
     @staticmethod
-    def extract_log(queues, response):
+    def extract_log(queue_ids: List[int], response: HttpResponse) -> None:
         writer = csv.writer(response)
         with connection.cursor() as cursor:
-            queue_ids = ', '.join(map(str, queues))
-
             cursor.execute(
-                'SELECT * FROM meeting_start_logs '
-                f'where queue_id in ({queue_ids})')
+                f'''SELECT
+                  meeting_start_logs.*,
+                  officehours_api_queue.name AS queue_name,
+                  officehours_api_queue.status AS queue_status
+                FROM
+                  meeting_start_logs
+                JOIN officehours_api_queue ON
+                  meeting_start_logs.queue_id = officehours_api_queue.id
+                WHERE queue_id IN ({', '.join(map(str, queue_ids))})'''
+            )
             rows = cursor.fetchall()
 
             column_names = map(lambda λ: λ[0], cursor.description)


jonespm:
Hi @lsloan ! I didn't notice you put this in. Because it looks like we're going to be giving direct access to the database, this change needs to happen against the database **view** rather than the Django view. This view was originally added via a migration in https://github.com/tl-its-umich-edu/remote-office-hours-queue/pull/501 and the intent is that we would modify this. 

My guess is you'll need another manual file migration. The only change needed to `view.py` would probably be that improved type hinting, since the new columns would be present. I'm not entirely sure the change needed but I think you can use a `CREATE OR REPLACE VIEW` here and have to duplicate basically the entire view inserting two new fields and doing a join

I think we can also pull out the `queue_id `into the `parsed_response` to make it cleaner.

I have not tested this but seems like it should work or be close based on what you have here.

```
from django.db import migrations

class Migration(migrations.Migration):

    dependencies = [
        ('officehours_api', '0028_alter_meeting_backend_type_and_more'),
    ]

    operations = [
        migrations.RunSQL(
            sql="""
                CREATE OR REPLACE VIEW meeting_start_logs AS
                WITH parsed_response AS (
                    SELECT
                        response::jsonb AS response,
                        (response::jsonb -> 'queue')::int AS queue_id
                    FROM
                        rest_framework_tracking_apirequestlog
                    WHERE
                        view::text = 'officehours_api.views.MeetingStart'
                        AND (response::jsonb ->> 'created_at') IS NOT NULL
                )
                SELECT
                    pr.queue_id AS queue_id,
                    (response -> 'attendees' -> 0 ->> 'id')::int AS attendee_id,
                    (response -> 'attendees' -> 0 ->> 'user_id')::int AS attendee_user_id,
                    response -> 'attendees' -> 0 ->> 'username' AS attendee_uniqname,
                    response -> 'attendees' -> 0 ->> 'last_name' AS attendee_last_name,
                    response -> 'attendees' -> 0 ->> 'first_name' AS attendee_first_name,
                    (response -> 'assignee' ->> 'id')::int AS host_id,
                    response -> 'assignee' ->> 'username' AS host_uniqname,
                    response -> 'assignee' ->> 'last_name' AS host_last_name,
                    response -> 'assignee' ->> 'first_name' AS host_first_name,
                    response -> 'backend_type' AS meeting_type,
                    response -> 'backend_metadata' ->> 'meeting_url' AS meeting_url,
                    response -> 'agenda' AS agenda,
                    to_timestamp(response::jsonb ->> 'created_at'::text, 'YYYY-MM-DD"T"HH24:MI:SS.US'::text) AS meeting_created_at,
                    q.name AS queue_name,
                    q.status AS queue_status
                FROM
                    parsed_response pr
                JOIN officehours_api_queue q ON pr.queue_id = q.id
            """,
            reverse_sql="""
                DROP VIEW IF EXISTS meeting_start_logs;
            """
        ),
    ]
```

jonespm (changes requested):
Requested changes to modify the database view ⬆️ rather than the views.py so that queries can be run directly against this view.

lsloan:
@jonespm, I've tested this in my dev. environment and it appears to work well.  This is ready for another review and I think it's ready for merging.  I made changes based on email from Anne Rickert and one recommended for more secure queries.

HemanginiShah:
Here is the list of items requested by Anne Rickert, along with a few additional ones from me.

1. Add **Queue name** to views as well as reporting download feature and Admin feature for report download
2. Add **Queue Status** column to views as well as reporting download feature and Admin feature for report download
3. ROHQ  "meeting_start_logs' view seems to be producing multiple identical records in some cases.
4. When in the future, we bring **Affiliation, Department, Major, Minor** columns to the report.

HemanginiShah:
Here is the list of items requested by Anne Rickert, along with a few additional ones from me.

Add Queue name to views as well as reporting download feature and Admin feature for report download
Add Queue Status column to views as well as reporting download feature and Admin feature for report download
ROHQ "meeting_start_logs' view seems to be producing multiple identical records in some cases.
When in the future, we bring Affiliation, Department, Major, Minor columns to the report.

HemanginiShah:
Please also check on the feedback provided by Anne Richert:

-- The "meeting_start_logs" view also seems to have several cases of the exact same host/attendee/meeting_created_at, but different meeting_url values. That seems odd since the meeting_created_at times are very specific.

lsloan:
Because @jonespm and I were wondering whether using `DISTINCT` in the view query could mask problems, I opened #656 to investigate why there are repeated lines from the log.  It turns out that the log has unique timestamps, it's just not in the JSON data queried by this view.  In that issue, @zqian suggested using `DISTINCT` in this view query although I was on the verge of removing it.

In a meeting on 02 Jun, Zhen explained that Anne's use of the view is for simple reporting and doesn't require highly accurate technical data.  That's why it's acceptable for the view query to use `DISTINCT` to remove the duplicate rows.  The other issue may stay open to investigate why some rows in the logs are (nearly?) identical except for timestamp, but for expediency, this PR is complete and #520 can be resolved.

PR #450, opened 2023-11-22
Added React-Helmet for titles and wrote tests

Nyu10:
Closes #436 

This PR adds dynamic page titles using React-Helmet.
We also added unit tests (jest) to ensure webpage loads the proper dynamic titles.
Unit test setup is a start to frontend test coverage. 

Note: we can remove the unit tests and make it another PR if neccessary
Screenshot 2023-11-22 at 12 20 38 PM

PR #636, opened 2025-03-14
635 cookie consent banner hotfix

jaydonkrooss:


jonespm (approved):
This appears to be working fine for me. Was Helmet required to get this working or just made it easier? We have this PR #450 where Helmet was planning on being added across the entire project but that was never completed.

jaydonkrooss:
> This appears to be working fine for me. Was Helmet required to get this working or just made it easier? We have this PR #450 where Helmet was planning on being added across the entire project but that was never completed.

It helped to load the script in React, as @pushyamig discovered. It was taking a while for me to do it the more "traditional" way with React effects. This way works, but I think it might not be necessary long term if we revisit doing it that way.

That Helmet PR looks fine to me, looks like it addresses a small accessibility problem. we could keep it in this project if we merge in that person's changes.

PR #647, opened 2025-05-16
feat: fixes #639, added watchman monitoring setup

### LINKED ISSUE #639: Add watchman monitoring set up
(Same content as the CCM watchman issue: https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/485 )

for projects without Watchman, Zabbix only monitors the base URL, checking for a 200 OK response. For applications that have Watchman implemented, the Zabbix web scenario is configured to monitor the Watchman URL. This provides an enhanced monitoring experience, as Watchman performs additional checks, such as database, cache, and storage validation. Additionally, Watchman’s functionality is extensible, offering potential for further custom checks as needed.

Instructor Tools has the sample implementation. https://github.com/tl-its-umich-edu/instructor-tools/blob/main/backend/settings.py#L267

PraveshKunwar:
Will fix issue #639 . This is based off of the implementation linked in the issue, and also based off of the [README.md](https://github.com/mwarkentin/django-watchman/blob/main/README.rst) from the django-watchman repo.

Not sure if the implementation is correct, but the /watchman/ returns the following screenshot and /status/ returns 200 OK.

Screenshot 2025-05-16 145915


pushyamig:
I will review this in the afternoon

src/officehours/settings.py
```diff
@@ -78,6 +78,10 @@ def str_to_bool(val):
     ]
 
 WATCHMAN_TOKENS = os.getenv('WATCHMAN_TOKENS')
+WATCHMAN_TOKEN_NAME = os.getenv('WATCHMAN_TOKEN_NAME', 'officehours-watchman-token')
+WATCHMAN_CHECKS = ('watchman.checks.caches', 'watchman.checks.databases')
+WATCHMAN_DISABLE_APM = True
+EXPOSE_WATCHMAN_VERSION = True
```
jonespm, line 84: This should also be configrable

src/officehours/settings.py
PraveshKunwar, line 1: leave out watchman_disable_apm and expose_watchman_version

PraveshKunwar:
commented out those configurations

src/officehours/settings.py
```diff
@@ -78,6 +78,11 @@ def str_to_bool(val):
     ]
 
 WATCHMAN_TOKENS = os.getenv('WATCHMAN_TOKENS')
+WATCHMAN_TOKEN_NAME = os.getenv('WATCHMAN_TOKEN_NAME', 'officehours-watchman-token')
+WATCHMAN_CHECKS = ('watchman.checks.caches', 'watchman.checks.databases')
```
pushyamig, line 82: I feel based on the Matt comment these should be configurable, I think add these 2 variables to `.env.sample`, uncomment here and make it default to `False`. So if anybody did not add it to `.env` the defaults are applied.

Something like this should work, 

config_to_bool = lambda value: str(value).lower() in ('true', '1', 'yes', 'on')
WATCHMAN_DISABLE_APM = config_to_bool(os.getenv('WATCHMAN_DISABLE_APM', False))

PraveshKunwar: alright, I think its ok now? thank you for the feedback.

pushyamig:
This change is working, i was able to run it locally. I will differ the approval @jonespm or @jaydonkrooss

jonespm (approved):
Looks good, this will be better for the monitoring.

PR #648, opened 2025-05-20
feat: handle twilio errors to resolve #491 (WIP)

### LINKED ISSUE #491: Alert user or require phone number validation on Twilio error
If a phone number gets back a failure response from Twilio then we should prevent any new messages from going to that number until the number is verified again. I think the phone verification will help some of this but seems like some errors like 30003 can happen even after verified. Verification should catch some others though

Common Twilio errors are
* 30003 - Unreachable destination handset
* 21211 - Invalid 'To' Phone Number
* 30006 - Landline or unreachable carrier

A sample workflow would be 
1) Twilio sends a error message
2) User in database is flagged in a new column as having a problem with their number
3) When user logs in they will see a banner indicating they have a problem with their phone and need to verify it in settings before receiving messages
4) ROHQ won't send future texts to problem numbers

It sounds like error 30003 might also just be temporary. So I'm not sure how common this is. I see about 50 30003 messages in the last 30 days. Maybe this isn't something that's high priority

PraveshKunwar:
this will fix issue #491 , WIP (will add a better description when done)

### DIFF
--- a/src/assets/src/components/preferences.tsx
+++ b/src/assets/src/components/preferences.tsx
@@ -2,11 +2,18 @@ import * as React from "react";
 import { useState, useEffect } from "react";
 import { Alert, Button, Form, FormGroup } from "react-bootstrap";
 import PhoneInput from "react-phone-input-2";
-import 'react-phone-input-2/lib/bootstrap.css'
+import "react-phone-input-2/lib/bootstrap.css";
 
 import * as api from "../services/api";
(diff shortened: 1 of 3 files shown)


PraveshKunwar:
will stop progress on this for a while, need a little more clarification before proceeding.

pushyamig:
@PraveshKunwar This is still a draft, but added a reviewer so do you want to open this for review or still WIP?

Issue #490, opened 2024-05-06
Messages in twilio notifications should have some indication they were sent from University of Michigan


jonespm:
In the [notifications](https://github.com/andcarls/remote-office-hours-queue/blob/master/src/officehours_api/notifications.py), there's a many text messages that are sent. 

These should have some indication they were sent from University of Michigan in the messages.

Ideally these could be pulled in from a configuration file.

PR #650, opened 2025-05-23
added UMich identifier to notifications (iss. #490)

PraveshKunwar:
Summary: Resolves issue #490 

This PR will add a UofM identifier in the notifications, so now when a user gets a notification sent to their phone number, it will look like:

[U-M Office Hours] Someone joined your queue example.com/manage/6/ (via Twilio logs)

Also, made this configurable so the identifier can be changed in the .env file.


lsloan (approved):
This is a very good effort, @PraveshKunwar.  I like that you made the identifier configurable (as specified in the issue requirements), a default identifier is used if it's not configured, and you've consolidated the formatting of the message body to a single function for re-usage.

:shipit:

Issue #410, opened 2023-05-12
Replace WebSocket client library


ssciolla:
> Not sure either about that reconnecting-websocket.

Thanks for weighing in @jonespm! I think dealing with this is somewhat important, since WebSocket connections seem to be one of the areas where things can go wrong with this application.

The libraries you mentioned are actually server WebSocket tools; `socket.io` comes with a client, but I don't think it's compatible with Channels websockets. This is relevant: https://socket.io/docs/v4/#what-socketio-is-not They do mention an alternative to `reconnecting-websocket` called `robust-websocket`, but it looks to be even more out of date. I'll have to look and see if Django Channels recommends any particular JavaScript client.

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/405#issuecomment-1536521848_

Good discussion about this on #405. `reconnecting-websocket` was a reasonable choice, but it is getting a bit old, is not being updated. @jonespm highlighted [`react-use-websocket`](https://www.npmjs.com/package/react-use-websocket) as an option and `sockjs-client` may be another. Essentially we are looking for an extension of native WebSocket behavior that also handles reconnecting and some other browsing irregularities.

jonespm:
I noticed today that https://www.npmjs.com/package/partysocket is a fixed replacement for reconnecting-websocket, so might be the easiest thing to use.

jonespm:
From @PraveshKunwar on the PR

Test Plan (what's been tested so far)

x is user in one incognito window, y is user in another incognito window

- [ ] Joining a queue (x joins y's queue) 
- [ ] Updating agenda (x adds topics they want to discuss with y) 
- [ ] y is ready for x (y clicks "ready for attendee" button, x gets notification that host is ready) 
- [ ] x cancels meeting with y 
- [ ] y adds x as an attendee for the queue via the "add attendee" button 
- [ ] x "leaves the line" for y's queue 
- [ ] y updates the queue name, changes on x's screen in realtime 
- [ ] x joins y's queue, y deletes queue while x is waiting, x gets notified 
- [ ] y deletes x from the queue

HemanginiShah:
Test Successful!

PR #655, opened 2025-05-28
feat: migrate WebSocket hook to use partysocket instead of reconnecting-websocket

PraveshKunwar:
Summary: This changes the useWebSocket hook to use partysocket instead of reconnecting-websocket.

This will resolve issue #410


PraveshKunwar:
this is ready for review

jonespm:
There isn't anything automated for tests. I suggested to Pravesh that to test it he probably should.

1) Start Office Hours on two different browsers as two different users
2) Try most of the functions like creating a meeting, having the user in the other window join the meeting and host start the meeting.
3) The application should dynamically update as actions are taken by the host and user. 

(I'd expand this test plan out to be more detailed, but this would be the basic idea of how it should be tested)

PraveshKunwar:
Test Plan (what's been tested so far)

x is user in one incognito window, y is user in another incognito window

1. Joining a queue (x joins y's queue) ✅
2. Updating agenda (x adds topics they want to discuss with y) ✅
3. y is ready for x (y clicks "ready for attendee" button, x gets notification that host is ready) ✅
4. x cancels meeting with y ✅
5. y adds x as an attendee for the queue via the "add attendee" button ✅
6. x "leaves the line" for y's queue ✅
7. y updates the queue name, changes on x's screen in realtime ✅
8. x joins y's queue, y deletes queue while x is waiting, x gets notified ✅
9. y deletes x from the queue ✅

These are all the things that I've tested, and they seem to work just fine. There might be more things that could be tested, but this was all that I could think of.

src/assets/src/hooks/useWebSocket.ts
```diff
@@ -1,75 +1,101 @@
...
+  const [error, setError] = useState(undefined as Error | undefined);
+  useEffect(() => {
+    const ws = new WebSocket(url, undefined, {
+      maxRetries: 10,
+    });
+    ws.onmessage = (e: MessageEvent) => {
+      const m = JSON.parse(e.data) as OfficeHoursMessage<T>;
+      console.log(m);
+      switch (m.type) {
+        case "init":
```
jonespm, line 25: These init and update perform exactly the same action. They can be combined to reduce redundancy

```
  case "init":
  case "update":
    onUpdate(m.content as T);
    break;
```

jonespm (approved):

PR #658, opened 2025-05-30
fix: removed passcode form from zoom meeting, not required anymore

### LINKED ISSUE #657: ROHQ not working with Zoom web app or devices without Zoom Workspace installed
It was reported by some users that ROHQ when launched with the Desktop app, either if they don't have Zoom Client installed or if they're using Chromebook or Linux, are unable to login. Zoom is asking for the passcode for meetings created by ROHQ. 

To reproduce:

* As a user, create a queue, then view as a visitor
* Join the meeting
* Manage the meeting and start the meeting as host
* Go back to view as vistor

You should now  have a link that goes to something like `https://umich.zoom.us/j/123456789` for that meeting. At the bottom of the page there is a link "Join from Browser". The pattern of this is `https://zoom.us/wc/join/123456789`. (The meeting ID goes on the end)

If you do this immediately it gives you a page that asks for Meeting Passcode. ROHQ is not sending or creating any passcode.

Image

There's probably some option that needs [to be set in Python](https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/930a52c3e3cb60e652c0f9c2c3289f5eb39abc98/src/officehours_api/backends/zoom.py#L141) to avoid prompting for this.

Testing out editing the ROHQ meeting in the Zoom Client, just saving it there causes the Passcode option to go away.
Image

(After editing and just saving the meeting in Zoom client)
Image

I'm not sure yet what's different, missing or making it ask for passcode. The [Zoom API](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/meetingCreate) has a number of options. My best guess might be to set `default_password=False` but if that doesn't change anything it could be some other setting there.

PraveshKunwar:
Summary: fixes #657 (has details on how to reproduce error aswell)

Fixes this issue by creating a "patched" work around, where we specify `password` to be `None` and `default_password=False`. I believe that in `pyzoom`, there is no option `default_password`, which would be the way to fix this error by setting it to `False`, and removing the `password` parameter as well.

So, I created a `patched_create_meeting` method as recommended by @jonespm to temporarily fix this error.

After making the changes, we get the below:

Screenshot 2025-05-30 151913

### DIFF
--- a/src/officehours_api/backends/zoom.py
+++ b/src/officehours_api/backends/zoom.py
@@ -10,6 +10,7 @@ from django.shortcuts import redirect
 
 from officehours_api.backends.backend_base import BackendBase
 from officehours_api.backends.types import IMPLEMENTED_BACKEND_NAME
+from officehours_api.patches import pyzoom_patch
 
 from pyzoom import ZoomClient
 from pyzoom.err import APIError as ZoomAPIError
@@ -24,7 +25,7 @@ class ZoomMeeting(TypedDict):
     id: str
     host_id: str
     topic: str
-    type: Literal[1, 2, 3, 4, 8]  # instant, scheduled, recurring/unfixed, PMI, recurring/fixed
+    type: Literal[1, 2, 3, 4, 8]
     status: str
     start_time: str
     duration: int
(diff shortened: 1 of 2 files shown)


jonespm (approved):
I've done some basic testing and the code seems fine for now until it's added to Zoom module. We can move on and test this in test.

lsloan:
I attempted testing on iPad.  I guess the group's preferred action is to comment in the issue, #657.

PR #659, opened 2025-05-30
fix: Update base images to latest veresioon (bookworm and node-20)

### LINKED ISSUE #583: Update base images to latest version (boomworm and node-20)
[Debian bullseye is EOL](https://endoflife.date/debian) and should be upgraded. Some additional backend dependencies may need to be updated or changed to accommodate this.

Also node-18 needs to be upgraded to node-20.

Both the Dockerfile and Dockerfile.openshift should be updated for now.

We also got an email that on June 28th these would be removed from the container service and we're still building there currently.

jonespm:
Fixes #583

This was straightforward. I tested that this built and is working.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,5 +1,5 @@
 # build react components for production mode
-FROM node:18-alpine AS node-webpack
+FROM node:20-alpine AS node-webpack
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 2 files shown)


PraveshKunwar (approved):
looks good

lsloan:
I found a typo.  `boomworm` should've been `bookworm`.  I fixed it.

Issue #637, opened 2025-03-14
May need to add configuration for CSRF_TRUSTED_ORIGINS


jonespm:
When testing on localhost with some proxies the CSRF is rejected. This doesn't appear to be an issue with prod but still might be convienent to add a configuration to set CSRF_TRUSTED_ORIGINS based on this: https://docs.djangoproject.com/en/5.1/ref/csrf/#how-it-works

This would mostly be seen if you are testing the Google Analytics code or something giving someone else access remotely. 

Until this is configured if you get a warning like

`CSRF verification failed. Request aborted.` then you have to add this to settings.py as

```
CSRF_TRUSTED_ORIGINS = os.getenv("CSRF_TRUSTED_ORIGINS", [])
```

jonespm:
For a test plan:

This seems to only be a problem with development using Loophole or possibly other proxies that don't forward headers correctly. This doesn't seem to be necessary with ngrok. 

So to test, start up loophole and try to login, you'll get this error. 
Configure the loophole domain you used as this setting 
`CSRF_TRUSTED_ORIGINS=https://loophole-test-domain-rohq.loophole.site`

Then try again and it will connect. We won't have to configure this for prod and there's no way I can think of to test on Openshift, but it is nice to have this incase we ever need to embed this or are doing development and this happens.

I've verified this setting works locally.

PR #661, opened 2025-06-02
fix(CSRF): add configuration for CSRF trusted origins

PraveshKunwar:
Summary: fixes issue #637 

Adds configuration for CSRF in settings.py


zqian:
Requested for a test plan in #637

jonespm (approved):
Okay this looks like it will be good with this default, thanks!

PR #670, opened 2025-06-13
Fix bug in queue data export (iss. #667)

### LINKED ISSUE #667: "Download All Queue History" causes error
hen you're on your queues page with at least a few queues, if you click the button on the bottom "Download All Queue History" there is an error displayed in the UI. This worked fine in the last release.

Image

We also get an email that says

```
ProgrammingError at /api/export_meeting_start_logs/
can't adapt type 'SafeDeleteQueryset'
```

that has a traceback. This would be a blocker for next release and one we need to look into.

lsloan:
Resolves #667.

The queue IDs may not always be a list of `int`.  It may be a `SafeDeleteQuerySet`, depending on which code calls the method.

### DIFF
--- a/src/officehours_api/views.py
+++ b/src/officehours_api/views.py
@@ -2,7 +2,6 @@ import csv
 import logging
 from datetime import datetime, timezone, timedelta
 from random import randint
-from typing import List
 
 from asgiref.sync import async_to_sync
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


src/officehours_api/views.py
```diff
@@ -336,13 +335,13 @@ def get(self, request, queue_id=None, format=None):
         return response
 
     @staticmethod
-    def extract_log(queue_ids: List[int], response: HttpResponse) -> None:
+    def extract_log(queue_ids, response: HttpResponse) -> None:
```
jonespm, line 338: I may leave in a type hint of either `Any `or probably even better `Iterable[int]`. `Iterable[int]` would cover this `QuerySet` that goes through a `.values_list` call.

jonespm (approved):
I ran this and it looks to fix the issue, will need more testing on test. 
Thanks.

PR #672, opened 2025-06-17
Update to use a shared-runner for tl repo

jonespm:


jonespm:
Chatting with Matthew Deeds, shared runners can't be used currently on Public repositories. Will close this for now and may revisit in the future if this changes or if we need shared runners.

PR #726, opened 2026-01-06
issue_675 added a What's New link

zqian:


jonespm:
I think ideally these 2 urls should come from the settings but seems fine if you want to hard code it for now. They probably won't change.

jonespm:
I think ideally these 2 urls should come from the settings but seems fine if you want to hard code it for now. They probably won't change.

zqian:
Cool! I have done further updates according to the latest comments. Ready for merge.

PR #729, opened 2026-01-08
issue_676 added an extra column  into the meeting_start_logs view

zqian:


lsloan (approved):
See my earlier comment on this PR WRT permissions lost when the view is dropped.  If that risk is acceptable, then this change is OK.

zqian:
I could not find a specific requester name on this feature from the project meeting notes so far. I guess we can just do our due diligence and test with a couple of queues data from the test server

zqian:
We have created a readonly DB user with access only to `meeting_start_logs` view. After merging this PR, we will verify whether the DB user still exists after drop-create the above view.

PR #776, opened 2026-04-03
Add date range filter to queue history export (#693)

### LINKED ISSUE #693: Provide a way to limit the results returned in a queue export
Currently when you export the queue, all events from all time are exported. It would be useful if there was a way to filter what was included in the export, at least by dates, possibly by other criteria.

We recently had a request [TDX 9001920](https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet?TicketID=9001920) that asked if the queue history could be deleted/cleared. However this would likely introduce additional complexity about whether it should be soft/hard deleted. And if it was hard deleted either intentionally or accidently it would not be recoverable and if it was soft deleted they might still want to export those events so we'd still have to provide an option. 

What the requester said is they were currently filtering by dates and this seems like the better solution which we could add a date filter as an option. Probably have two calendar pickers for a range, or a quick dropdown with last week/last month/last 3 months/last year. The API would have to be extended to allow for dates. 

We might also want a "report preview" page where the report is displayed in a table and then it can be adjusted. This seems like it would be a more complicated feature. 

This looks related to the ticket #516 where deleted queues were included in the full report but no indication about that. The user may want to exclude/include deleted queues and filter by date.

JohnnySonLee:
Resolves #693

## Summary

Adds a date range dropdown menu to queue history exports for individual and all queues. Users can now select "All history" (default), "Last 90 days", "Last 180 days", or "Last 365 days" before downloading a CSV. 

### Backend (`src/officehours_api/views.py`)
- `ExportMeetingStartLogs` view now accepts an optional `start_date` query parameter (format: `YYYY-MM-DD`)
- When provided, filters the `meeting_start_logs` query with `AND meeting_created_at >= start_date`
- Returns `400 Bad Request` with a descriptive message for invalid date formats
- backward compatible: omitting the parameter returns all data, same behavior as before

### Frontend
- **`src/assets/src/components/common.tsx`**: Added a date range dropdown next to each per-queue Download button in the QueueTable
- **`src/assets/src/components/manage.tsx`**: Added a date range dropdown next to the "Download All Queue History" button on the Manage page. The existing confirmation modal is preserved.
- **`src/assets/src/services/api.ts`**: Updated `exportQueueHistoryLogs` and `exportAllQueueHistoryLogs` to accept an optional `days` parameter, which computes a `start_date` and appends it as a query string

### 3 New Tests (`src/officehours_api/test_views.py`)
- `test_export_meeting_start_logs_with_start_date_includes_all`: Verifies that a past start_date returns all data
- `test_export_meeting_start_logs_with_start_date_future`: Verifies that a future start_date excludes all data (only header row returned)
- `test_export_meeting_start_logs_with_invalid_start_date`: Verifies that an invalid date string returns 400

### DIFF
--- a/src/assets/src/components/SingleQueueHistoryDownload.tsx
+++ b/src/assets/src/components/SingleQueueHistoryDownload.tsx
@@ -17,7 +17,7 @@ const SingleQueueHistoryDownload: React.FC<SingleQueueHistoryDownloadProps> = ({
 
     return (
         <div className="queue-history-filter">
-            <Form.Group>
+            <Form.Group controlId={`queueHistoryFilter-${queueId}`}>
                 <Form.Label visuallyHidden>
                     {`Date range filter for Queue ID ${queueId}`}
                 </Form.Label>
--- a/src/assets/src/components/SingleQueueHistoryDownload.tsx
+++ b/src/assets/src/components/SingleQueueHistoryDownload.tsx
@@ -1,0 +1,43 @@
+import { faFileDownload } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import React, { useState } from "react";
+import { Button, Form } from "react-bootstrap";
+
+interface SingleQueueHistoryDownloadProps {
+    queueId: number;
+    onDownload: (queueId: number, days?: number) => Promise<void>;
+}
+
+const SingleQueueHistoryDownload: React.FC<SingleQueueHistoryDownloadProps> = ({ queueId, onDownload }) => {
+    const [selectedDays, setSelectedDays] = useState<number | undefined>(undefined);
+
+    const handleDownload = async () => {
+        await onDownload(queueId, selectedDays);
+    };
+
+    return (
+        <div className="queue-history-filter">
+            <Form.Group>
+                <Form.Label visuallyHidden>
+                    {`Date range filter for Queue ID ${queueId}`}
+                </Form.Label>
+                <Form.Select
+                    className="queue-history-filter-select"
+                    value={selectedDays ?? ''}
+                    onChange={(e) => setSelectedDays(e.target.value ? Number(e.target.value) : undefined)}
+                >
+                    <option value="">All history</option>
+                    <option value="90">Last 90 days</option>
+                    <option value="180">Last 180 days</option>
+                    <option value="365">Last 365 days</option>
+                </Form.Select>
+            </Form.Group>
+            <Button className="queue-history-filter-btn" onClick={handleDownload}>
+                <span style={{paddingRight:"8px"}}><FontAwesomeIcon icon={faFileDownload} /></span>
+                Download
+            </Button>
+        </div>
+    );
+};
+
+export default SingleQueueHistoryDownload;
--- a/src/assets/src/components/common.tsx
+++ b/src/assets/src/components/common.tsx
@@ -11,6 +11,7 @@ import { useInitFocusRef } from "../hooks/useInitFocusRef";
 import { QueueAttendee, QueueBase, User } from "../models";
 import { sortQueues } from "../sort";
 import { ValidationResult } from "../validation";
+import SingleQueueHistoryDownload from "./SingleQueueHistoryDownload";
 
 type BootstrapButtonTypes = "info" | "warning" | "success" | "primary" | "alternate" | "danger";
 
@@ -527,12 +528,6 @@ interface QueueTableProps {
 
 export function QueueTable (props: QueueTableProps) {
     const linkBase = props.manageLink ? '/manage/' : '/queue/'
-    // Track which dropdown option is selected: undefined = all history, 90 = 90 days, etc
-    const [selectedDays, setSelectedDays] = useState<number | undefined>(undefined);
-    const handleQueueHistoryExportSubmit = async (queueId: number) => {
-        if (props.handleCSVDownload === undefined) return;
-        await props.handleCSVDownload(queueId, selectedDays);
-    }
 
     const sortedQueues = sortQueues(props.queues.slice());
     const queueItems = sortedQueues.map(q => (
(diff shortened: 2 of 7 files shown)


src/assets/src/components/common.tsx
jaydonkrooss, line 1: A selection for one queue impacts all the queues, because selectedDays state is shared. This is not the ideal pattern for a table with unique rows.

You could fix this by making the state a Map of numbers, but I highly suggest instead moving these changes into a new component with it's own internal state of option selected. Perhaps a new file with the component name SingleQueueHistoryDownload.

src/assets/src/components/common.tsx
```diff
@@ -16,7 +16,7 @@ type BootstrapButtonTypes = "info" | "warning" | "success" | "primary" | "altern
 
 export const DisabledMessage = <em></em>
 
-
+
```
jaydonkrooss: remove added whitespace

src/officehours_api/views.py
```diff
@@ -368,24 +368,45 @@ def get(self, request, queue_id=None, format=None):
...
+        start_date_str = request.query_params.get('start_date')
+        start_date = None
+        # If no filter chosen, return everything
+        if start_date_str:
+            try:
+                # attach UTC to avoid issues with postgres and comparing timezone-aware column against a naive datetime
+                start_date = datetime.strptime(start_date_str, '%Y-%m-%d').replace(tzinfo=timezone.utc)
+            except ValueError:
+                return Response(
+                    {'detail': 'Invalid start_date formate. Use YYYY-MM-DD.'},
```
jaydonkrooss: Typo fix
```suggestion
                    {'detail': 'Invalid start_date format. Use YYYY-MM-DD.'},
```

src/officehours_api/views.py
```diff
@@ -368,24 +368,45 @@ def get(self, request, queue_id=None, format=None):
         # Otherwise, get all the logs for the queues the user is a host of
         else:
             filename = f"meeting_start_logs_{username}.csv"
-
+        # Parse optional start_date queury parameter in order to filter by date
```
jaydonkrooss: Typo fix
```suggestion
        # Parse optional start_date query parameter in order to filter by date
```

src/assets/src/components/common.tsx
```diff
@@ -552,10 +554,23 @@ export function QueueTable (props: QueueTableProps) {
             </td>
             {props.includeCSVDownload && props.handleCSVDownload && (
                 <td className="align-middle" aria-label={`History for Queue ID ${q.id}`}>
-                    <Button onClick={() => handleQueueHistoryExportSubmit(q.id)}>
-                        <span style={{paddingRight:"8px"}}><FontAwesomeIcon icon={faFileDownload} /></span>
-                        Download
-                    </Button>
+                    <div style={{display: "flex", alignItems: "center", flexWrap: "nowrap"}}>
+                        <Form.Select
```
jaydonkrooss: Since you're bringing in [Bootstrap Form elements](https://react-bootstrap.netlify.app/docs/forms/overview), use `Form.Group` and `Form.Label` (with visually hidden on to replace the use of `aria-label`), it'll be more robust for accessibility and future changes/validation

src/assets/src/components/common.tsx
```diff
@@ -552,10 +554,23 @@ export function QueueTable (props: QueueTableProps) {
...
+                            value={selectedDays ?? ''}
+                            onChange={(e) => setSelectedDays(e.target.value ? Number(e.target.value) : undefined)}
+                            style={{width: "auto", flex: "0 0 auto", marginRight: "8px"}}
+                        >
+                            <option value="">All history</option>
+                            <option value="90">Last 90 days</option>
+                            <option value="180">Last 180 days</option>
+                            <option value="365">Last 365 days</option>
+                        </Form.Select>
+                        <Button style={{whiteSpace: "nowrap"}} onClick={() => handleQueueHistoryExportSubmit(q.id)}>
```
jaydonkrooss: I know this is a deviation that you didn't start, but let's try to keep the project's frontend pattern in-tact and remove/replace the use of the `style` attribute in favor of css classes. This project mainly uses main.css to define the style sheet, so it'll be preferable to stick with that in this PR.

jaydonkrooss (changes requested):
This appears to work well. My only major suggestion is to fix the an issue that I'm seeing with multiple rows sharing the same state. You'll want to create a new component for the single download queue portions, alongside some cleaning up of the styling.

src/assets/src/components/common.tsx
JohnnySonLee: ok, I'm seeing that as well. I'll work on that.

JohnnySonLee: It seems resolved on my end. Let me know if its not on yours.

src/assets/src/components/SingleQueueHistoryDownload.tsx
```diff
@@ -0,0 +1,43 @@
...
+const SingleQueueHistoryDownload: React.FC<SingleQueueHistoryDownloadProps> = ({ queueId, onDownload }) => {
+    const [selectedDays, setSelectedDays] = useState<number | undefined>(undefined);
+
+    const handleDownload = async () => {
+        await onDownload(queueId, selectedDays);
+    };
+
+    return (
+        <div className="queue-history-filter">
+            <Form.Group>
```
jaydonkrooss: You're almost there, the final change I'd do is add a unique `controlId` to `Form.Group`, which is provided in the [Bootstrap documentation](https://react-bootstrap.netlify.app/docs/forms/overview#overview), both here and in `manage.tsx`. It allows for the child Label and Select elements to associate together for screen readers.

jaydonkrooss (changes requested):
Sorry for the delay in reviewing this; the changes look great, there's one last edit that I think is important for accessibility. I'll try to be ready to accept the change, it should be very quick! Thanks again for contributing.

jaydonkrooss (approved):

